PR #2, opened 2019-02-06
Added support for Basic Auth

mkanoor:
This allows the API to be accessed externally with a userid/password
for a RHN user.
Moved the security definition in one place to be shared by all end
points of the API.


mkanoor:
@chambridge 
Please review

PR #6, opened 2019-02-07
Fixed Model Inheritance

### LINKED ISSUE #3: Incomplete Models
**Describe the bug**
Inherited models dont contain all the attributes

**To Reproduce**
1. Run openapi-generator on the json file and create a client
2. open the docs directory where the client is generated 
3. Examine the RoleIn.md file
4. It only lists **name** as the single property  **access** is missing

**Expected behavior**
The object definition should have included name and access properties

**Screenshots**
# RBACApiClient::RoleIn

## Properties
Name | Type | Description | Notes
------------ | ------------- | ------------- | -------------
**name** | **String** |  | 

This is true for several other objects in the definition, the problem seems to be a missing object node in the swagger definition when doing inheritance
https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/

```
allOf:     # Combines the BasicErrorModel and the inline model
        - $ref: '#/components/schemas/BasicErrorModel'
        - type: object
```

mkanoor:
Fixes #3

### DIFF
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -1,1509 +1,1570 @@
 {
-    "openapi": "3.0.0",
-    "info": {
-        "description": "The API for Role Based Access Control.",
-        "version": "1.0.0",
-        "title": "Role Based Access Control",
-        "license": {
-            "name": "AGPL-3.0",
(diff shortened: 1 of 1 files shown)


mkanoor:
@chambridge Please review

PR #21, opened 2019-02-26
Add code to remove policy in cases it could be orphaned. Closes #17.

### LINKED ISSUE #17: Test Case for orphaned policy
**Describe the bug**
Create a test case that covers the potential for orphaned policy.

Policies could be orphaned in two ways:
- Group is deleted
- Policy contained only one role and that role is deleted.

In both of the above cases the associated policy should be removed.

chambridge:


chambridge:
@mkanoor Test added to handle orphaned policy.

PR #43, opened 2019-03-20
Add access definitions for working with various RBAC APIs. Closes #22.

### LINKED ISSUE #22: Add permissions controls for RBAC APIs
## User Story
As a user, I want to control who can manage groups, roles, and polices so that its function is not open to all user.

## Impacts
API, Docs

## Role
Account Admin or user with RBAC permission

## Assumptions
- Only Account Admins will be allowed to create and manage groups (retains limited access to findUsers API)
- Any user can be given the access to manage roles and policies

## API Details

A user that is an account admin, as specified by the `is_org_admin` value in the user object of the `x-rh-identity` header, can perform any RBAC request.

Any user can call the `/access/` endpoint and the username contained in the `x-rh-identity` header will be used. Only an account admin can pass the `username` query parameter to determine access for a user.

Group management, `/group/*`,can only be managed by an account admin.

Role management can be done by an account administrator or a user with the following permissions:
- create/update/delete: `rbac:role:write` you must specify resource definition; can provide a list of uuid and the field or `*`
- get/list: `rbac:role:read` you must specify resource definition; can provide a list of uuid and the field or `*`

Policy management can be done by an account administrator or a user with the following permissions:
- create/update/delete: `rbac:policy:write` you must specify resource definition; can provide a list of uuid and the field or `*`
- get/list: `rbac:policy:read` you must specify resource definition; can provide a list of uuid and the field or `*`

## Acceptance Criteria

- [ ] Account admin can perform any RBAC call
- [ ] All users can call access endpoint
- [ ] Only account admin can pass username query param to access endpoint (otherwise 403)
- [ ] Only a user with appropriate permissions can create/update/delete roles (either a specified set or all)
- [ ] Only a user with appropriate permissions can get/list roles (either a specified set or all)
- [ ] Only a user with appropriate permissions can create/update/delete policies (either a specified set or all)
- [ ] Only a user with appropriate permissions can get/list policies (either a specified set or all)

chambridge:
* Principals API will remain open until integration with the back office proxy is complete.


chambridge:
Will add a few more test cases to up the diff coverage.

@mkanoor Feel free to add any comments. Tried to explain some of the permissions aspects for using the RBAC API in the RST documentation. See if it makes sense to you and you understand the necessary access needed to be defined to use the various APIs.

docs/source/management/role.rst
```diff
@@ -0,0 +1,68 @@
...
+- role
+- policy
+- aws.account
+- openshift.cluster
+
+Operation defines the application logic action, for example::
+
+- read
+- write
+- execute
```
mkanoor: @chambridge we dropped execute in favor of order, if you are not using execute we can skip it  here

chambridge: removed

PR #80, opened 2019-05-16
Fix Role/Policy to match spec

psav:
* Role was tweaked to return created/modified inline with spec
* Policy was tweaked to return created/modified inline with spec

Both these modifications were inline with existing code that was in
place for Group object.


chambridge:
@psav Thanks for catching this miss. However the models should mirror the following from the Group model: https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/group/model.py#L35

After you have made these updates you need to run `make make-migrations` to get the updated migration file to check-in. If you prefer I do this change that is fine too.

chambridge:
Nevermind, I see that these are serializer updates that just weren't exposed from the model. LGTM 👍

PR #116, opened 2019-08-08
Get roles for access principal. Closes #115.

### LINKED ISSUE #115: List the current roles assigned to a user
## User Story
The currently logged in user (normal user or admin user) would like to know the roles that he is assigned to.

## Impacts
The current get_principal_access list all the permissions assigned to a user, we would like to use the same api but limit the scope to roles. The default scope for the get_principal_access could be access but a user could curtail it at the role level.

## Role
A user

## Assumptions
- default type is `permission`

## API Details

| API  | Detail|
 | --- | --- |
 | **Method** | GET |
 | **Path** | /api/v1/access/?type=role |

## Acceptance Criteria

- [ ] Request returns all roles for a user when `type=role` is provided
- [ ] Request returns 400 when bad type is provided

chambridge:
FYI @mkanoor

### DIFF
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -1047,6 +1047,9 @@
             }
           },
           {
+            "$ref": "#/components/parameters/TypeFilter"
+          },
+          {
             "$ref": "#/components/parameters/QueryLimit"
           },
           {
@@ -1154,6 +1157,20 @@
         "description": "Parameter for ordering resource by value.",
         "schema": {
           "type": "string"
+        }
+      },
+      "TypeFilter": {
+        "in": "query",
+        "name": "type",
+        "required": false,
+        "description": "Parameter for obtaining access by type.",
+        "schema": {
+          "type": "string",
+          "enum": [
+            "permission",
+            "role"
+          ],
+          "default": "permission"
         }
       }
     },
@@ -1699,7 +1716,13 @@
               "data": {
                 "type": "array",
                 "items": {
-                  "$ref": "#/components/schemas/Access"
+                  "oneOf": [{
+                      "$ref": "#/components/schemas/Access"
+                    },
+                    {
+                      "$ref": "#/components/schemas/RoleOut"
+                    }
+                  ]
                 }
               }
             }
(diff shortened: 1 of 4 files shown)


docs/source/specs/openapi.json
```diff
@@ -1699,7 +1716,13 @@
               "data": {
                 "type": "array",
                 "items": {
-                  "$ref": "#/components/schemas/Access"
+                  "oneOf": [{
```
mkanoor: @chambridge Since this API returns one of the 2 objects would it make sense to have a separate API get_principal_roles distinct from get_principal_access?

chambridge: I don't have a strong opinion either way. I think it works as is, since we are looking at the breakdown of access for a principal. Role is a higher level than the permissions they would see. However, if you feel like it would be best to separate the two, I'm open to suggestions for the API endpoint. The above currently lives off: `api/rbac/v1/access/`

The following already exist and perform other functions:
`api/rbac/v1/principals/`
`api/rbac/v1/groups/`
`api/rbac/v1/roles/`
`api/rbac/v1/policies/`

mkanoor: @chambridge My preference would be a get_principal_roles in the RolesAPI class which would complement the get_principal_access in the AccessAPI class.

chambridge: Ok. What would the endpoint look? The OpenApi operationId doesn't really matter.

_GET_ `api/rbac/v1/roles/` - Returns all roles for the account, not for the calling user and any change to that functionality break the on-boarding flow and UI.

Could do something like:
_GET_ `api/rbac/v1/roles/current/` or _GET_ `api/rbac/v1/roles/me/` but those don't seem like an improvement to me over what is currently suggested.

mkanoor: @chambridge How about api/rbac/v1/roles?scope=principal and the default is
api/rbac/v1/roles which uses the default scope=tenant and doesn't break backward compatibility

chambridge: ok. I'm going to close this. and I'll work thru the above.

PR #138, opened 2019-10-24
Add more default role definitions

coderbydesign:
**Note:** _We are in the process of exploring migrating all default definitions to another
repository. In the meantime, this should continue to be the source of truth._

This PR adds default role definitions for the following applications:

- Compliance
- Insights
- Drift
- Vulnerability

Each of these applications needs a base, global read/write permission for
itself, as well as inventory, which is a dependent service which will be
receiving permissions from the upstream service. These permissions are set on
a base role for each application.

### DIFF
--- a/rbac/management/role/definitions/compliance.json
+++ b/rbac/management/role/definitions/compliance.json
@@ -1,0 +1,18 @@
+{
+  "roles": [
+    {
+      "name": "Compliance Access",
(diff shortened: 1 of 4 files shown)


coderbydesign:
Closing in favor of https://github.com/RedHatInsights/rbac-config/pull/2

PR #144, opened 2019-11-06
added account_filter variable to fix principal adding issue

aleccohan:
Added an account_filter variable to fix the principal adding issue caused by the BOP API.

See comments:
--> https://projects.engineering.redhat.com/browse/RHCLOUD-2354

### DIFF
--- a/rbac/management/principal/proxy.py
+++ b/rbac/management/principal/proxy.py
@@ -68,21 +68,27 @@ class PrincipalProxy:  # pylint: disable=too-few-public-methods
             params['offset'] = offset
         return params
 
-    @staticmethod
-    def _process_data(data, account):
+    def _process_data(self, data, account, account_filter):
         """Process data for uniform output."""
         processed_data = []
         for item in data:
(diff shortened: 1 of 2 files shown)


rbac/management/principal/proxy.py
```diff
@@ -164,10 +170,16 @@ def request_principals(self, account, limit=None, offset=None):
...
                                       self.path,
                                       account_principals_path)
-        return self._request_principals(url, params=params)
+
+        # For v1 account users endpoints are already filtered by account
+        return self._request_principals(url, params=params, account_filter=False)
 
     def request_filtered_principals(self, principals, account=None, limit=None, offset=None):
         """Request specific principals for an account."""
+        if account is None:
```
lphiri: I was going to propose that we just use presence of "account" and get rid of  "account_filter", but I think we should leave things as they are as a way to document the behavior.

aleccohan: I see!  That should be a pretty simple change if we want to make it after we test it out!

PR #148, opened 2019-11-08
Add `system` flag to policy

coderbydesign:
This adds a `system` flag on all policy records, defaulted to `false`. In coming
functionality, we will get/set a system policy per group, to bucket all roles
into as we obfuscate the concept of the group > policies > roles relationship with
group > roles.


rbac/management/migrations/0007_policy_system.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.2.4 on 2019-11-08 18:58
+
+from django.db import migrations, models
```
astrozzc: Cool! I did not noticed there is migrations file to change.

PR #149, opened 2019-11-12
Move default role to separate repo

astrozzc:
Default roles are separated into another repo. They would be deployed separately
but would be consumed by this service as they were in the same path. Any new
commits to the default roles in this repo would be ignored.


coderbydesign:
@astrozzc are we going to plan on leaving the definitions in the repo for local dev?

astrozzc:
> @astrozzc are we going to plan on leaving the definitions in the repo for local dev?

Yes, as we discussed, just leave them as they were for local dev. Once the service gets running in clusters, this path would be overwritten by the configs from the new repo.

PR #150, opened 2019-11-15
Seed default group when tenant is created

astrozzc:
When a tenant gets created, a default group would be created for the tenant.
All tenant would be under this group initially. And platform roles would be
assigned to the group through a specific platform policy.

Update would be in a following PR.

### DIFF
--- a/rbac/management/group/definer.py
+++ b/rbac/management/group/definer.py
@@ -1,0 +1,88 @@
+#
+# Copyright 2019 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+"""Handler for system defined group."""
+import logging
+
+from django.core.exceptions import ValidationError
+from management.group.model import Group
+from management.policy.model import Policy
+from management.role.model import Role
+from tenant_schemas.utils import tenant_context
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+
+def seed_group(tenant):
+    """For a tenant create or update default group."""
+    with tenant_context(tenant):
+        name = 'Platform group'
+        group_description = 'Default group that contains default permissions for new users.'
+
+        group, group_created = Group.objects.get_or_create(
+            platform_default=True,
+            defaults={
+                'description': group_description,
+                'name': name,
+                'system': True
+            }
+        )
+
+        if group.system:
+            platform_roles = Role.objects.filter(platform_default=True).values_list('uuid', flat=True)
+            add_roles(group, platform_roles, replace=True)
+            logger.info('Finished seeding default group %s for tenant %s.', name, tenant.schema_name)
+        else:
+            logger.info('Default group %s is managed by tenant %s.', name, tenant.schema_name)
+    return tenant
+
+
+def add_roles(group, roles, replace=False):
+    """Process list of roles and add them to the group."""
+    system_policy_name = 'System Policy for Group {}'.format(group.uuid)
+    system_policy, system_policy_created = Policy.objects.get_or_create(system=True,
+                                                                        group=group,
+                                                                        name=system_policy_name)
+    if system_policy_created:
+        logger.info('Created new system policy for tenant.')
+    else:
+        if replace:
+            system_policy.roles.clear()
+
+    for role_id in roles:
+        try:
+            role = Role.objects.get(uuid=role_id)
+            system_policy.roles.add(role)
+        except (Role.DoesNotExist, ValidationError):
+            logger.info('No role with id %s to save to group.', role_id)
+
+    system_policy.save()
+
+
+def remove_roles(group, role_ids):
+    """Process list of roles and remove them from the group."""
+    roles = []
+    for role_id in role_ids:
+        try:
+            role = Role.objects.get(uuid=role_id)
+            roles.append(role)
+        except (Role.DoesNotExist, ValidationError):
+            logger.info('No role with id %s to delete from group.', role_id)
+
+    for policy in group.policies.all():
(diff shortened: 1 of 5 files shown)


rbac/management/group/definer.py
```diff
@@ -0,0 +1,60 @@
...
+
+def _create_group(tenant, role_list):
+    """Create the group object in the database."""
+    with tenant_context(tenant):
+        name = "Platform group"
+        group_description = "Default group that contains default permissions for new users."
+        policy_description = "Platform default policy that is binded with default group."
+
+        principals = list(Principal.objects.all())
+        group = Group.objects.create(name=name,
```
coderbydesign: I think we'll just want to create this when it doesn't exist? Seeds are run more than just when a tenant is created.

rbac/management/group/definer.py
```diff
@@ -0,0 +1,60 @@
...
+                                     description=group_description,
+                                     system=True)
+        group.principals.add(*principals)
```
coderbydesign: The concept of system groups is that we don't need to add users to it. Otherwise we have to manage new users, etc. A "system" group should just imply all users have access (existing and all new), so we really need it to behave a bit differently, where all users get this group by default. Essentially `principal.groups` would be all groups for the principal, plus the system group.

rbac/management/group/model.py
```diff
@@ -32,6 +33,8 @@ class Group(models.Model):
     name = models.CharField(max_length=150, unique=True)
     description = models.TextField(null=True)
     principals = models.ManyToManyField(Principal, related_name='group')
+    roles = models.ManyToManyField(Role, related_name='group')
+    system = models.BooleanField(default=False)
```
coderbydesign: Is there a migration for adding this `system` field?

rbac/management/group/definer.py
```diff
@@ -0,0 +1,60 @@
...
+        policy.roles.add(*role_list)
+        policy.save()
+        logger.info('Created group %s for tenant %s.', name, tenant.schema_name)
+
+    return group
+
+
+def seed_group(tenant):
+    """For a tenant create default group."""
+    platform_role_list = list(Role.objects.filter(platform=True, system=True))
```
coderbydesign: I think the other flag is going to be "platform_default"? @wcmitchell unless you were going to go with just "platform"?

coderbydesign: Looks like @wcmitchell has it set here: https://github.com/RedHatInsights/insights-rbac/pull/151 😄

rbac/management/group/definer.py
astrozzc: I see, thanks for clarifying this. Then we should overwrite the principal groups method.

rbac/management/group/definer.py
astrozzc: Right, I would rebase my PR according to his.

rbac/management/group/definer.py
astrozzc: Right, I was thinking separate updating existing tenant into a followup PR. I will add that here then.

astrozzc:
The lint check always complains about the import order, after a few tries, I added the ignore after the import.

rbac/management/group/definer.py
```diff
@@ -0,0 +1,88 @@
...
+
+def seed_group(tenant):
+    """For a tenant create or update default group."""
+    with tenant_context(tenant):
+        name = 'Platform group'
+        group_description = 'Default group that contains default permissions for new users.'
+
+        group, group_created = Group.objects.get_or_create(
+            platform_default=True,
+            defaults={
```
coderbydesign: Just to clarify, for this `get_or_create`, the `get` will only query for `platform_default=True`, correct? The `defaults` are not a part of the get query? 

Just wanted to confirm, since the `system` flag will not always be `True`.

rbac/management/group/definer.py
```diff
@@ -0,0 +1,88 @@
...
+    """Process list of roles and add them to the group."""
+    system_policy_name = 'System Policy for Group {}'.format(group.uuid)
+    system_policy, system_policy_created = Policy.objects.get_or_create(system=True,
+                                                                        group=group,
+                                                                        name=system_policy_name)
+    if system_policy_created:
+        logger.info('Created new system policy for tenant.')
+    else:
+        if replace:
+            system_policy.roles.clear()
```
coderbydesign: So this will remove all roles, but only for the platform group, correct? If the tenant has another group with custom roles that have been added, it would not be affected by this, since we're only ever replacing roles on the system policy for the platform default group.

tests/management/group/test_definer.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def test_default_group_seeding_skips(self):
+        """Test that default groups with system flag false will be skipped during seeding"""
+
+        self.modify_default_group(system=False)
+
+        try:
+            seed_group(self.tenant)
+        except Exception:
+            self.fail(msg='update seed_group encountere an exception')
```
coderbydesign: Nitpick typo (and one other below).
```suggestion
            self.fail(msg='update seed_group encountered an exception')
```

tests/management/group/test_definer.py
```diff
@@ -0,0 +1,79 @@
...
+            group.roles().get(name="RBAC Administrator")
+
+    def test_default_group_seeding_reassign_roles(self):
+        """Test that previous assigned roles would be eliminated before assigning new roles."""
+        self.modify_default_group()
+
+        try:
+            seed_group(self.tenant)
+        except Exception:
+            self.fail(msg='update seed_group encountere an exception')
```
coderbydesign: Same here.
```suggestion
            self.fail(msg='update seed_group encountered an exception')
```

rbac/management/group/definer.py
astrozzc: Yes, it will only query for platform_default=True. It will only use the defaults to create.

rbac/management/group/definer.py
astrozzc: Yes, so the default for "replace" flag is false. For other groups other than the default group, we would just use default value, which would not. We would only set this to true when adding roles to default group.

I was thinking making the function for any one who might be interested in replacing the roles.

Do you want me to add another condition such as group.system is true to enforce this could only apply to platform group?

tests/management/group/test_definer.py
astrozzc: Good catch! Thanks 😁

tests/management/group/test_definer.py
astrozzc: Will change 👍

rbac/management/group/definer.py
coderbydesign: I don't think that's necessary, and might actually couple things too much to that implementation. I just more wanted to confirm there wouldn't be any unintended side-effects. 😄

PR #151, opened 2019-11-15
Add system group flag

wcmitchell:
Just adding in the flags we'll need for system groups and platform default groups.


rbac/management/group/view.py
```diff
@@ -206,6 +207,14 @@ def destroy(self, request, *args, **kwargs):
         @apiSuccessExample {json} Success-Response:
             HTTP/1.1 204 NO CONTENT
         """
+        group = get_object_or_404(Group, uuid=kwargs.get('uuid'))
+        if group.platform_default:
```
coderbydesign: Would we be able to throw a test in to validate this?

wcmitchell: Yeah, realized the need for that and working on it as we speak! Also adding the new field to the group serializer.

wcmitchell: Added tests in most recent commit

PR #153, opened 2019-11-18
Get roles from groups

coderbydesign:
957fc28
Adds two helper methods on group instances to get role info through policies:
```
group.roles()
group.role_count()
```
----
bb257c7 
Adds a `roles` collection to the GET request of `/groups/UUID/`, as well as
a `roleCount` field.

Also adds `roleCount` to the `/groups/` endpoint for each group entry.

This is to support the management of roles via groups, while still using the
underlying policy data model.


coderbydesign:
@chambridge - Ah, good call! I had the spec updates in another branch and ended up breaking this out to make for a few smaller PRs. I'll add that in this PR for these changes.

coderbydesign:
@chambridge - Updated `openapi.json` in 39a806acbf4e27f7b1f575b80ffe5bbc326c6ea0

docs/source/specs/openapi.json
```diff
@@ -1434,6 +1434,46 @@
...
+              "principals": {
+                "type": "array",
+                "items": {
+                  "$ref": "#/components/schemas/Principal"
+                }
+              }
+            }
+          },
+          {
+            "type": "object",
```
chambridge: Any reason why this is two `type: object` references?
I would have done the yaml as:
```
    GroupWithPrincipalsAndRoles:
      allOf:
        - $ref: '#/components/schemas/Group'
        - $ref: '#/components/schemas/UUID'
        - $ref: '#/components/schemas/Timestamped'
        - type: object
          required:
            - principals
            - roles
          properties:
            principals:
              type: array
              items:
                $ref: '#/components/schemas/Principal'
            roles:
              type: array
              items:
                $ref: '#/components/schemas/Role'
```

With the resulting JSON:
```
      "GroupWithPrincipalsAndRoles": {
        "allOf": [
          {
            "$ref": "#/components/schemas/Group"
          },
          {
            "$ref": "#/components/schemas/UUID"
          },
          {
            "$ref": "#/components/schemas/Timestamped"
          },
          {
            "type": "object",
            "required": [
              "principals",
              "roles"
            ],
            "properties": {
              "principals": {
                "type": "array",
                "items": {
                  "$ref": "#/components/schemas/Principal"
                }
              },
              "roles": {
                "type": "array",
                "items": {
                  "$ref": "#/components/schemas/Role"
                }
              }
            }
          }
        ]
      },
```

Doesn't seem to complain in the editor, so it may not be wrong to do it that way. I was just wondering.

coderbydesign: No, I think you're right. I've updated it. Thanks!

tests/management/group/test_model.py
```diff
@@ -0,0 +1,60 @@
...
+
+        with tenant_context(self.tenant):
+            self.group = Group.objects.create(name='groupA')
+            self.roleA = Role.objects.create(name='roleA')
+            self.roleB = Role.objects.create(name='roleB')
+            self.policy = Policy(name='policyA', group=self.group)
+            self.policy.save()
+            self.policy.roles.add(self.roleA)
+            self.policy.save()
+            self.group.policies.add(self.policy)
```
astrozzc: Wow, how great! This is our first model test! Just curious, could this line be removed?

PR #158, opened 2019-11-26
Add system flag for group

astrozzc:
This flag would be used to determine if a group would be managed by user
or us.


astrozzc:
> Looks good @astrozzc! The only suggestion I'd have is if we're surfacing this in the API GETs, we probably want to add it (and it looks like the `platform_default` actually), to the `GroupOut` properties in `openapi.json`?

Good call! Added it.

PR #159, opened 2019-11-27
Block API role creation for non-cost-mgmt

wcmitchell:
Taking a swing at preventing role creation involving permissions for
apps other than cost-management.

### DIFF
--- a/rbac/management/role/view.py
+++ b/rbac/management/role/view.py
@@ -16,6 +16,8 @@
 #
 
 """View for role management."""
+import os
+
 from django.db import transaction
 from django.db.models.aggregates import Count
 from django.shortcuts import get_object_or_404
@@ -29,7 +31,8 @@ from rest_framework.filters import OrderingFilter
 from .model import Role
 from .serializer import RoleSerializer
 
-APP_WHITELIST = ['app', 'cost-management']
+TESTING_APP = os.getenv('TESTING_APPLICATION', '')
+APP_WHITELIST = [TESTING_APP, 'cost-management']
 
 
 class RoleFilter(filters.FilterSet):
--- a/rbac/management/role/view.py
+++ b/rbac/management/role/view.py
@@ -123,7 +123,7 @@ class RoleViewSet(mixins.CreateModelMixin,
                 ]
             }
         """
-        for perm in request.data.get('access') or []:
+        for perm in request.data.get('access', []):
             app = perm.get('permission').split(':')[0]
             if app not in APP_WHITELIST:
                 key = 'role'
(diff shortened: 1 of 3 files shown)


wcmitchell:
This is....likely not the most elegant solution, but it seems to do what we need. I set up a whitelist for the apps where we'll allow role creation, currently limited to 'cost-management' and 'app', which is used in much of our testing. Adding 'app' to the whitelist shouldn't break anything or open any security holes and seemed preferable to changing all of the testing to creating roles to 'cost-management.' As other apps come online we can add them to the whitelist as needed. Currently the whitelist is just set up alongside the role views, but it may be something to move out to a settings or configuration file somewhere.

rbac/management/role/view.py
```diff
@@ -29,6 +29,8 @@
 from .model import Role
 from .serializer import RoleSerializer
 
+APP_WHITELIST = ['app', 'cost-management']
```
coderbydesign: Hm, trying to think if there'd be a more graceful way to handle the test allowance here. Having to put something specific in the app code for testing isn't ideal. Can we maybe make use of the .env file? Or would it be a huge headache to update the tests to just reference `cost-management`? 

It just might be confusing having this `app` allowance here, and although like you said it doesn't expose anything, it does allow you to create those orphaned `app` roles.

rbac/management/role/view.py
```diff
@@ -142,7 +153,8 @@ def list(self, request, *args, **kwargs):
         @apiSuccess {Object} links  The object containing links of results.
         @apiSuccess {Object[]} data  The array of results.
 
-        @apiSuccessExample {json} Success-Response:
+        @apiSuccessExample {json} Success-Re
+        sponse:
```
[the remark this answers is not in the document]
wcmitchell: Yep! Lemme fix that >_<

rbac/management/role/view.py
wcmitchell: We could update the testing to use 'cost-management' but that feels hacky rather than using a specific 'testing' application. Maybe something with the `.env` file would work. I'll look into that a bit.

rbac/management/role/view.py
```diff
@@ -121,6 +123,15 @@ def create(self, request, *args, **kwargs):
                 ]
             }
         """
+        for perm in request.data.get('access') or []:
```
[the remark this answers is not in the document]
wcmitchell: Yeah, that's probably more Pythonic, lemme switch that.

wcmitchell:
Fixed the two quick wins, still poking about for testing application and blocking users who don't have access to cost-management from accessing it. I'm not entirely sure how to implement the second, so I'll be digging into that tomorrow

rbac/management/role/view.py
coderbydesign: @wcmitchell - Sounds good! Likewise, coupling the app logic to tests also doesn't feel right, if we're able to avoid it.

PR #160, opened 2019-12-03
Merge `platform-rbac` feature branch into `master`

coderbydesign:
Now that our deploy pipeline is setup to protect production from `master`, we should get our feature branch into `master` so that we can delivery current work into CI and begin working directly off `master` again.

~**TODO:**~
~- [ ] merge #159 into `platform-rbac`~


astrozzc:
I just clicked the update branch because I have some changes in master related to smoke test. Which would not affect this rebase.

PR #161, opened 2019-12-03
Add platform_default attribute when seeding roles

astrozzc:
We have added platform_default for roles which would be part of the platform default group.
This PR is to add platform_default attribute for roles to distinguish them.


rbac/management/role/definitions/ansible-automation.json
```diff
@@ -0,0 +1,19 @@
+{
```
astrozzc: This is added on purpose to make tests successful locally.

PR #168, opened 2019-12-06
Fix apidoc comments

coderbydesign:
This helps format the rendering of the Django REST framework UI for the api.

**Before:**
before

**After:**
after

### DIFF
--- a/rbac/management/access/view.py
+++ b/rbac/management/access/view.py
@@ -26,8 +26,8 @@ from rest_framework.views import APIView
 
 
 class AccessView(APIView):
-    """Obtain principal access list.
-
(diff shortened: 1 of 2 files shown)


coderbydesign:
@astrozzc Does it? I'm not seeing it surface locally or in CI for roles, or any others.

astrozzc:
> @astrozzc Does it? I'm not seeing it surface locally or in CI for roles, or any others.

Oh, I thought this format would cause this problem: https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/role/view.py#L75

NVM if not. I have some problem setting up the Django REST framework

PR #170, opened 2019-12-09
Adding functionality and tests for system flag flips

wcmitchell:
We want to set the system flag to false on any altered platform_default
groups

### DIFF
--- a/rbac/management/group/view.py
+++ b/rbac/management/group/view.py
@@ -255,10 +255,6 @@ class GroupViewSet(mixins.CreateModelMixin,
                 "name": "GroupA"
             }
         """
-        group = Group.objects.filter(uuid=kwargs.get('uuid')).first()
-        if group and group.platform_default:
-            group.system = False
-            group.save()
         return super().update(request=request, args=args, kwargs=kwargs)
 
     def add_principals(self, group, principals, account):
--- a/rbac/management/group/view.py
+++ b/rbac/management/group/view.py
@@ -22,7 +22,7 @@ from django.db.models.aggregates import Count
 from django.shortcuts import get_object_or_404
 from django.utils.translation import gettext as _
 from django_filters import rest_framework as filters
-from management.group.definer import add_roles, remove_roles
+from management.group.definer import add_roles, remove_roles, set_system_flag_post_update
 from management.group.model import Group
 from management.group.serializer import (GroupInputSerializer,
                                          GroupPrincipalInputSerializer,
@@ -442,9 +442,7 @@ class GroupViewSet(mixins.CreateModelMixin,
             if serializer.is_valid(raise_exception=True):
                 roles = request.data.pop(ROLES_KEY, [])
             add_roles(group, roles)
-            if group.platform_default:
-                group.system = False
-                group.save()
+            set_system_flag_post_update(group)
             response_data = GroupRoleSerializerIn(group)
         elif request.method == 'GET':
             serialized_roles = [RoleMinimumSerializer(role).data for role in group.roles()]
@@ -461,9 +459,7 @@ class GroupViewSet(mixins.CreateModelMixin,
             serializer = GroupRoleSerializerIn(data={'roles': role_ids})
             if serializer.is_valid(raise_exception=True):
                 remove_roles(group, role_ids)
-                if group.platform_default:
-                    group.system = False
-                    group.save()
+                set_system_flag_post_update(group)
 
             return Response(status=status.HTTP_204_NO_CONTENT)
 
--- a/rbac/management/group/view.py
+++ b/rbac/management/group/view.py
@@ -255,6 +255,10 @@ class GroupViewSet(mixins.CreateModelMixin,
                 "name": "GroupA"
             }
         """
+        group = Group.objects.filter(uuid=kwargs.get('uuid')).first()
+        if group and group.platform_default:
+            group.system = False
+            group.save()
         return super().update(request=request, args=args, kwargs=kwargs)
 
     def add_principals(self, group, principals, account):
@@ -442,6 +446,9 @@ class GroupViewSet(mixins.CreateModelMixin,
             if serializer.is_valid(raise_exception=True):
                 roles = request.data.pop(ROLES_KEY, [])
             add_roles(group, roles)
+            if group.platform_default:
+                group.system = False
+                group.save()
             response_data = GroupRoleSerializerIn(group)
         elif request.method == 'GET':
             serialized_roles = [RoleMinimumSerializer(role).data for role in group.roles()]
@@ -458,6 +465,9 @@ class GroupViewSet(mixins.CreateModelMixin,
             serializer = GroupRoleSerializerIn(data={'roles': role_ids})
             if serializer.is_valid(raise_exception=True):
                 remove_roles(group, role_ids)
+                if group.platform_default:
+                    group.system = False
+                    group.save()
 
             return Response(status=status.HTTP_204_NO_CONTENT)
 
--- a/tests/management/group/test_view.py
+++ b/tests/management/group/test_view.py
@@ -56,7 +56,7 @@ class GroupViewsetTests(IdentityRequest):
             self.group.principals.add(self.principal)
             self.group.save()
 
-            self.defGroup = Group(name='groupDef', platform_default=True)
+            self.defGroup = Group(name='groupDef', platform_default=True, system=True)
             self.defGroup.save()
             self.defGroup.principals.add(self.principal)
             self.defGroup.save()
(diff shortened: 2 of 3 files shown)


rbac/management/group/view.py
```diff
@@ -253,6 +253,10 @@ def update(self, request, *args, **kwargs):
                 "name": "GroupA"
             }
         """
+        group = Group.objects.filter(uuid=kwargs.get('uuid')).first()
```
astrozzc: I am wondering should we modify the flag here? I thought we would only change it to False if we just change the roles. Here this update seems only updating the name or description?

wcmitchell: I went back and forth on that a bit, and my feeling was that we'd likely want ANY edit to a platform_default role, even just a name change, to modify the system flag. If we don't care so much in this case, we can pull this flag change out, though.

lphiri: @wcmitchell @astrozzc suggest confirming with Nicole. But I also suggest we flip the switch for any change to keep things simple.

tests/management/group/test_view.py
```diff
@@ -176,6 +176,21 @@ def test_update_group_success(self, mock_request):
...
+        group = Group.objects.filter(name='groupDef').first()
+        updated_name = group.name + '_update'
+        test_data = {'name': updated_name}
+        url = reverse('group-detail', kwargs={'uuid': group.uuid})
+        client = APIClient()
+        response = client.put(url, test_data, format='json', **self.headers)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        self.assertIsNotNone(response.data.get('uuid'))
+        self.assertEqual(updated_name, response.data.get('name'))
```
coderbydesign: Should we have an assertion here to check the `system` flag value?

tests/management/group/test_view.py
```diff
@@ -311,6 +326,46 @@ def test_add_group_roles_system_policy_create_success(self):
...
+    def test_system_flag_update_on_add(self):
+        """Test that adding a role to a platform_default group flips the system flag."""
+        url = reverse('group-roles', kwargs={'uuid': self.defGroup.uuid})
+        client = APIClient()
+        test_data = {'roles': [self.roleB.uuid, self.dummy_role_id]}
+
+        response = client.post(url, test_data, format='json', **self.headers)
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertFalse(Group.objects.filter(uuid=self.defGroup.uuid).first().system)
```
coderbydesign: Maybe a nit, but if you wanted to just reference `self.defGroup.system` directly here again, you could also do something like:

```
self.defGroup.refresh_from_db()
self.assertFalse(self.defGroup.system)
```

tests/management/group/test_view.py
```diff
@@ -311,6 +326,46 @@ def test_add_group_roles_system_policy_create_success(self):
         self.assertEqual(roles[0].get('description'), self.role.description)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
 
+    def test_system_flag_update_on_add(self):
+        """Test that adding a role to a platform_default group flips the system flag."""
+        url = reverse('group-roles', kwargs={'uuid': self.defGroup.uuid})
+        client = APIClient()
+        test_data = {'roles': [self.roleB.uuid, self.dummy_role_id]}
+
```
coderbydesign: Might be worth adding an assertion before the request as well, just as a sanity check to ensure the initial value?

```
self.assertTrue(self.defGroup.system)
```

tests/management/group/test_view.py
```diff
@@ -311,6 +326,46 @@ def test_add_group_roles_system_policy_create_success(self):
...
+        response = client.delete(url, format='json', **self.headers)
+
+        self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
+        self.assertFalse(Group.objects.filter(uuid=self.defGroup.uuid).first().system)
+
+
+    def test_system_flag_update_on_rename(self):
+        """Test that renaminga platform_default group flips the system flag."""
+        group = Group.objects.filter(uuid=self.defGroup.uuid).first()
```
coderbydesign: Same as above, with being able to reference `self.defGroup` directly if you want?

tests/management/group/test_view.py
```diff
@@ -311,6 +326,46 @@ def test_add_group_roles_system_policy_create_success(self):
...
+    def test_system_flag_update_on_remove(self):
+        """Test that removing a role from a platform_default group flips the system flag."""
+        url = reverse('group-roles', kwargs={'uuid': self.defGroup.uuid})
+        client = APIClient()
+        url = '{}?roles={}'.format(url, self.roleB.uuid)
+
+        self.policy.roles.add(self.roleB)
+        self.policy.save()
+
+        response = client.delete(url, format='json', **self.headers)
```
coderbydesign: Same as above if you wanted to add an assertion before, and reload the object after?

tests/management/group/test_view.py
```diff
@@ -311,6 +326,46 @@ def test_add_group_roles_system_policy_create_success(self):
...
+
+        self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
+        self.assertFalse(Group.objects.filter(uuid=self.defGroup.uuid).first().system)
+
+
+    def test_system_flag_update_on_rename(self):
```
coderbydesign: Is there a difference in what this is supposed to test vs `test_update_platform_default_group_success`?

rbac/management/group/view.py
```diff
@@ -440,6 +444,9 @@ def roles(self, request, uuid=None):
             if serializer.is_valid(raise_exception=True):
                 roles = request.data.pop(ROLES_KEY, [])
             add_roles(group, roles)
+            if group.platform_default:
```
coderbydesign: Since this logic:
```
if group.platform_default:
    group.system = False
    group.save()
```
is used in three different places, maybe we could break it out into a helper method like `set_system_flag_on_update(group)` or something? Then just call that from each of the three places? The first instance where you check for the group first could just be:
```
if group:
    set_system_flag_on_update(group)
```
and the other places it could just call it directly. That would also allow for that logic to be unit tested if we wanted, and if the logic changed in the future we could just updated it once.

rbac/management/group/view.py
coderbydesign: Based on the UX presentation last demo, I'm not sure the group name for the default groups will even be editable via the UI? It looks like they plan to auto-update the name when a role is added/removed, which are the triggers related to managing/not managing the group. So my only thought would be, that may be a disconnect between the UI and the API.

Maybe we actually want to prevent the names of platform default groups from being edited through the API, and have a hook during this process to update the name in a standard way for consistency, per the UX expectations? So, flip the flag to false and prepend the name in a standard way, but don't allow the platform default groups' names to be edited. Just a thought.

astrozzc: OK👍 I  will keep in mind of the APIs UI have to add info for customers to maintain the group for them.

tests/management/group/test_view.py
```diff
@@ -176,6 +176,21 @@ def test_update_group_success(self, mock_request):
         self.assertIsNotNone(response.data.get('uuid'))
         self.assertEqual(updated_name, response.data.get('name'))
 
+    @patch('management.principal.proxy.PrincipalProxy.request_filtered_principals',
+           return_value={'status_code': 200, 'data': []})
+    def test_update_platform_default_group_success(self, mock_request):
+        """Test that updating a default group changes system flag."""
+        group = Group.objects.filter(name='groupDef').first()
```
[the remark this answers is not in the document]
wcmitchell: Oh, yep, good idea. I think I'd gotten into the habit of pulling from the DB when verifying other things during testing. I'll fix that real quick.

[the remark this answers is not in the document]
wcmitchell: Actually just ended up pulling this entire test method out since it was testing functionality we don't need.

tests/management/group/test_view.py
wcmitchell: I did not know that! Thanks! I'll use that instead.

tests/management/group/test_view.py
wcmitchell: Pulled this method out, testing functionality we don't need

tests/management/group/test_view.py
wcmitchell: Nope, I don't think so. Just got lost in the code and wrote the same test twice, I think. Pulled it out.

tests/management/group/test_view.py
wcmitchell: Good idea, adding that

PR #171, opened 2019-12-10
Fix spec

astrozzc:
The doc file has mismatch with the actural api respons.

### DIFF
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -613,7 +613,18 @@
             "content": {
               "application/json": {
                 "schema": {
-                  "$ref": "#/components/schemas/GroupWithPrincipalsAndRoles"
+                  "type": "object",
+                  "required": [
+                    "data"
+                  ],
+                  "properties": {
+                    "data": {
+                      "type": "array",
+                      "items": {
+                        "$ref": "#/components/schemas/RoleOut"
+                      }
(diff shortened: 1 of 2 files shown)


docs/source/specs/openapi.json
```diff
@@ -613,7 +613,7 @@
             "content": {
               "application/json": {
                 "schema": {
-                  "$ref": "#/components/schemas/GroupWithPrincipalsAndRoles"
+                  "$ref": "#/components/schemas/GroupRolesPagination"
```
coderbydesign: I think this will just have `RoleOut`, not `GroupRolesPagination`, won't it?

catastrophe-brandon:
Looks good to me!

docs/source/specs/openapi.json
catastrophe-brandon: If we change it to RoleOut, let me know. For the sake of charging forward with test automation I used GroupRolesPagination when I generated my most recent client source.

coderbydesign: Yeah, I don't think we're returning the paginated schema for a POST response. We're just returning the `RoleOut` schema, unless I'm mistaken.

astrozzc: Right, it will return a list of RoleOut. Just updated 👍

PR #172, opened 2019-12-11
Access endpoint query performance

coderbydesign:
This PR addresses some n+1 queries related to the `/access/` endpoint, which 
exist due to the nested relationship queries, by eager loading some of the
associations where possible.

Since we reuse a handful of the methods for getting data by principal, by abstracting 
two lookup options:
```
prefetch_lookups_for_ids
prefetch_lookups_for_groups
```
we can specify what data we'd like to preload. This PR just covers the `/access/`
eager loading, by preloading `policies`, `roles` and `access` when we get groups during
the access chain, via `prefetch_lookups_for_groups`.

We don't want to always eager load these relations though, since we can also get
just `groups`, or `roles` scoped to a `principal`, which reuses those methods. That
would cause those queries to respond slower.

As a follow-up, we can look at adding any other eager loading for those scoped
endpoints for `groups` and `roles`.

The other option: `prefetch_lookups_for_ids` specifies which relation(s) to load when
we resort the queryset of objects by ID. 

**Note:** we are currently doubling some DB hits because of this additional sort. 
We get the objects via a queryset, put them in a unique set (which means we no 
longer have the queryset), and then query again from that set to get a sorted queryset. 
It appears we should be able to improve this as well.

**Local data setup:**
--
**Principal Count:** 1
**Group Count:** 2
**Policy Count:** 20
**Role Count:** 850
**Access Count:** 85,000 (same application)
**Resource Definition Count:** 85,000 (same application)

**Results:**
--
**Before:**
```
/access/?application=<app_name>&limit=10
2.165s average response time

/access/?application=<app_name>&limit=<total_count>
5.174s average response time
```

**After:**
```
/access/?application=<app_name>&limit=10
576ms average response time (73.4% improvement)

/access/?application=<app_name>&limit=<total_count>
2.146s average response time (58.5% improvement, with a payload of 1.5mb)
```

### DIFF
--- a/rbac/management/querysets.py
+++ b/rbac/management/querysets.py
@@ -135,7 +135,8 @@ def get_access_queryset(request):
     return get_object_principal_queryset(request,
                                          PRINCIPAL_SCOPE,
                                          Access,
-                                         **{APPLICATION_KEY: app})
+                                         **{APPLICATION_KEY: app,
+                                            'prefetch_lookups': 'resourceDefinitions'})
 
 
 def get_object_principal_queryset(request, scope, clazz, **kwargs):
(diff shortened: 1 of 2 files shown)


chambridge:
COOL 😎

PR #174, opened 2019-12-11
Attempt to add the TESTING_APPLICATION env var

catastrophe-brandon:
Try and make the smoke tests happy again by adding "test" application to the whitelist.


catastrophe-brandon:
This change shouldn't be needed anymore.

PR #178, opened 2019-12-16
Scoped roles endpoint query performance

coderbydesign:
**Address n+1 queries in GET `/roles/?scope=principal`** (47f32e4)
Similar to #172, when roles are scoped to principal, we end up running into
n+1 issues due to needing the policy/role records for the groups for a principal.

This resolves that by added the necessary prefetch lookups on the `get_role_queryset`
calls, and sets up a base query to be reused/extended upon.

We can also setup a `base_query` queryset to build upon in `get_role_queryset`.

---

**Optimize role serializer queries** (b7f299c)
We can remove n+1 queries for roles by prefetching `access` records in the queries
where the serializer is being used on roles.

We don't want to add this to the default query in `group#roles()` because we also use that method for getting the count, in which case we'd be prematurely fetching unnecessary data.

--- 

**Results:** _(same dataset as #172)_
--
**Before:**
```
/roles/?scope=principal
2.79s average response time

/groups/<uuid>/roles/
2.72s average response time
```

**After:**
```
/roles/?scope=principal
729ms average response time (73.87% improvement)

/groups/<uuid>/roles/
978ms average response time (64.04% improvement)
```

### DIFF
--- a/rbac/management/querysets.py
+++ b/rbac/management/querysets.py
@@ -81,12 +81,15 @@ def get_group_queryset(request):
 def get_role_queryset(request):
     """Obtain the queryset for roles."""
     scope = request.query_params.get(SCOPE_KEY, ACCOUNT_SCOPE)
+    base_query = Role.objects.prefetch_related('access').annotate(policyCount=Count('policies', distinct=True))
     if scope != ACCOUNT_SCOPE:
-        return get_object_principal_queryset(request, scope, Role)
+        return get_object_principal_queryset(request, scope, Role,
+                                             **{'prefetch_lookups_for_ids': 'access',
+                                                'prefetch_lookups_for_groups': 'policies__roles'})
     if ENVIRONMENT.get_value('ALLOW_ANY', default=False, cast=bool):
-        return Role.objects.annotate(policyCount=Count('policies', distinct=True))
+        return base_query
     if request.user.admin:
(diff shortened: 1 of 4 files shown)


rbac/management/querysets.py
```diff
@@ -95,8 +98,8 @@ def get_role_queryset(request):
     if not res_list:
         return Role.objects.none()
     if '*' in res_list:
-        return Role.objects.annotate(policyCount=Count('policies', distinct=True))
-    return Role.objects.filter(uuid__in=res_list).annotate(policyCount=Count('policies', distinct=True))
+        return base_query
+    return base_query.filter(uuid__in=res_list)
```
astrozzc: LGTM overall, but as described in the note section, the "filter" would make a new queryset. Should we use prefetch object instead? https://docs.djangoproject.com/en/1.7/ref/models/querysets/#prefetch-related

coderbydesign: @astrozzc based on our conversation, it looks like you're correct, however since this is isolated to a case for RBAC on RBAC where a user would have explicit permissions for a subset of roles, I don't think we need to be concerned with this not adding performance improvements to this instance. Thanks for the detailed find!

https://stackoverflow.com/questions/12973929/why-does-djangos-prefetch-related-only-work-with-all-and-not-filter/12974801#12974801

PR #180, opened 2019-12-17
Protect platform_default groups on update as well as delete

wcmitchell:

### DIFF
--- a/rbac/management/group/view.py
+++ b/rbac/management/group/view.py
@@ -97,8 +97,9 @@ class GroupViewSet(mixins.CreateModelMixin,
             return GroupInputSerializer
         return GroupSerializer
 
-    def protect_default_groups(self, group, action):
+    def protect_default_groups(self, uuid, action):
         """Deny modifications on platform_default groups."""
+        group = get_object_or_404(Group, uuid=uuid)
         if group.platform_default:
             key = 'group'
             message = '{} cannot be performed on platform default groups.'.format(action.upper())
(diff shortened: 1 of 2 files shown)


rbac/management/group/view.py
```diff
@@ -255,6 +259,8 @@ def update(self, request, *args, **kwargs):
                 "name": "GroupA"
             }
         """
+        group = get_object_or_404(Group, uuid=kwargs.get('uuid'))
```
astrozzc: Nit: could we put this into the function just use uuid as input? And do you mind adding a unit test for the update, just like the one you did for delete?

rbac/management/group/view.py
```diff
@@ -97,6 +97,16 @@ def get_serializer_class(self):
             return GroupInputSerializer
         return GroupSerializer
 
+    def protect_default_groups(self, group, action):
+        """Deny modifications on platform_default groups."""
+        if group.platform_default:
+            key = 'group'
+            message = '{} cannot be performed on platform default groups.'.format(action)
```
coderbydesign: One minor suggestion: the responses might read better if we uppercase the action?
```suggestion
            message = '{} cannot be performed on platform default groups.'.format(action.upper())
```

```
{
  "errors": [
    {
      "detail": "UPDATE cannot be performed on platform default groups.",
      "source": "group",
      "status": 400
    }
  ]
}
```

rbac/management/group/view.py
wcmitchell: Sure! That seems reasonable. I realized after opening this PR that I'd forgotten a test. I added one in a later commit but it seems that didn't get properly added into the PR. I'll get it re-added. Thanks!

PR #182, opened 2019-12-19
Fix 500 instead of 404 status on request with invalid uuid

wcmitchell:

### DIFF
--- a/rbac/management/group/view.py
+++ b/rbac/management/group/view.py
@@ -19,7 +19,6 @@
 import logging
 
 from django.db.models.aggregates import Count
-from django.shortcuts import get_object_or_404
 from django.utils.translation import gettext as _
 from django_filters import rest_framework as filters
 from management.group.definer import add_roles, remove_roles, set_system_flag_post_update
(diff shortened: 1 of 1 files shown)


rbac/management/group/view.py
```diff
@@ -97,9 +96,9 @@ def get_serializer_class(self):
             return GroupInputSerializer
         return GroupSerializer
 
-    def protect_default_groups(self, uuid, action):
+    def protect_default_groups(self, action):
         """Deny modifications on platform_default groups."""
-        group = get_object_or_404(Group, uuid=uuid)
+        group = self.get_object()
```
astrozzc: Nice! Works like magic!!! I was so curious why it works differently although it uses get_object_or_404 too, and found this, they redefined this get_object_or_404: https://github.com/encode/django-rest-framework/blob/master/rest_framework/generics.py#L7

astrozzc:
This happens to roles, which also uses this get_object_or_404: https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/role/view.py#L250

PR #183, opened 2019-12-19
Add the `x-rh-identity` header to the openapi.json spec

coderbydesign:
We should express that the x-rh-identity header is required for all requests to
resource-related endpoints in the spec, along with a schema.

https://swagger.io/docs/specification/describing-parameters/#header-parameters

### DIFF
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -81,6 +81,9 @@
         "operationId": "listPrincipals",
         "parameters": [
           {
+            "$ref": "#/components/parameters/IdentityHeader"
+          },
+          {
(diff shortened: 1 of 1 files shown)


coderbydesign:
I'm not sure we want to do this, actually. Since API clients are built from the spec, this will change client code and enforce that the header is supplied on each request.

While it's true that we do require this on each request, it does not technically apply to external consumers of the API. For external consumers, the identity header is not of importance, as the gateway handles proxying the request from Akamai, to the internal services, with the identity header.

This really only applies to service-to-service requests, where internal applications are communicating directly with RBAC, circumventing the gateway.

We should add this elsewhere to document for integration, but not enforce it in the spec.

PR #184, opened 2019-12-19
Add validation for access list

astrozzc:
Input of access is not a list would cause 500 instead of 400.
This should be fix and warn customers about it.
Also, if access object is not wel-defined, it should be 400.


astrozzc:
> Nice. I like the change to not only validate the list and prevent a 500 there, but also the serializer validation on the data in the list. 👍 That really helps with the clarity of the message for malformed requests.
> 
> Just to clarify, currently roles can be created without any access supplied?

That's a good point, maybe it is better to limit that? Not to allow any empty list?

PR #185, opened 2019-12-19
Put properties inside allOf

astrozzc:
If the properties is outside 'allOf', the openapi generator will ignore it.

### DIFF
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -1777,26 +1777,28 @@
           },
           {
             "$ref": "#/components/schemas/Timestamped"
+          },
+          {
+            "properties": {
+              "principalCount": {
+                "type": "integer",
+                "minimum": 0
+              },
+              "roleCount": {
+                "type": "integer",
+                "minimum": 0
+              },
+              "system": {
+                "type": "boolean",
+                "default": false
+              },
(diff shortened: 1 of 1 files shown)


astrozzc:
Tested with openapi generator and this time it has the platform_default:
# GroupOut
  
## Properties
Name | Type | Description | Notes
------------ | ------------- | ------------- | -------------
**name** | **str** |  |
**description** | **str** |  | [optional]
**uuid** | **str** |  |
**created** | **datetime** |  |
**modified** | **datetime** |  |
**principal_count** | **int** |  | [optional]
**role_count** | **int** |  | [optional]
**system** | **bool** |  | [optional] [default to False]
**platform_default** | **bool** |  | [optional] [default to False]

catastrophe-brandon:
Thanks for checking it with the generator!

astrozzc:
> It's crazy to me that the openapi editor shows it as being correct even when it's not nested correctly:
> 
> Screen Shot 2019-12-20 at 8 29 34 AM
> 
> I know we'd talked about that a week or so ago, but obviously it does affect the generated content. Nice change!

Right, that's hard to debug. We have similar issue some time ago that's why I come up this solution quick, I have checked this is the last one 😁

PR #186, opened 2019-12-19
Update default group name to mirror that presented in UX

wcmitchell:


rbac/management/group/definer.py
```diff
@@ -30,7 +30,7 @@
 def seed_group(tenant):
     """For a tenant create or update default group."""
     with tenant_context(tenant):
-        name = 'Platform group'
+        name = 'Default user access'
```
astrozzc: I am afraid this might not update the current group name. We just get the platform_default=True group and update the roles in it below.

wcmitchell: Ahh, that's a good point. Let me fix that.

wcmitchell: Thinking more on it, would I need to put something in the code to handle existing group names, or just get a new migration created to affect the existing groups?

coderbydesign: I think we'd probably want a data migration to update the existing data as a one-off vs code to handle this. Moving forward, since the group name has a unique validation and disallow changing the name, we wouldn't need to worry about the names of these default groups being incorrect. Thoughts?

wcmitchell: That kinda fits along with my thinking, yeah.

astrozzc: Agree 👍

wcmitchell:
I *think* this may work to alter existing groups to the new names as needed. Seems to work in testing

rbac/management/migrations/0011_group_naming.py
```diff
@@ -0,0 +1,22 @@
...
+from django.db import migrations, models
+
+
+def update_default_name(apps, schema_editor):
+    Group = apps.get_model('management', 'Group')
+    for group in Group.objects.all():
+        if group.system:
+            group.name = "Default user access"
+        else:
+            group.name = "Custom default user access"
```
coderbydesign: I don't think we want to update all groups, do we? We just need to ensure that currently, any 'system' group is "Default user access". Then on update, we'll bump it to "Custom default user access", and flip the `system` flag but the data migration just needs to effectively sync the data as it is now. 

Otherwise normal groups will pickup the name "Custom default user access", and actually I would think that might break since there's a unique constraint on group names?

wcmitchell: Oh, yeah. I was thinking about current, formerly system groups, but this is a bit broad for that and those may not really be a concern. I'll just strike the else.

rbac/management/migrations/0011_group_naming.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 2.2.4 on 2019-11-26 17:03
+
+from django.db import migrations, models
+
+
+def update_default_name(apps, schema_editor):
+    Group = apps.get_model('management', 'Group')
+    for group in Group.objects.all():
+        if group.system:
+            group.name = "Default user access"
```
coderbydesign: Just ran this locally, and my existing system groups didn't update. I think we want to do `group.save()` after this?
```suggestion
            group.name = "Default user access"
            group.save()
```

PR #192, opened 2020-01-24
Add migration aimed at removing RBAC on RBAC roles

wcmitchell:
This seems like it works from some local testing, but definitely needs a couple of extra eyes on it.


wcmitchell:
Ugh, and tests failing. let me fix that. Left in some PDB statements

rbac/management/migrations/0012_remove_RonR_resources.py
```diff
@@ -0,0 +1,38 @@
...
+        # get the role for the permission (bubbling up)
+        role = rbac_permission.role
+
+        # check to see if there are any access objects that are not RBAC
+        non_rbac_permissions = role.access.exclude(permission__contains='rbac:')
+
+        # if so, just delete the access object we know is RBAC, and leave the role
+        if non_rbac_permissions:
+            # this will still delete the resource definitions for the access object
+            access.delete()
```
coderbydesign: This actually needs to be `rbac_permission` here, I believe?
```suggestion
            rbac_permission.delete()
```

rbac/management/migrations/0012_remove_RonR_resources.py
```diff
@@ -0,0 +1,38 @@
...
+        else:
+            # otherwise, we only have RBAC access objects, so delete everything
+            role.delete()
+
+
+
```
coderbydesign: Just a nit, but we can probably do away with a few of these newlines?

rbac/management/migrations/0012_remove_RonR_resources.py
wcmitchell: Yep, I think you're right. That's what I get for playing with naming just before submitting >_<

PR #198, opened 2020-02-06
Add role access GET endpoint: /roles/<uuid>/access/

coderbydesign:
This add the `GET` endpoint: `/roles/<uuid>/access/` to the API. While the access objects are currently returned as a nested resource within `/roles/<uuid>/`, the UI is requesting that we provide an explicit endpoint for retrieving access objects for a given role, in order to paginate and eventually filter on this data directly.

The payload will look like:
```
{
  "meta": {
    "count": 2,
    "limit": 10,
    "offset": 0
  },
  "links": {
    "first": "/api/rbac/v1/roles/3ecac858-4f48-4393-a26b-914ba73477aa/access/?limit=10&offset=0",
    "next": null,
    "previous": null,
    "last": "/api/rbac/v1/roles/3ecac858-4f48-4393-a26b-914ba73477aa/access/?limit=10&offset=0"
  },
  "data": [
    {
      "permission": "inventory:*:*",
      "resourceDefinitions": []
    },
    {
      "permission": "ansible-automation:*:*",
      "resourceDefinitions": []
    }
  ]
}
```

And is represented in the openapi.json spec:
Screen Shot 2020-02-06 at 10 15 37 AM

Currently there is no filtering on this endpoint, only pagination.

### DIFF
--- a/rbac/management/role/view.py
+++ b/rbac/management/role/view.py
@@ -332,10 +332,9 @@ class RoleViewSet(mixins.CreateModelMixin,
         except (Role.DoesNotExist, ValidationError):
             raise Http404
 
-        if role:
-            access = AccessSerializer(role.access, many=True).data
-            page = self.paginate_queryset(access)
-            return self.get_paginated_response(page)
+        access = AccessSerializer(role.access, many=True).data
+        page = self.paginate_queryset(access)
+        return self.get_paginated_response(page)
 
     def validate_and_get_access_list(self, data):
         """Validate if input data contains valid access list and return."""
--- a/rbac/management/role/view.py
+++ b/rbac/management/role/view.py
@@ -18,14 +18,17 @@
 """View for role management."""
 import os
 
+from django.core.exceptions import ValidationError
 from django.db import transaction
 from django.db.models.aggregates import Count
+from django.http import Http404
 from django.utils.translation import gettext as _
 from django_filters import rest_framework as filters
 from management.permissions import RoleAccessPermission
 from management.querysets import get_role_queryset
 from management.role.serializer import AccessSerializer, RoleMinimumSerializer
 from rest_framework import mixins, serializers, viewsets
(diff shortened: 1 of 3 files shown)


rbac/management/role/view.py
```diff
@@ -321,6 +324,19 @@ def update(self, request, *args, **kwargs):
...
+        """Return access objects for specified role."""
+        try:
+            role = Role.objects.get(uuid=uuid)
+        except (Role.DoesNotExist, ValidationError):
+            raise Http404
+
+        if role:
+            access = AccessSerializer(role.access, many=True).data
+            page = self.paginate_queryset(access)
+            return self.get_paginated_response(page)
```
astrozzc: Maybe put these into the try block? 

Will there be a use case to make the default no pagination, like the one you did for /access endpoint? (currently default limit is 10)

coderbydesign: Hey @astrozzc! In this case, since the access objects aren't being used the same as the `/access/?application=` endpoint, we'll be keeping the default pagination setup. The experience team actually requested it all be paginated as well.

astrozzc: Got it, thanks 👍

astrozzc: How about putting these into the try block? That way we dont need the if statement

coderbydesign: @astrozzc Actually, we can probably just get rid of the conditional altogether now, since the `try` above will either return a role, or fail with a 404. Or is there another potential edge-case we should try to catch?

astrozzc: > @astrozzc Actually, we can probably just get rid of the conditional altogether now, since the `try` above will either return a role, or fail with a 404. Or is there another potential edge-case we should try to catch?

Sure, both works!

coderbydesign: @astrozzc - updated!

PR #199, opened 2020-02-06
Add GET method to /group/:uuid/principals/

wcmitchell:

### DIFF
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -549,18 +549,7 @@
             "content": {
               "application/json": {
                 "schema": {
-                  "type": "object",
-                  "required": [
-                    "principals"
-                  ],
-                  "properties": {
-                    "principals": {
-                      "type": "array",
-                      "items": {
-                        "$ref": "#/components/schemas/PrincipalPagination"
-                      }
-                    }
-                  }
+                  "$ref": "#/components/schemas/PrincipalPagination"
                 }
               }
             }
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -557,7 +557,7 @@
                     "principals": {
                       "type": "array",
                       "items": {
-                        "$ref": "#/components/schemas/PrincipalOut"
+                        "$ref": "#/components/schemas/PrincipalPagination"
                       }
                     }
                   }
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -525,6 +525,74 @@
           }
         }
       },
+      "get": {
+        "tags": [
+          "Group"
+        ],
+        "summary": "Get a list of principals from a group in the tenant",
+        "operationId": "getPrincipalsFromGroup",
+        "parameters": [
+          {
+            "name": "uuid",
+            "in": "path",
+            "description": "ID of group from which to get principals",
+            "required": true,
+            "schema": {
+              "type": "string",
+              "format": "uuid"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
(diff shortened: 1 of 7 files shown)


wcmitchell:
Oh, right! I need to do that too. Thanks for the reminder!

rbac/management/group/view.py
```diff
@@ -360,7 +385,8 @@ def principals(self, request, uuid=None):
             if isinstance(resp, dict) and 'errors' in resp:
                 return Response(status=resp['status_code'], data=resp['errors'])
             output = GroupSerializer(resp)
-            return Response(status=status.HTTP_200_OK, data=output.data)
+        elif request.method == 'GET':
+            output = PrincpalSerializer(group.principals.all(), many=True)
```
coderbydesign: I'm guessing we'll probably want to have this data be paginated like all other (apart from `/access/`) endpoints?

rbac/management/group/view.py
```diff
@@ -32,6 +32,7 @@
 from management.permissions import GroupAccessPermission
 from management.principal.model import Principal
 from management.principal.proxy import PrincipalProxy
+from management.principal.serializer import PrincpalSerializer
```
coderbydesign: Minor type here and below where it's called. 
```suggestion
from management.principal.serializer import PrincipalSerializer
```

rbac/management/group/view.py
wcmitchell: Yeah, I went back and forth on that. Wasn't sure if it was worth paginating a bare list of Principals, but I can do that, no problem!

wcmitchell:
This worked in local testing, but I'm a bit concerned about the simplicity of the proxy call and how it will behave in real-world situations.

rbac/management/group/view.py
```diff
@@ -360,7 +383,16 @@ def principals(self, request, uuid=None):
             if isinstance(resp, dict) and 'errors' in resp:
                 return Response(status=resp['status_code'], data=resp['errors'])
             output = GroupSerializer(resp)
-            return Response(status=status.HTTP_200_OK, data=output.data)
+            response = Response(status=status.HTTP_200_OK, data=output.data)
+        elif request.method == 'GET':
+            page = self.paginate_queryset(group.principals.all())
+            serializer = PrincipalSerializer(page, many=True)
+            usernames = serializer.data
```
coderbydesign: I think this is going to end up being an ordered dictionary here, which is what the `users` value in the [POST payload](https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/principal/proxy.py#L188) will be set to, and it looks like we actually [need a list](https://backoffice-proxy-insights-services.ext.us-east.aws.preprod.paas.redhat.com/docs/#/default/postUsers)?

wcmitchell: It has been returning an OrderedDict, but I wasn't sure if it needed that level of object specificity, but I can do a quick translation of the data into a list here.

docs/source/specs/openapi.json
```diff
@@ -525,6 +525,74 @@
...
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "List of principals attached to group",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "object",
```
coderbydesign: I think here for the 200 we may just need:
```
          "200": {
            "description": "List of principals attached to group",
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/PrincipalPagination"
                }
              }
            }
          }
```
The output in https://editor.swagger.io/ for this as is doesn't match the actual response, as it nests everything in a principal lists.

PR #205, opened 2020-02-12
Add settings for smoke test run

catastrophe-brandon:
To get the smoke tests to pick up the required settings from Jenkins we need to do the following:

1. Merge this change into master
2. Set up a PR with this change targeting stable. (We can incorporate this change into the existing PR for stable once we have it in master).

This is required because the groovy file with the change needs to be available in the rbac container when running the smoke tests.


catastrophe-brandon:
I'm not sure if we'll be able to merge this to master if we get approvals. I hate to ask, but will we need to temporarily disable checks to get it through?

PR #206, opened 2020-02-14
Add x-rh-identity security scheme

semtexzv:
Adds a new security scheme which denotes requirement of sending `x-rh-identity` header.
Requirement for automatically generated clients to work.

Same approach is taken by [host based inventory](https://github.com/RedHatInsights/insights-host-inventory/blob/b21784c73528cc64773efee11ea5820b5e8cb109/swagger/api.spec.yaml#L345).

### DIFF
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -12,6 +12,9 @@
   "security": [
     {
       "basic_auth": []
+    },
+    {
+      "identity_auth": []
     }
   ],
   "tags": [
@@ -1681,6 +1684,12 @@
         "type": "http",
         "description": "The userid/password is needed when accessing this API externally",
         "scheme": "basic"
+      },
+      "identity_auth": {
+        "type": "apiKey",
+        "in": "header",
+        "name": "x-rh-identity",
+        "description": "Internal authorization scheme for cross-service communication"
       }
     },
     "schemas": {


coderbydesign:
Hey @semtexzv! We'd actually opened a PR (https://github.com/RedHatInsights/insights-rbac/pull/183) a while ago to add the requirement of the header, but I don't believe we can safely do this, as it only applies to service-to-service requests.

Any consumer of the API coming through https://cloud.redhat.com/api/rbac/ has the `x-rh-identity` header added by the platform gateway, and external clients shouldn't be concerned with with that:

> I don't think we want to do this, actually. Since API clients are built from the spec, this will change client code and enforce that the header is supplied on each request.
> 
> While it's true that we do require this on each request, it does not technically apply to external consumers of the API. For external consumers, the identity header is not of importance, as the gateway handles proxying the request from Akamai, to the internal services, with the identity header.
> 
> This really only applies to service-to-service requests, where internal applications are communicating directly with RBAC, circumventing the gateway.
> 
> We should add this elsewhere to document for integration, but not enforce it in the spec.

I'm not sure if there's another/better way to represent this to help with your auto-generated client?

cc: @wcmitchell @astrozzc

semtexzv:
Yes, this is unfortunate, that we have 2 authorization schemes depending on from where the service is called, but from the OpenAPI spec perspective this is fine. By having 2 security definitions applied to an endpoint, We allow consumer of the api to use either of them. Allowing both of these in the spec and using out-of-band information, like additional documentation & naming the internal security scheme `internal` should cover our use-case without disruption to external consumers of the API.

astrozzc:
> Yes, this is unfortunate, that we have 2 authorization schemes depending on from where the service is called, but from the OpenAPI spec perspective this is fine. By having 2 security definitions applied to an endpoint, We allow consumer of the api to use either of them.
> 
> Allowing both of these in the spec and using out-of-band information, like additional documentation & naming the internal security scheme `internal` should cover our use-case without disruption to external consumers of the API.

LGTM

docs/source/specs/openapi.json
```diff
@@ -1681,6 +1684,11 @@
         "type": "http",
         "description": "The userid/password is needed when accessing this API externally",
         "scheme": "basic"
+      },
+      "identity_auth": {
+        "type": "apiKey",
+        "in": "header",
```
coderbydesign: Can we add a description here so it's apparent that this is only necessary when accessing the API directly, outside of cloud.redhat.com?

coderbydesign:
@semtexzv got it, thanks! I just had one suggestion which might help communicate this a bit better in the spect, but otherwise that makes sense to me.

docs/source/specs/openapi.json
coderbydesign: Thanks!

PR #207, opened 2020-02-14
Add count metadata to /principals/ endpoint

wcmitchell:

### DIFF
--- a/rbac/management/principal/view.py
+++ b/rbac/management/principal/view.py
@@ -116,8 +116,13 @@ class PrincipalView(APIView):
         response_data = {}
         if status_code == status.HTTP_200_OK:
             data = resp.get('data', [])
+            if isinstance(data, dict):
+                count = data.get('userCount')
+                data = data.get('users')
+            else:
+                count = None
             response_data['meta'] = {
-                'count': None
+                'count': count
             }
             response_data['links'] = {
                 'first': f'{path}?limit={limit}&offset=0{usernames_filter}',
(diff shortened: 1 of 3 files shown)


rbac/management/principal/view.py
```diff
@@ -116,8 +116,13 @@ def get(self, request):
         response_data = {}
         if status_code == status.HTTP_200_OK:
             data = resp.get('data', [])
+            if isinstance(data, dict):
```
astrozzc: Will there be another followup PR? The above [proxy.request_principals](https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/principal/view.py#L112) still using the default and calling the v1 endpoint

It would be great if a unit test is added 😁

wcmitchell: Made some changes, should actually use the right endpoint now, but I didn't set up a test yet. I'm not sure how to reasonably test these changes when we'd largely just be passing in the full data from the mock patch. Any suggestions?

astrozzc: Right, since you have tested them with mock data, it should be good. Could test that once it is deployed in CI to verify.
I am fine with the changes. @coderbydesign Keith, do you have any suggestions?

coderbydesign: Since we're using the new version of the API, it's probably worth at least updating the mocked return values in all of the specs (example: https://github.com/RedHatInsights/insights-rbac/blob/master/tests/management/principal/test_view.py#L57) to reflect the v2 response?

That would test this as well in the process. We could leave one of the existing tests, and maybe add a check to ensure the `count = None` branch is executed if the payload is formatted like v1 still?

astrozzc: Good idea 👍

PR #208, opened 2020-02-17
Change default group description

PanSpagetka:
Hello, as a part of https://projects.engineering.redhat.com/browse/RHCLOUD-4258 at https://marvelapp.com/48ia9ib/screen/64603663 screen is requested to change description of default group. Since it is just a string change I have tried to make a PR. Please tell me if it needs to be changed at multiple places or any other necessary changes ;)

Here is PR with discussion about string change: https://github.com/RedHatInsights/insights-rbac-ui/pull/145

### DIFF
--- a/rbac/management/group/definer.py
+++ b/rbac/management/group/definer.py
@@ -31,7 +31,10 @@ def seed_group(tenant):
     """For a tenant create or update default group."""
     with tenant_context(tenant):
         name = 'Default user access'
-        group_description = 'Default group that contains default permissions for new users.'
+        group_description = (
+            'This group contains the roles that all users inherit by default. '
+            'Adding or removing roles in this group will affect permissions for all users in your organization.'
+        )
 
         group, group_created = Group.objects.get_or_create(
             platform_default=True,


rbac/management/group/definer.py
```diff
@@ -31,7 +31,8 @@ def seed_group(tenant):
     """For a tenant create or update default group."""
     with tenant_context(tenant):
         name = 'Default user access'
-        group_description = 'Default group that contains default permissions for new users.'
+        group_description = 'These are the roles all users inherit by default.\
```
astrozzc: Thanks for refining the description, now it looks much better! 
How about change "These are the roles all ..." to "This group defines the roles all ..." ?

mmenestr:
Talked with Nicole, and we ended up with this 

> "This group contains the roles that all users inherit by default." instead of "These are all the roles all....".

So instead of saying "defines", we thought "contains" was a good medium, and took the "This group ..." wording!

astrozzc:
> Talked with Nicole, and we ended up with this:
> 
> > This group contains the roles that all users inherit by default.
> 
> instead of "These are all the roles all....".
> 
> So instead of saying "defines", we thought "contains" was a good medium, and took the "This group ..." wording!

Sounds good to me 👍

rbac/management/group/definer.py
```diff
@@ -31,7 +31,8 @@ def seed_group(tenant):
     """For a tenant create or update default group."""
     with tenant_context(tenant):
         name = 'Default user access'
-        group_description = 'Default group that contains default permissions for new users.'
+        group_description = 'This group contains the roles that all users inherit by default.\
```
coderbydesign: I believe this will add some unnecessary whitespace to the saved string. Maybe something like this?
```python
group_description = ("This group contains the roles that all users inherit by default. "
                    "Adding or removing roles in this group will affect permissions for all users in your organization.")
```

PanSpagetka: Thanks, it is fiexed now :)

coderbydesign:
@PanSpagetka before this goes in, I just wanted to confirm that it's ok that existing default groups in CI/QA will still have the old value? This will only apply to new tenants. If that's not ok, we'll need to include a migration, or manually update the existing groups.

PanSpagetka:
@coderbydesign that is actually a good point. It would be better to update existing groups. However, I am not very familiar with Django framework, could you or @astrozzc do it?

astrozzc:
> @coderbydesign that is actually a good point. It would be better to update existing groups. However, I am not very familiar with Django framework, could you or @astrozzc do it?

I might be able to work on our data base. Let me check with Keith

astrozzc:
@PanSpagetka I have manually update the descriptions in CI/QA 👍

PR #211, opened 2020-02-18
Add url filters to /groups/<uuid>/roles/

coderbydesign:
This adds the following url param filters to roles for groups:

```
/api/rbac/v1/groups/<uuid>/roles/?role_name=
/api/rbac/v1/groups/<uuid>/roles/?role_description=
```

Using filters prefixed with `role_` for clarity within the group view, as well
as the fact that there would be a naming collision within DRF with the existing
group ordering fields being used, primarily for `name`.

### DIFF
--- a/rbac/management/group/view.py
+++ b/rbac/management/group/view.py
@@ -44,6 +44,7 @@ USERNAMES_KEY = 'usernames'
 ROLES_KEY = 'roles'
 EXCLUDE_KEY = 'exclude'
 VALID_EXCLUDE_VALUES = ['true', 'false']
+VALID_GROUP_ROLE_FILTERS = ['role_name', 'role_description']
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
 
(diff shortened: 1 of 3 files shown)


rbac/management/group/view.py
```diff
@@ -482,14 +483,26 @@ def roles(self, request, uuid=None):
 
         return Response(status=status.HTTP_200_OK, data=response_data.data)
 
+    def filtered_roles(self, roles, request):
+        """Return filtered roles for group from query params."""
+        role_filters = {}
+        for param_name, param_value in request.query_params.items():
+            if param_name in VALID_GROUP_ROLE_FILTERS:
+                attr_filter_name = param_name.replace('role_', '')
```
astrozzc: Could we just use 'name' and 'description' without the 'role_'? which is easier to type

rbac/management/group/view.py
```diff
@@ -482,14 +483,26 @@ def roles(self, request, uuid=None):
...
+                role_filters[f'{attr_filter_name}__icontains'] = param_value
```
astrozzc: So we want to support just case insensitive search (Tom Jerry when we input tom), not returning e.g. Tommy when we input tom

rbac/management/group/view.py
coderbydesign: @astrozzc unfortunately, it doesn't appear so. I noted in the description of the PR the issue I was running into with the `name` parameter. Might be worth looking into further, though.

rbac/management/group/view.py
coderbydesign: Correct, the other filters we have setup are case insensitive, so I implemented this one the same.

PR #212, opened 2020-02-18
Clean up DB env config

kylape:
- Hard coded db engine (for simplicity's sake)
- Standardized DB env var names

Needs to be merged along with https://github.com/RedHatInsights/e2e-deploy/pull/1109


kylape:
uhhh I pushed a commit that isn't showing up on this PR

.env.example
```diff
@@ -1,12 +1,12 @@
 DATABASE_SERVICE_NAME=POSTGRES_SQL
```
coderbydesign: We could probably pull this out as well?

tox.ini
```diff
@@ -21,8 +21,8 @@ setenv =
   DATABASE_SERVICE_NAME={env:DATABASE_SERVICE_NAME:POSTGRES_SQL}
```
coderbydesign: We could probably pull this out as well?

coderbydesign:
@kylape I added @chambridge to this and the e2e PR for some visibility if he's able to give us a few to call out any red flags since he has some context. 😄

chambridge:
Changes are fine. Legacy skeleton brought over from other projects.

PR #215, opened 2020-02-22
Update endpoint perms to allow nonadmin to list groups in tenant

wcmitchell:


wcmitchell:
Huh, that's odd @coderbydesign, it looked like it was working for me in local testing, but I was honestly only half away when I was working on it. I'll look back at it and see what it was doing, thanks!

PR #216, opened 2020-02-24
Ensure default groups are returned using `username` filter on `/groups/`

coderbydesign:
This fixes a bug where the platform default group for a tenant was not being
returned when the `?username=` filter is used.

There are two reasons this is occurring. First is that the queryset being build
when a username filter is used does not include the default groups. Second is that
the current filter lookup builds the query to filter on the principal relationship
with groups, by principal username. Since the platform default groups aren't related
directly to principal records, they will always be excluded from the queryset.

To resolve this, we should always merge the platform default group set in on the
queryset as well as the filter, similar to what we do when we get groups for a principal.

- sets up a helper on the Group model to get the `platform_default_set`
- reuses helper in utils, the group view, and the group queryset
- implements a custom filter for the group username filter to also include the
  platform default group set

### DIFF
--- a/rbac/management/querysets.py
+++ b/rbac/management/querysets.py
@@ -63,7 +63,7 @@ def get_group_queryset(request):
         if username != identity_username:
             return Group.objects.none()
         else:
-            return Group.objects.filter(principals__username=username) | Group.platform_default_set()
+            return Group.objects.filter(principals__username=username)
     access = request.user.access
     access_op = 'read'
     if request.method in ('POST', 'PUT'):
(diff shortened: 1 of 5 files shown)


rbac/management/querysets.py
```diff
@@ -63,7 +63,7 @@ def get_group_queryset(request):
         if username != identity_username:
             return Group.objects.none()
         else:
-            return Group.objects.filter(principals__username=username)
+            return Group.objects.filter(principals__username=username) | Group.platform_default_set()
```
wcmitchell: Just to capture the discussion from the review meeting, I think we can remove this in favor of just having the union with Group.platform_default_set() in group/view.py

coderbydesign: Thanks for the reminder/review! I'll push an update.

coderbydesign: @wcmitchell updated!

PR #218, opened 2020-02-24
Add username filter on groups principals get

coderbydesign:
This reuses a similar pattern added for filtering roles on groups, so it's now
broken out into a helper to build the filters for nested resources on groups.

Now, a filter `?principal_username=` is supported on /groups/<uuid>/principals/
in order to return principal details for a group, filtered by username.

### DIFF
--- a/tests/management/group/test_view.py
+++ b/tests/management/group/test_view.py
@@ -17,7 +17,7 @@
 """Test the group viewset."""
 import random
 from decimal import Decimal
-from unittest.mock import patch
+from unittest.mock import patch, ANY
 from uuid import uuid4
 
 from django.urls import reverse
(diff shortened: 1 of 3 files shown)


tests/management/group/test_view.py
```diff
@@ -454,6 +454,34 @@ def test_all_role_filters_for_group_roles_match(self):
         self.assertEqual(len(roles), 1)
         self.assertEqual(roles[0].get('uuid'), str(self.role.uuid))
 
+    @patch('management.principal.proxy.PrincipalProxy.request_filtered_principals',
+        return_value={'status_code': 200, 'data': []})
+    def test_principal_username_filter_for_group_roles_no_match(self, mock_request):
```
astrozzc: Is this should be group_principals? and below

PR #219, opened 2020-02-24
Remove catalog and approval from app whitelist

wcmitchell:


astrozzc:
I guess Nicole would inform those teams been removed

PR #220, opened 2020-02-24
Optionally add group count or data

astrozzc:
Support adding additional field to display when listing roles.

This PR adds groupInCount and groups_in fields which represents
the total number of groups the roles is in and the groups which
contains the role respectively.

### DIFF
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -976,6 +976,24 @@
           },
           {
             "$ref": "#/components/parameters/OrderByFilter"
+          },
+          {
+            "name": "add_fields",
+            "in": "query",
+            "required": false,
+            "description": "Parameter for add list of fields to display for roles.",
+            "schema": {
+              "type": "array",
+              "items": {
+                "type": "string",
+                "enum": [
+                  "groups_in",
+                  "groups_in_count"
+                ]
+              }
+            },
+            "explode": false,
+            "style": "form"
           }
         ],
         "responses": {
@@ -984,7 +1002,7 @@
             "content": {
               "application/json": {
                 "schema": {
-                  "$ref": "#/components/schemas/RolePagination"
+                  "$ref": "#/components/schemas/RolePaginationDynamic"
                 }
               }
             }
@@ -1934,6 +1952,21 @@
           }
         }
       },
+      "AdditionalGroup": {
+        "required": [
+          "name"
+        ],
+        "properties": {
+          "name": {
+            "type": "string",
+            "example": "GroupA"
+          },
+          "uuid": {
+            "type": "string",
+            "example": "234df936-abb4-4238-a1c9-d91fc540c702"
+          }
+        }
+      },
       "GroupOut": {
         "allOf": [
           {
@@ -2207,6 +2240,27 @@
           }
         ]
       },
+      "RolePaginationDynamic": {
+        "allOf": [
+          {
+            "$ref": "#/components/schemas/ListPagination"
+          },
+          {
+            "type": "object",
+            "required": [
+              "data"
+            ],
+            "properties": {
+              "data": {
+                "type": "array",
+                "items": {
+                  "$ref": "#/components/schemas/RoleOutDynamic"
+                }
+              }
+            }
+          }
+        ]
+      },
       "RoleOut": {
         "allOf": [
           {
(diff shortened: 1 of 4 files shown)


rbac/management/role/view.py
```diff
@@ -36,6 +36,10 @@
 
 TESTING_APP = os.getenv('TESTING_APPLICATION')
 APP_WHITELIST = ['cost-management', 'catalog', 'approval']
+ADDITIONAL_FIELD_KEY = 'add_fields'
```
coderbydesign: If this can only be one field, maybe we should change the value to `add_field`. If we're able to support multiple fields, maybe changing the constant name to `ADDITIONAL_FIELDS_KEY` makes more sense.

rbac/management/role/view.py
```diff
@@ -349,3 +363,17 @@ def validate_and_get_access_list(self, data):
...
+    def validate_and_get_additional_field_key(self, params):
+        """Validate the add field key."""
+        field = params.get(ADDITIONAL_FIELD_KEY)
+        if field is not None and field not in VALID_FIELD_VALUES:
+            key = 'detail'
+            message = '{} query parameter value {} is invalid. booleans are valid inputs.'.format(
+                ADDITIONAL_FIELD_KEY,
+                field)
+            raise serializers.ValidationError({key: _(message)})
+        if field is None:
```
coderbydesign: This works, but we may also be able to use list comprehension here to remove the `None` value and avoid this conditional? Something like:

```python
all_fields = LIST_ROLE_FIELDS + [field]
return [f for f in all_fields if f]
```

rbac/management/role/serializer.py
```diff
@@ -160,6 +161,60 @@ def get_applications(self, obj):
         return obtain_applications(obj)
 
 
+class DynamicFieldsModelSerializer(serializers.ModelSerializer):
```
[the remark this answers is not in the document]
astrozzc: As we discussed offline, decided to do this when this is used in more places

rbac/management/role/view.py
```diff
@@ -349,3 +363,17 @@ def validate_and_get_access_list(self, data):
         for access in access_list:
             AccessSerializer(data=access).is_valid(raise_exception=True)
         return access_list
+
+    def validate_and_get_additional_field_key(self, params):
+        """Validate the add field key."""
+        field = params.get(ADDITIONAL_FIELD_KEY)
+        if field is not None and field not in VALID_FIELD_VALUES:
+            key = 'detail'
+            message = '{} query parameter value {} is invalid. booleans are valid inputs.'.format(
```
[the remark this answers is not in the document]
astrozzc: Fixed 👍

rbac/management/role/view.py
```diff
@@ -349,3 +363,17 @@ def validate_and_get_access_list(self, data):
         for access in access_list:
             AccessSerializer(data=access).is_valid(raise_exception=True)
         return access_list
+
+    def validate_and_get_additional_field_key(self, params):
+        """Validate the add field key."""
+        field = params.get(ADDITIONAL_FIELD_KEY)
```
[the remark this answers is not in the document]
astrozzc: Fixed, now it supports add_fields=groups_in_count+groups_in

rbac/management/role/view.py
astrozzc: Changed

rbac/management/role/view.py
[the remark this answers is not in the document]
coderbydesign: Hey @astrozzc. It doesn't appear there's an actual web spec standard, but the consensus seems to be to user a comma delimiter to represent an OR operator, and & for an AND operator. So in this case it would probably make more sense to use a comma, to be able to have: `?add_fields=groups_in_count,groups_in`?

https://stackoverflow.com/a/24728298/1331784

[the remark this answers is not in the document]
astrozzc: Right, that's what I tried first using comma, but I keep getting error of invalid input. Let me try to find the issue why comma is not working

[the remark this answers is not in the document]
astrozzc: I found the reason, I put an extra space in between` ?add_fields=groups_in_count, groups_in?`, no wonder it is failing. Updated. Now it looks much better 👍

docs/source/specs/openapi.json
```diff
@@ -2247,6 +2297,55 @@
...
+          },
+          {
+            "$ref": "#/components/schemas/UUID"
+          },
+          {
+            "$ref": "#/components/schemas/Timestamped"
+          },
+          {
+            "type": "object",
+            "required": ["policyCount", "accessCount", "applications", "system", "platform_default"],
```
coderbydesign: Formatting suggestion for consistency.
```suggestion
            "required": [
              "policyCount", 
              "accessCount", 
              "applications", 
              "system", 
              "platform_default"
            ],
```

docs/source/specs/openapi.json
```diff
@@ -984,7 +998,7 @@
             "content": {
               "application/json": {
                 "schema": {
-                  "$ref": "#/components/schemas/RolePagination"
+                  "$ref": "#/components/schemas/RolePaginationV2"
```
coderbydesign: Can we maybe name this in line with the other change, like `RolePaginationDynamic`?

docs/source/specs/openapi.json
```diff
@@ -976,6 +976,20 @@
...
+          },
+          {
+            "name": "add_fields",
+            "in": "query",
+            "required": false,
+            "description": "Parameter for add list of fields to display for roles.",
+            "schema": {
+              "type": "array",
+              "items": {
+                "type": "string"
```
coderbydesign: I think it might be good to add the values allowed here?
```suggestion
                "type": "string",
                "enum": [
                  "groups_in",
                  "groups_in_count"
                ]
```

astrozzc: Thanks, this is really good

docs/source/specs/openapi.json
astrozzc: Sure, fixed

docs/source/specs/openapi.json
astrozzc: 👍

PR #223, opened 2020-02-27
Ensure username_list includes all principal data

coderbydesign:
Previously, `username_list` was being populated with just the first value in
the list from:

```
list(usernames[0].values())
```

Since `usernames` in this case was a list of dictionaries, it only grabbed the
value of the first entry, which results in the `/groups/<uuid>/principals/`
endpoint only ever returning one principal.

This change ensures that we're building a list from all items in the serialized
data.


coderbydesign:
Note: working on trying to add a mock expectation to the specs to exploit this, but the mock ordering is proving to be an issue. Wanted to at least get this PR out for visibility.

PR #225, opened 2020-02-27
Fix count bug on retriving user count from proxy

wcmitchell:
The way we were passing the user list through to _process_data was consuming the user count and not readding it afterwards, resulting in count always coming back None. In testing this seems to clear that up!


wcmitchell:
Oh, that's a good point. Let me see if I can get something that would have caught this put in

wcmitchell:
Ah! The unit test didn't catch it because I'm mocking out the response from request_principals rather than actually creating a proxy instance and calling it or creating a new unittest for the proxy. I'll do the second and get a submission in including that.

PR #226, opened 2020-02-27
Add exclude option for group principals

astrozzc:
Adding an exclude option for group principals so that one api call
should be able to find all principals excluding the ones under this
group.

### DIFF
--- a/rbac/management/group/view.py
+++ b/rbac/management/group/view.py
@@ -40,6 +40,8 @@ from rest_framework.decorators import action
 from rest_framework.filters import OrderingFilter
 from rest_framework.response import Response
 
+from api.common.pagination import StandardResultsSetPagination
+
 
 USERNAMES_KEY = 'usernames'
 ROLES_KEY = 'roles'
@@ -393,19 +395,7 @@ class GroupViewSet(mixins.CreateModelMixin,
             output = GroupSerializer(resp)
             response = Response(status=status.HTTP_200_OK, data=output.data)
         elif request.method == 'GET':
-            principals_from_params = self.filtered_principals(group, request)
-            page = self.paginate_queryset(principals_from_params)
-            serializer = PrincipalSerializer(page, many=True)
-            principal_data = serializer.data
-            if principal_data:
-                username_list = [principal['username'] for principal in principal_data]
-            else:
-                username_list = []
-            proxy = PrincipalProxy()
-            resp = proxy.request_filtered_principals(username_list, account)
-            if isinstance(resp, dict) and 'errors' in resp:
-                return Response(status=resp.get('status_code'), data=resp.get('errors'))
-            response = self.get_paginated_response(resp.get('data'))
+            return self.obtain_principals(request, group)
         else:
             if USERNAMES_KEY not in request.query_params:
                 key = 'detail'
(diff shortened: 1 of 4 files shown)


rbac/management/group/view.py
```diff
@@ -579,3 +568,43 @@ def validate_and_get_exclude_key(self, params):
...
+        else:
+            username_list = []
+
+        # Call BOP to get principals
+        proxy = PrincipalProxy()
+
+        if exclude == 'false':
+            resp = proxy.request_filtered_principals(username_list, request.user.account)
+        else:
+            resp = proxy.request_principals(request.user.account, TOTAL_PRINCIPALS)
```
coderbydesign: Instead of setting `TOTAL_PRINCIPALS` here to `1000`, maybe we should just use `StandardResultsSetPagination.max_limit`? We use that in a few other places.

rbac/management/group/view.py
```diff
@@ -579,3 +568,43 @@ def validate_and_get_exclude_key(self, params):
...
+        principals_from_params = self.filtered_principals(group, request)
+
+        exclude = self.validate_and_get_exclude_key(request.query_params)
+
+        if exclude == 'false':
+            page = self.paginate_queryset(principals_from_params)
+            serializer = PrincipalSerializer(page, many=True)
+        else:
+            # Get all the principals to be excluded
+            serializer = PrincipalSerializer(principals_from_params, many=True)
```
coderbydesign: Does this mean that we won't be paginating the results on our end, either?

rbac/management/group/view.py
```diff
@@ -579,3 +568,43 @@ def validate_and_get_exclude_key(self, params):
...
+            resp = proxy.request_principals(request.user.account, TOTAL_PRINCIPALS)
+
+        # Deal with error
+        if isinstance(resp, dict) and 'errors' in resp:
+            return Response(status=resp.get('status_code'), data=resp.get('errors'))
+
+        if exclude == 'false':
+            response_data = resp.get('data')
+        else:
+            data_without_pagination = [principal for principal in resp.get('data').get('users')
```
coderbydesign: Ah, maybe I spoke too soon. 😄I assume this is what handles the pagination on our end?

coderbydesign: Testing this locally, and the response object is from the `v1` endpoint, so there's no `users` object on the `data`. I think this needs to just be:
```suggestion
            data_without_pagination = [principal for principal in resp.get('data')
```

rbac/management/group/view.py
astrozzc: Good to know! Will change this

rbac/management/group/view.py
astrozzc: Right, we have to exclude all of them

rbac/management/group/view.py
astrozzc: Ah, right, this is the one that handles pagination of the data after exclusion.

astrozzc: hmm, maybe there is problem here. If you look at a few lines above, if exclude is true, it calls "proxy.request_principals" which is calling the v2: https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/principal/proxy.py#L172

coderbydesign: Ah, right. but if `request_filtered_principals` is called, it will use `v1`. So there are two different possible payloads it seems.

astrozzc: Yes, the two lines above `response_data = resp.get('data')` is for dealing this situation

coderbydesign:
@astrozzc should we close this for now based on the conversation in [the ticket](https://projects.engineering.redhat.com/browse/RHCLOUD-4194?focusedCommentId=2189041&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2189041)?

PR #227, opened 2020-02-28
Fix the additional group in spec file

astrozzc:
This AdditionalGroup is an optional return. If the name is set as
required. Empty return of group list would cause error.


coderbydesign:
cc: @catastrophe-brandon

PR #230, opened 2020-03-02
Promote Changes to QA - 3/2/2020

coderbydesign:
**Brings the following PRs into stable to move from CI to QA:**

#225 - Fix count bug on retriving user count from proxy
#227 - Fix the additional group in spec file
#228 - Fix groups_in definition in spec file


coderbydesign:
cc: @catastrophe-brandon

PR #233, opened 2020-03-04
Service to service PSK auth

coderbydesign:
This allows for internal services to make service-to-service requests on behalf
of an account/org, instead of a principal request. These requests will have "admin" access.

In order to do so, the service must supply the following headers. If the headers
are provided, but the request comes through cloud.redhat.com, the identity header
will take precedence, meaning this is only allowed when the request comes directly
to the RBAC service from inside the cluster. 

Here are the headers:
```
x-rh-rbac-psk: (supplied by request from the platform infrastructure team)
x-rh-rbac-account: (the account/org ID the request is being made for)
x-rh-rbac-client-id: (the name of the service making the request, which is tied 
to the supplied PSK. confirm with platform infrastructure)
```

- in `IdentityHeaderMiddleware`, when no identity header is supplied, check for
  system auth headers
- if all headers are present, validate the PSK
- load PSKs from env secret via `SERVICE_PSKS`, which are specific to the client/service
- validate against the primary or alt secret, which is setup to allow for key rotation
- if validation fails, 401
- if validation passes, set the `request.user` object as a system user for the account
- set the tenant for the request based on the `request.user.account` set from 
  the `x-rh-rbac-account` header. we may restrict this per PSK in the future.
- process the request in `RolesTenantMiddleware` if the user is `system`

Documentation on requesting and using PSKs will be added to the platform docs.

**Testing Locally**
---
To test this PR locally, first disable the local setting of the identity header in dev_middleware.py by [commenting this line out](https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/rbac/dev_middleware.py#L53)

Next, start the server with:
```
make serve SERVICE_PSKS='{"catalog": {"secret": "abc123"}}'
```

Verify that you cannot access any endpoints requiring auth:
```
curl http://localhost:8000/api/rbac/v1/roles/ -v
```

Verify that if you pass in the correct headers/values, you _can_ access the endpoint:
```
curl http://localhost:8000/api/rbac/v1/roles/ -v -H 'x-rh-rbac-psk: abc123' -H 'x-rh-rbac-account: 10001' -H 'x-rh-rbac-client-id: catalog'
```

Change the `x-rh-rbac-client-id`, `x-rh-rbac-psk` and `x-rh-rbac-account` header values to see that you should get back a 401 (or 400 with an account that doesn't exist).

You can also send a request _with_ the identity header explicitly in the curl command along with the service-to-service headers to verify that the identity header will take precedence.


rbac/rbac/middleware.py
```diff
@@ -181,16 +218,20 @@ def process_request(self, request):  # noqa: C901
...
             is_admin = json_rh_auth.get('identity', {}).get('user', {}).get('is_org_admin')
-        except (KeyError, JSONDecodeError) as decode_err:
+        except (KeyError, JSONDecodeError):
+            if IdentityHeaderMiddleware._system_auth(request):
+                return
+
             logger.error('Could not obtain identity on request.')
-            raise decode_err
+            HttpResponseUnauthorizedRequest()
+            return
```
astrozzc: is this should be return HttpResponseUnauthorizedRequest()?

astrozzc: But seems the unit test is working.

coderbydesign: This will be hit when there's no identity header, and no valid service PSK. Previously this returned a 500 to the client. This change just ensures it returns a 401.

PR #234, opened 2020-03-04
Config separate

astrozzc:
Rename the roles to reduce confusion that these files are only used for local testing.

Default roles are separated into another repo. They would be deployed separately
but would be consumed by this service as they were in the same path.
    
Any new commits of default roles in this repo would be ignored.


astrozzc:
Tried gitignore the whole folder before which does not work since we want to keep the ones for local testing.

Tried git hooks to ignore the repo which seems not convenient to use. Require manually execution of commands to make it effective.

PR #236, opened 2020-03-06
Add functionality for ordering /group/uuid/roles/ and /group/uuid/pri…

wcmitchell:
…ncipals/


rbac/management/group/view.py
```diff
@@ -430,6 +433,7 @@ def roles(self, request, uuid=None):
         @apiHeader {String} token User authorization token
 
         @apiParam (Path) {String} id Group unique identifier.
+        @apiParam (Query) {String} order_by Determine ordering of returned roles.
```
astrozzc: Should we add this for the principals? And a test similar to the one you added for roles

rbac/management/group/view.py
```diff
@@ -47,6 +47,9 @@
 VALID_EXCLUDE_VALUES = ['true', 'false']
 VALID_GROUP_ROLE_FILTERS = ['role_name', 'role_description']
 VALID_GROUP_PRINCIPAL_FILTERS = ['principal_username']
+ORDERING_PARAM = 'order_by'
+VALID_ROLE_ORDER_FIELDS = ['name', 'modified', 'policyCount']
```
coderbydesign: I'm getting an error when I try to order by `policyCount`, it looks like because it's not actually a field on the role.

coderbydesign: Also, curious if since we have to name the filters prepended with `role_` and `principal_` if we shouldn't do the same here on the ordering for consistency on nested resources? Open to suggestions there.

rbac/management/group/view.py
wcmitchell: Oh! Yes! I definitely need to update the docs. I wasn't thinking of adding a test for the principals, since it would essentially just be directly returning whatever I throw in the mock data, but I'll look back at it and see what I can put together.

rbac/management/group/view.py
wcmitchell: Oh, that's odd. I pulled those from the fields in the model, I thought. I'll look back at it. Thanks!

wcmitchell: Hmm, that's a good thought. The ordering param is generic, though, where the sorting fields will depend on what's being sorted. Maybe a way to import the sorting fields would be best, so they're not sitting in the group codebase, but that doesn't really alter how the ordering_param is set, but may kick the decision can down the road a bit?

astrozzc: Oh, I encounter similar problem before. Our models does not have policyCount, it is annotated in the queryset . Adding annotation to the queryset would solve the problem.

rbac/management/group/view.py
catastrophe-brandon: I think that if we can enumerate the valid values for order_by in the openapi doc that would be great. Also, will there be a parameter to control ascending/descending order?

rbac/management/group/view.py
wcmitchell: Thanks! Just got that added in

rbac/management/group/view.py
wcmitchell: I'll work on adding them into the docs, @catastrophe-brandon. Thanks for pointing that out! I'm working on allowing both ascending and descending order and trusting in the BOP to manage the conversation with IT for us

wcmitchell:
Closing for now, until we can get back sorted principals from BOP

PR #237, opened 2020-03-06
Add username filter to roles

coderbydesign:
Adds a param to filter roles by username on `GET /roles/?username`.

This is only available for org admins, and will override the user in the identity header.


rbac/management/querysets.py
```diff
@@ -87,6 +87,18 @@ def get_role_queryset(request):
         return get_object_principal_queryset(request, scope, Role,
                                              **{'prefetch_lookups_for_ids': 'access',
                                                 'prefetch_lookups_for_groups': 'policies__roles'})
+
+    username = request.query_params.get('username')
+    if username:
+        decoded = request.user.identity_header.get('decoded', {})
+        identity_username = decoded.get('identity', {}).get('user', {}).get('username')
+        if username != identity_username and not request.user.admin:
+            return Role.objects.none()
```
astrozzc: I am curious why we return this none() instead of warning about permission error?

coderbydesign: Hey @astrozzc - I tend to agree that this may be a case where we'd want to return a 403, but I didn't want to change the existing pattern of behavior (similar to the way that [groups has been setup](https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/querysets.py#L71)) without realizing the potential side-effects.

astrozzc: Sure 👍

PR #238, opened 2020-03-09
Enable remediations allowance for creating roles

coderbydesign:


coderbydesign:
@lphiri yep! I added a story for that a while ago and it was tagged tech-debt in previous planning: https://projects.engineering.redhat.com/browse/RHCLOUD-3791

PR #239, opened 2020-03-09
Role group

astrozzc:
Supports filter groups by a list of role name. Either fiter by any
or all of the role names in the list (default is any).
    
usage: e.g. ?role_names=roleA,roleB&role_discriminator=all


rbac/management/group/view.py
```diff
@@ -59,12 +61,37 @@ def username_filter(queryset, field, value):
...
+        roles_list = value.split(',')
+
+        discriminator = validate_and_get_key(
+            self.request.query_params,
+            ROLE_DISCRIMINATOR_KEY,
+            VALID_ROLE_ROLE_DISCRIMINATOR,
+            'any')
+
+        if discriminator == 'any':
+            return queryset.filter(policies__roles__name__in=roles_list)
```
coderbydesign: Just curious, is this expected to be a partial match? Other filters work that way, but the expectation of this may be different?

astrozzc: Good point. From our conversation, it sounds like they want an exact match, but let me confirm with their use case.

coderbydesign: Sounds good! I think the exception is `/principals/?usernames=` which is an exact match on a list.

astrozzc: Confirmed with them and they want an exact match but case insensitive.

I have updated the code and test to be case insensitive

PR #241, opened 2020-03-12
Fix spec for /roles/{uuid}/access

astrozzc:
The spec is not configured right. Extra layer of data should be removed.


catastrophe-brandon:
Thanks for the quick action on this! At a glance this looks reasonable.

catastrophe-brandon:
Looks like this might need a merge or a rebase :)

PR #242, opened 2020-03-12
Add handling for parameter to allow ordering of group roles

wcmitchell:
Splitting this out from the principal sorting pending updates to allow that

### DIFF
--- a/rbac/management/group/view.py
+++ b/rbac/management/group/view.py
@@ -35,6 +35,7 @@ from management.principal.proxy import PrincipalProxy
 from management.principal.serializer import PrincipalSerializer
 from management.querysets import get_group_queryset, get_object_principal_queryset
 from management.role.model import Role
+from management.role.view import RoleViewSet
 from rest_framework import mixins, serializers, status, viewsets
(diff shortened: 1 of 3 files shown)


rbac/management/group/view.py
```diff
@@ -556,7 +571,14 @@ def obtain_roles(self, request, group):
 
         filtered_roles = self.filtered_roles(roles, request)
 
-        return [RoleMinimumSerializer(role).data for role in filtered_roles]
+        annotated_roles = filtered_roles.annotate(policyCount=Count('policies', distinct=True))
+
+        if ORDERING_PARAM in request.query_params:
+            ordered_roles = self.order_queryset(annotated_roles, VALID_ROLE_ORDER_FIELDS,
+                                                request.query_params.get(ORDERING_PARAM))
+            if ordered_roles:
```
astrozzc: Since you have added validation error in the order_queryset method, maybe we could get rid of this if statement?

wcmitchell: Good point!

PR #243, opened 2020-03-13
Add a uuid filter to GETs on the /groups/ endpoint

wcmitchell:

### DIFF
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -224,8 +224,7 @@
               "type": "array",
               "items": {
                 "type": "string"
-              },
-              "collectionFormat": "csv"
+              }
             },
             "explode": false,
             "style": "form"
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -216,6 +216,21 @@
             }
           },
           {
+            "name": "uuid",
+            "in": "query",
+            "description": "A list of UUIDs to filter listed groups.",
+            "required": false,
+            "schema": {
+              "type": "array",
+              "items": {
+                "type": "string"
+              },
+              "collectionFormat": "csv"
+            },
+            "explode": false,
+            "style": "form"
+          },
+          {
             "$ref": "#/components/parameters/OrderByFilter"
           }
         ],
--- a/rbac/management/group/view.py
+++ b/rbac/management/group/view.py
@@ -17,6 +17,7 @@
 
 """View for group management."""
 import logging
+from uuid import UUID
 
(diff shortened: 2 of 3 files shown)


docs/source/specs/openapi.json
```diff
@@ -215,6 +215,21 @@
...
+            "name": "uuid",
+            "in": "query",
+            "description": "A list of UUIDs to filter listed groups.",
+            "required": false,
+            "schema": {
+              "type": "array",
+              "items": {
+                "type": "string"
+              },
+              "collectionFormat": "csv"
```
coderbydesign: It looks like this is a 2.0 thing, not compatible with 3.0? https://swagger.io/docs/specification/describing-parameters/

wcmitchell: Oh, yep. Dang. I had hopes it'd help clarify, but I'll pull it out

rbac/management/group/view.py
```diff
@@ -59,12 +59,19 @@ def username_filter(queryset, field, value):
         filtered_set = queryset.filter(**filters)
         return filtered_set | Group.platform_default_set()
 
+    def uuid_filter(queryset, field, values):
+        """Filter for group uuid lookup."""
+        filters = {f'{field}__in': values.split(',')}
```
coderbydesign: We may want a UUID validation here? If I give an invalid UUID as one of the values in the query param, I get a 500.

wcmitchell: 👍

rbac/management/group/view.py
```diff
@@ -59,12 +59,19 @@ def username_filter(queryset, field, value):
...
+        filtered_set = queryset.filter(**filters)
+        return filtered_set
+
     name = filters.CharFilter(field_name='name', lookup_expr='icontains')
     username = filters.CharFilter(field_name='principals', method=username_filter)
+    uuid = filters.CharFilter(field_name='uuid', method=uuid_filter)
```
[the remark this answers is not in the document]
wcmitchell: I went ahead and threw this in. I don't think it'll affect things much as they are now, but it seems to be what django-filter recommends, and doesn't seem a negative in any way I can see.

rbac/management/group/view.py
wcmitchell: Added in a simple check that'll throw back a 400 and an 'invalid uuid' message.

rbac/management/group/view.py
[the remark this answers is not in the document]
astrozzc: Right, in my case I have to use the self. And it also remove the annoying lint warning 😁

PR #245, opened 2020-03-17
Ensure /role/?username= returns count data

coderbydesign:
A bug was discovered where `accessCount` and `policyCount` data was missing from
the dataset when `?username=` was supplied on the `/role/` GET endpoint.

This is because of the fact that the queryset is different based on whether or
not that username param is supplied. The annotations for those two fields need
to be added.

### DIFF
--- a/rbac/management/querysets.py
+++ b/rbac/management/querysets.py
@@ -78,15 +78,22 @@ def get_group_queryset(request):
     return Group.objects.none()
 
 
+def annotate_roles_with_counts(queryset):
+    """Annotate the queryset for roles with counts."""
+    return queryset.annotate(policyCount=Count('policies', distinct=True),
+                             accessCount=Count('access', distinct=True))
+
+
 def get_role_queryset(request):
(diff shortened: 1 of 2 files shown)


rbac/management/querysets.py
```diff
@@ -95,9 +95,12 @@ def get_role_queryset(request):
         if username != identity_username and not request.user.admin:
             return Role.objects.none()
         else:
-            return get_object_principal_queryset(request, PRINCIPAL_SCOPE, Role,
-                                                 **{'prefetch_lookups_for_ids': 'access',
-                                                    'prefetch_lookups_for_groups': 'policies__roles'})
+            queryset = get_object_principal_queryset(request, PRINCIPAL_SCOPE, Role,
```
astrozzc: Good fix! This reminds me the above line 87, if scope is principal, maybe we should add this annotation too?

coderbydesign: Done!

PR #248, opened 2020-03-17
Support passing sort_order as parameter

astrozzc:
BOP now supports sort order parameter. This PR adds that parameter
when sending request to BOP.

Either ascending or desending is supported. If no order is specified,
ascending is default.

### DIFF
--- a/rbac/management/principal/view.py
+++ b/rbac/management/principal/view.py
@@ -16,6 +16,7 @@
 #
 
 """View for principal management."""
+from management.utils import validate_and_get_key
 from rest_framework import status
 from rest_framework.response import Response
 from rest_framework.views import APIView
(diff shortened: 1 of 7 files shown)


rbac/management/principal/view.py
```diff
@@ -25,6 +26,8 @@
 from ..permissions.admin_access import AdminAccessPermission
 
 USERNAMES_KEY = 'usernames'
+SORTORDER_KEY = 'sort_order'
+VALID_SORTORDER_VALUE = ['asc', 'des']
```
coderbydesign: I believe the keyword for most database order modifiers should be `desc` in this case? It might make it more intuitive.

coderbydesign: Ah, it looks like that's what it is in the BOP, eh?

astrozzc: Right, there is enum in BOP before my changes: https://github.com/RedHatInsights/backoffice-proxy/blob/master/spec.json#L456-L457

I agree that desc is more intuitive. I think it might be worth changing it.

PR #249, opened 2020-03-18
Add active flag to principal response

coderbydesign:
This adds the `is_active` flag from the BOP to the `/principals/` GET response:

```json
{
  "meta": {
    "count": 1
  },
  "links": {
    "first": "/api/rbac/v1/principals/?limit=1&offset=0",
    "next": null,
    "previous": null,
    "last": null
  },
  "data": [
    {
      "username": "red_hat_user",
      "email": "red_hat_user@redhat.com",
      "first_name": "Red Hat",
      "last_name": "User",
      "is_active": true
    }
  ]
}
```

### DIFF
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -1928,6 +1928,9 @@
           "last_name": {
             "type": "string",
             "example": "Smith"
+          },
+          "is_active": {
+            "type": "boolean"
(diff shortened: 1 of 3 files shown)


coderbydesign:
Thanks you, @astrozzc for catching that! I'll update the spec.

PR #250, opened 2020-03-18
Promote Changes to QA - 3/18/2020

coderbydesign:
**Brings the following PRs into stable to move from CI to QA:**

#233 - Service to service PSK auth
#234 - Config separate
#235 - Document inverse `order_by` filter in description
#237 - Add username filter to roles
#238 - Enable remediations allowance for creating roles
#239 - Role group
#241 - Fix spec for /roles/{uuid}/access
#243 - Add a uuid filter to GETs on the /groups/ endpoint 
#245 - Ensure /role/?username= returns count data
#248 - Support passing sort_order as parameter


astrozzc:
The failure check is due to some version warning, which I think it should be fine for now?

We have open PRs to bump the django version, which I think we could try to deal with after the next release?

coderbydesign:
@astrozzc - Yeah I disabled that check for `stable`, so I'm not sure why it's running currently, but it shouldn't affect us moving.

PR #251, opened 2020-03-19
Group principal sort

wcmitchell:
Taking another swing at this. Letting BOP handle username ordering, but trying to order by the other required fields locally, post BOP. Not sure how this will interact with paging or performance, could use some input there.


rbac/management/group/view.py
```diff
@@ -573,6 +584,22 @@ def roles(self, request, uuid=None):
 
         return Response(status=status.HTTP_200_OK, data=response_data.data)
 
+    def ordered_principals(self, principals, order_field):
+        """Return principals ordered by order_field."""
+        all_fields = VALID_PRINCIPAL_ORDER_FIELDS + ['-' + field for field in VALID_PRINCIPAL_ORDER_FIELDS]
+        if order_field.lower() in all_fields:
```
astrozzc: Overall LGTM! Maybe we could remove this and the VALID_PRINCIPAL_ORDER_FIELDS? Because 1. the 'sorted' would report KeyError if it is not a valid key to be sorted; 2. that would support sorting by all available fields including the 'is_active' added by Keith. Or we should have control over the fields to be sorted? Any thoughts?

rbac/management/group/view.py
```diff
@@ -573,6 +584,22 @@ def roles(self, request, uuid=None):
...
+            if order_field.startswith('-'):
+                reverse = True
+                order_field = order_field[1:]
+            else:
+                reverse = False
+            try:
+                return sorted(principals, key=operator.itemgetter(order_field), reverse=reverse)
+            except KeyError:
+                key = 'detail'
+                message = f'Failed ot sort on {order_field}'
```
astrozzc: Nic: to 😁

wcmitchell: Hah! Good eyes! Thanks!

rbac/management/group/view.py
wcmitchell: Hmm, these are some good points. I'll poke around a bit and see what I can do with it

wcmitchell:
Oh! Right! Good eye @coderbydesign, I'd had a check previously, but pulled it out for the check in ordered_principals(). With that gone, so's the check. Easy fix.

rbac/management/group/view.py
```diff
@@ -451,7 +451,13 @@ def principals(self, request, uuid=None):
             else:
                 username_list = []
             proxy = PrincipalProxy()
-            resp = proxy.request_filtered_principals(username_list, account)
+            all_valid_fields = VALID_PRINCIPAL_ORDER_FIELDS + ['-' + field for field in VALID_PRINCIPAL_ORDER_FIELDS]
+            if request.query_params.get(ORDERING_PARAM):
+                sort_field = validate_and_get_key(request.query_params, ORDERING_PARAM, all_valid_fields, 'username')
+                sort_order = 'desc' if sort_field == '-username' else 'asc'
```
coderbydesign: I do believe the valid value here is `des` for the BOP? https://backoffice-proxy-insights-services.ext.us-east.aws.preprod.paas.redhat.com/docs/#/
```suggestion
                sort_order = 'des' if sort_field == '-username' else 'asc'
```

I know @astrozzc ran into this in #248 as well.

PR #263, opened 2020-04-07
Fix group filter with invalid username

astrozzc:
When invalid username is given, default group will still be returned,
which should be fixed. The username have to be validated.

### DIFF
--- a/rbac/management/querysets.py
+++ b/rbac/management/querysets.py
@@ -23,6 +23,7 @@ from management.policy.model import Policy
 from management.role.model import Access, Role
 from management.utils import (APPLICATION_KEY,
                               access_for_principal,
+                              get_principal,
                               get_principal_from_request,
                               groups_for_principal,
                               policies_for_principal,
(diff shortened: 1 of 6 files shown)


rbac/management/querysets.py
```diff
@@ -65,15 +66,11 @@ def get_group_queryset(request):
...
     username = request.query_params.get('username')
     if username:
-        decoded = request.user.identity_header.get('decoded', {})
-        identity_username = decoded.get('identity', {}).get('user', {}).get('username')
-        if username != identity_username and not request.user.admin:
-            return Group.objects.none()
-        else:
-            return Group.objects.filter(principals__username=username)
+        get_principal(username, request.user.account)
+        return Group.objects.filter(principals__username=username) | Group.platform_default_set()
```
coderbydesign, line 70: Do we want to be removing the check here to verify that if the `username` doesn't match, the user must be an admin? This is exposing more data than it does on `master` currently. If I'm not an admin, currently I can't filter by another username. Is this because we've opened up the `/groups/` endpoint in general to everyone now?

astrozzc: right, that's my consideration, since we don't have restriction on the `/groups/` endpoint. Customers, including non-admins (I tested it, if I am not wrong?), could use `/groups/{uuid}/` to see all the principals inside one group. So it would seem weird if we restrict here only allowing admin, then non-admin users could see all group principals, but can't filter it.

coderbydesign: Yeah, it looks like this [1] is explicitly allowing access on `/groups/` but then if you use the principal scope, if you have access to the group you can access `/groups/{uuid}/`. 

So since the filter is only on `/groups/`, I'd say it's reasonable to allow that filter for anyone, since it will only ever be a subset of what `/groups/` would be anyway?

[1] 5a0f198c-7dfb-41a1-bca5-9862a3789d1d

PR #265, opened 2020-04-08
Performance improvements

j00bar:
There are two major enhancements in this PR:

1. The middleware processing requests comprises and extraordinarily large amount of the processing time, largely due to many repetitive single-row requests to the database for `User`, `Tenant`, and `Principal` records.
  * `Tenant` records are cached in-memory, given they are only really used for determining schema names and availabilities.
  * `User` has been converted from a `Model` to a simply `object` class.
  * Whereas there were two custom middleware before, one for Tenant handling and one for Authentication handling, they have been combined.
  * The unit test suite has been updated accordingly.
2. The `AccessView` API resource has been wrapped in a Redis-based caching layer.
  * The resolved access policy for a `Principal` per-app is cached as serialized JSON in Redis.
  * Any writes to `Group`, `Role`, or `Policy` objects eject the caches for the `Principal` objects affected.
  * The caching layer is written to fall back to the database if Redis is unavailable.

On my laptop, with a given randomly generated dataset, I'm seeing average service times against the `access` endpoint with the `master` branch of 38ms. With this branch, the same dataset offers average service times of 13ms.


j00bar:
Looking over the coverage report, it's clear some of the signal-handler logic isn't being fired. We should write tests specifically to ensure cache invalidation.

rbac/management/cache.py
```diff
@@ -0,0 +1,84 @@
...
+    def save_policy(self, uuid, application, policy):
+        """Write the policy for a given user for a given app to Redis."""
+        if not settings.ACCESS_CACHE_ENABLED:
+            return
+        try:
+            logger.info('Caching policy for uuid %s', uuid)
+            with self.connection.pipeline() as pipe:
+                pipe.hset(
+                    self.key_for(uuid),
+                    application,
```
coderbydesign, line 73: Hey @j00bar, just wanted to note here based on our conversation yesterday in planning, that we will also need to account for `/access/?application=` with an empty application param value in order to return all permissions for a principal. 

In the upcoming sprint we'll also be supporting multiple applications `/access/?application=approval,catalog` but it looks like that may be accounted for here already since the key is based on the param value?

rbac/management/permissions/role_access.py
```diff
@@ -40,5 +40,5 @@ def has_permission(self, request, view):
             role_write = request.user.access.get('role', {}).get('write', [])
             if role_write:
                 return True
-
+        # raise RuntimeError(f'{request.user.__dict__}, {request.__dict__}')
```
coderbydesign, line 43: Is this a remnant of testing?

rbac/management/principal/proxy.py
```diff
@@ -113,8 +113,14 @@ def _get_proxy_service(self):  # pylint: disable=no-self-use
         }
         return proxy_conn_info
 
-    def _request_principals(self, url, account=None, account_filter=False, method=requests.get, params=None, data=None):
+    def _request_principals(self, url, account=None, account_filter=False, method=requests.get,  # noqa: C901
+                            params=None, data=None):
         """Send request to proxy service."""
+        if settings.BYPASS_BOP_VERIFICATION:
```
coderbydesign, line 119: This is a very helpful addition!

rbac/management/cache.py
j00bar: We can definitely amend this for those cases. Redis has `HMGET` and `HGETALL`, but the tricky thing is that not all of the applications data will necessarily be cached yet.

It feels like whichever one - this or the other PR - merges first, we need to rebase and amend there?

rbac/management/permissions/role_access.py
j00bar: Yarr. I'll remove.

rbac/management/principal/proxy.py
j00bar: 🤜💥🤛

j00bar:
Moar tests written. (And bugs found and squashed)

rbac/management/utils.py
```diff
@@ -43,19 +43,22 @@ def validate_psk(psk, client_id):
...
 
     try:
         principal = Principal.objects.get(username__iexact=username)
     except Principal.DoesNotExist:
-        key = 'detail'
-        message = 'No data found for principal with username {}.'.format(username)
-        raise serializers.ValidationError({key: _(message)})
+        if not qs_user:
+            # If we're lookup up ourselves, create the Principal
+            principal = Principal.objects.create(username=username)
```
j00bar, line 57: You need `django.db.transaction.atomic` here.

dkuc:
Looking forward to this. Calls to rbac are adding about 120ms to advisor currently.
``` bash
time curl -s -H 'x-rh-identity:eyJlbnRpdGxlbWVudHMiOnsiaW5zaWdodHMiOnsiaXNfZW50aXRsZWQiOnRydWV9LCJjb3N0X21hbmFnZW1lbnQiOnsiaXNfZW50aXRsZWQiOmZhbHNlfSwiYW5zaWJsZSI6eyJpc19lbnRpdGxlZCI6ZmFsc2V9LCJzZXR0aW5ncyI6eyJpc19lbnRpdGxlZCI6dHJ1ZX0sIm9wZW5zaGlmdCI6eyJpc19lbnRpdGxlZCI6dHJ1ZX0sInNtYXJ0X21hbmFnZW1lbnQiOnsiaXNfZW50aXRsZWQiOmZhbHNlfSwic3Vic2NyaXB0aW9ucyI6eyJpc19lbnRpdGxlZCI6dHJ1ZX0sIm1pZ3JhdGlvbnMiOnsiaXNfZW50aXRsZWQiOmZhbHNlfX0sImlkZW50aXR5Ijp7ImludGVybmFsIjp7ImF1dGhfdGltZSI6MCwib3JnX2lkIjoiNTk1NTU4OCJ9LCJhY2NvdW50X251bWJlciI6IjU0NDgzNTciLCJhdXRoX3R5cGUiOiJqd3QtYXV0aCIsInVzZXIiOnsiZmlyc3RfbmFtZSI6IlJpY2FyZG8gQWxvbnNvIGRvcyBTYW50b3MiLCJpc19hY3RpdmUiOnRydWUsImlzX2ludGVybmFsIjpmYWxzZSwibGFzdF9uYW1lIjoiTmFzY2ltZW50byIsImxvY2FsZSI6ImVuX1VTIiwiaXNfb3JnX2FkbWluIjp0cnVlLCJ1c2VybmFtZSI6InJpY2FyZG9hbG9uc29zIiwiZW1haWwiOiJyYWxvbnNvQHJlZGhhdC5jb20ifSwidHlwZSI6IlVzZXIifX0='  -H 'Content-Type:application/json'  http://rbac.rbac-prod.svc:8080/api/rbac/v1/access/?application=insights > /dev/null

real	0m0.121s
user	0m0.005s
sys	0m0.005s
```

j00bar:
@coderbydesign Rebased. I'll figure out that failing test. I'm looking at the multi-app capabilities, and it actually should work just fine with multi-app querysets... it will just be a little memory inefficient for Redis. If you ask for `app1,app2`, it will store a key for those two apps together. If you ask for `app1`, it will store `app1` separately from the `app1,app2` lookup. Does that sound alright with you?

j00bar:
I think we're okay to go.

coderbydesign:
> If you ask for app1,app2, it will store a key for those two apps together. If you ask for app1, it will store app1 separately from the app1,app2 lookup. Does that sound alright with you?

@j00bar yes, that makes sense to me!

rbac/management/principal/proxy.py
```diff
@@ -113,8 +113,14 @@ def _get_proxy_service(self):  # pylint: disable=no-self-use
...
+            to_return = []
+            for principal in data['users']:
```
coderbydesign, line 121: @j00bar - Testing this locally, and noticed one minor thing. If we call `_request_principals()` from `request_filtered_principals()` this works as expected, because `payload['users']` will always exist.

However, for the `/api/rbac/v1/principals/` endpoint, we call this from `request_principals()`, which does not filter the users, [so will never have `payload['users']` populated](https://github.com/RedHatInsights/insights-rbac/pull/265/files#diff-42895888fc7b254a2fa368acc404c6feR204). 

Maybe we just need a check to ensure `data['users']` exists before appending the principals? Also, for `/api/rbac/v1/principals/`, [the view is expecting a `status_code` on the response object](https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/principal/view.py#L156), so we may want to just add `status=200` to that dictionary we return?

Issue #264, opened 2020-04-08
Access API is returning other applications and not only the requested 


josejulio:
**Describe the bug**
The `/access/` endpoint is not doing an exact match with the `application` param, we noticed this when we decided to rename from `custom-policies` to `policies`. We noticed that the call:
`/api/rbac/v1/access/?application=policies` still returned the `custom-policies` permission.

**To Reproduce**
1. Go to https://ci.cloud.redhat.com/ and login
2. Go to https://ci.cloud.redhat.com/api/rbac/v1/access/?application=po
3. See that it returns every permission of the application that have `po` on their name

```
{
    "meta": {
        "count": 7,
        "limit": 1000,
        "offset": 0
    },
    "links": {
        "first": "/api/rbac/v1/access/?application=po&limit=1000&offset=0",
        "next": null,
        "previous": null,
        "last": "/api/rbac/v1/access/?application=po&limit=1000&offset=0"
    },
    "data": [
        {
            "permission": "catalog:portfolios:read",
            "resourceDefinitions": [
                {
                    "attributeFilter": {
                        "key": "scope",
                        "value": "group",
                        "operation": "equal"
                    }
                }
            ]
        },
        ...
    ]
}
```

**Expected behavior**
I would expect the `application` to be an exact match of the application that we require, as show in the [openapi.json](https://ci.cloud.redhat.com/api/rbac/v1/openapi.json) doc

> The application name to obtain access for the principal

**Context (please complete the following information):**
 - OS: Fedora 
 - Browser: Chrome
 - Commit Hash from status endpoint: `c107d88022cf94b7744da5d35ec7ca98e0f25f75`

coderbydesign:
Hi @josejulio - we actually have [a bug](https://projects.engineering.redhat.com/browse/RHCLOUD-5269) logged for this currently, and will be addressing it soon. I'll update this issue once a PR is in. Thank you!

petracihalova:
closing the issue - fixed by #266

PR #266, opened 2020-04-09
Ensure the /access/?application= check is on permission name only

coderbydesign:
There's currently a bug where the `/?application=` param value is being compared
against the entire permission string. This is causing scenarios where resource types
or actions in the permission can cause it to match on the application name since
it's a partial match on the entire permission.

Here is [an issue](https://github.com/RedHatInsights/insights-rbac/issues/264) that was logged against this as well.

For instance, a permission `catalog:*:foobar` would be returned with a query:
`/access/?application=foo` when it should only respect the `catalog` substring
as the partial being matched against.

Also, the same permission would return for a query: `/access/?application=cat`.

This does two things:
- restricts the matching to the application name of the permission only
- requires an exact match, to align with the spec.

**Testing Locally**
- checkout master
- modify a permission on a role that would normally return on a given ?application=
  query so that the it looks something like `app-name:foo:bar`
- query `/access/?application=foo` and see that the permission returns
- query `/access/?application=app` and see that the permission returns
- now checkout this feature branch and run the same query against `foo` and `app`.
  you should not see the permission returned, but `/access/?application=app-name`
  should work.

### DIFF
--- a/tests/management/access/test_view.py
+++ b/tests/management/access/test_view.py
@@ -87,6 +87,8 @@ class AccessViewTests(IdentityRequest):
         url = reverse('role-list')
         client = APIClient()
         response = client.post(url, test_data, format='json', **self.headers)
+        print("RESPONSE: ")
+        print(response.data)
         self.assertEqual(response.status_code, status.HTTP_201_CREATED)
         return response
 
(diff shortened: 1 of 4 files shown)


tests/management/access/test_view.py
```diff
@@ -87,6 +87,8 @@ def create_role(self, role_name, in_access_data=None):
         url = reverse('role-list')
         client = APIClient()
         response = client.post(url, test_data, format='json', **self.headers)
+        print("RESPONSE: ")
+        print(response.data)
```
astrozzc, line 91: Overall LGTM! You forget to remove this debug code 😁

PR #273, opened 2020-04-17
Add ability to search for principal by exact e-mail

wcmitchell:
I managed to test this locally using a dev_user with the insights-qa account number and valid client-id and token as env vars.

### DIFF
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -105,6 +105,15 @@
               "type": "string",
               "enum": ["asc", "desc"]
             }
+          },
+          {
+            "name": "email",
+            "in": "query",
+            "description": "Exact e-mail address of principal to search for",
+            "required": false,
+            "schema": {
+              "type": "string"
+            }
           }
         ],
         "responses": {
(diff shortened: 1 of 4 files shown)


wcmitchell:
> 
> 
> I noticed one small issue while testing. In [this conditional](https://github.com/RedHatInsights/insights-rbac/pull/273/files#diff-d79aff1aab75d2782ee2ac72ab0782cdR137-R141), we don't account for the payload structure from `/usersBy` which is a list (of one or none), so the count is null: `"meta": {"count": null}`.
> 
> This may be a naive approach, but something like this may work, assuming we won't get the count back from the BOP. The only gotcha would be if there's a response where we limit the number of results returned for pagination, in which case we wouldn't want to rely on the length of the list locally. Maybe we can check for the count and do this as a last resort?
> 
> ```
>             if isinstance(data, dict):
>                 count = data.get('userCount')
>                 data = data.get('users')
>             elif isinstance(data, list):
>                 count = len(data)
>             else:
>                 count = None
> ```

I actually think this will be a reasonable fix. The only instance in which we should get a list back currently is in a findusers request by email. Since we're only allowing exact e-mail match, we should only get back one user per-account. In the future, should we move to a partial match, we'll likely need to revisit this.

PR #274, opened 2020-04-20
Fix tenant creation exception

astrozzc:
Can't create tenant outside the public schema is thrown, which is caused
concurrent requests: when a bunch of requests try to get tenant, it does
not exist, so they try to create one but only one succeeded, others throw
exceptions not handled.

This PR tries to fix that issue by using get_or_create instead of create.

### DIFF
--- a/rbac/rbac/middleware.py
+++ b/rbac/rbac/middleware.py
@@ -93,11 +93,8 @@ class RolesTenantMiddleware(BaseTenantMiddleware):
             user = User.objects.get(username=request.user.username)
             tenant = user.tenant
             schema_name = tenant.schema_name
-        try:
-            tenant = model.objects.get(schema_name=schema_name)
-        except model.DoesNotExist:
-            tenant = model(schema_name=schema_name)
-            tenant.save()
+
+        tenant, _ = model.objects.get_or_create(schema_name=schema_name)
         return tenant
 
 
@@ -122,9 +119,9 @@ class IdentityHeaderMiddleware(MiddlewareMixin):  # pylint: disable=R0903
         """
         schema_name = create_schema_name(account)
         try:
-            with transaction.atomic():
-                tenant = Tenant(schema_name=schema_name)
-                tenant.save()
+            tenant, created = Tenant.objects.get_or_create(schema_name=schema_name)
+
+            if created:
                 logger.info('Created new tenant from account_id %s.', account)
                 seed_roles(tenant=tenant, update=False)
                 seed_group(tenant=tenant)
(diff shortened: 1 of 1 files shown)


rbac/rbac/middleware.py
```diff
@@ -123,11 +120,12 @@ def _create_tenant(account):
         schema_name = create_schema_name(account)
         try:
             with transaction.atomic():
-                tenant = Tenant(schema_name=schema_name)
-                tenant.save()
-                logger.info('Created new tenant from account_id %s.', account)
-                seed_roles(tenant=tenant, update=False)
-                seed_group(tenant=tenant)
+                tenant, created = Tenant.objects.get_or_create(schema_name=schema_name)
```
catastrophe-brandon, line 123: Noob question: does get_or_create raise any kind of errors? This is more for my knowledge than an actual critique of any kind.

astrozzc: I only know that if the key specified is not unique, e.g., several tenants with schema_name 'acct1234' exist, it will return MultipleObjectsReturned. But this would not happen in our case since this is the primary key for tenant.

rbac/rbac/middleware.py
```diff
@@ -253,10 +250,8 @@ def process_request(self, request):  # noqa: C901
                                                              request)
 
             with tenant_context(tenant):
-                try:
-                    Principal.objects.get(username__iexact=username)
-                except Principal.DoesNotExist:
-                    Principal.objects.create(username=username)
+                created = Principal.objects.get_or_create(username=username)[1]
```
coderbydesign, line 253: It might be more clear to do something like this?
```suggestion
                _, created = Principal.objects.get_or_create(username=username)
```

rbac/rbac/middleware.py
```diff
@@ -123,11 +119,12 @@ def _create_tenant(account):
         schema_name = create_schema_name(account)
         try:
             with transaction.atomic():
```
wcmitchell, line 121: From the docs, it seems that get_or_create is atomic, so maybe we don't need this? Not sure. https://docs.djangoproject.com/en/3.0/ref/models/querysets/#get-or-create

rbac/rbac/middleware.py
astrozzc: 👍

rbac/rbac/middleware.py
astrozzc: Yea, I think you are right!

astrozzc:
> 👍 Looks good, but per our meeting with @j00bar we might want to see if the changes in #265 resolve this first?

Sure, sounds good.

astrozzc:
> 👍 Looks good, but per our meeting with @j00bar we might want to see if the changes in #265 resolve this first?

I just checkout his branch and added the principal get_or_create, the tests all pass

PR #283, opened 2020-05-01
WIP: Abstract seeding into a module for reuse

coderbydesign:
**ToDo:**
- [ ] update tests

We need to decouple the group and role seeding from the application start process
in deployed environments, and tie them to our `rbac-config-sync` job: https://github.com/RedHatInsights/e2e-deploy/blob/master/templates/rbac/config-sync.yaml
in order to not run seeding when a pod comes up, but instead only when we trigger
a change based on our config updating.

This will help to reduce the pod startup time, and avoid unnecessary runs of the
seeding when there are no changes being made.

By default, we'll keep this enabled unless `ROLE_SEEDING_ENABLED` or `GROUP_SEEDING_ENABLED`
are set to `False` respectively.

Adding the `./rbac/manage.py seeds` command will allow us to run the seeds on
demand, in the case of a deployed instance where we wish to disable default seeding
and sync them with config changes, where we can update the config and run seeding
ad-hoc.

**Local Testing**
- to test that the current behavior is unchanged, ensure both `ROLE_SEEDING_ENABLED`
  and `GROUP_SEEDING_ENABLED` are unset locally. in this case, seeds should run
  when you start your local server. you should see seeds run in your console.
- to test that seeding can be disabled by default, set `ROLE_SEEDING_ENABLED`
  and `GROUP_SEEDING_ENABLED` both to `False` and restart your local server. you
  should not see seeds run in the console.
- to test the seeds can be triggered manually, from the command like run: `./rbac/manage.py seeds`
  you should see seeds run in the console.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -83,6 +83,20 @@ If a docker container running Postgres is not feasible, it is possible to run Po
 
     make run-migrations
 
+Seeds
+^^^^^
+
+Default roles and groups are automatically seeded when the application starts by default unless either of the following environment variables are set to 'False' respectively: ::
+
+  ROLE_SEEDING_ENABLED
+  GROUP_SEEDING_ENABLED
+
+Locally these are sourced from `/rbac/management/role/definitions/*.json`, while the config maps in deployed instances are source from our `RBAC config repo`_. **If any changes to default roles/groups are required, they should be make there.**
+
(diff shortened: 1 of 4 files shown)


coderbydesign:
Closing in favor of #287

PR #286, opened 2020-05-07
Tune migrations and seeds

coderbydesign:
**Run migrations in parallel**
---
2169573
By default, this will enable our migrations to run in 2 parallel processes, which
will help speed up our database seeding process as more tenants are created.

https://django-tenant-schemas.readthedocs.io/en/latest/use.html?highlight=connection%20pool#migrate-schemas-in-parallel

**Ensure we close thread connections while seeding**
---
9748494
We need to ensure we close the DB connection(s) which end up being opened with
`ThreadPoolExecutor()` explicitly, otherwise they are not released, and we end
up hitting our PostgreSQL limit in environments where this is set to the default (100).

What happens is that when we run seeds, each tenant/thread opens a new database
connection. When the number of tenants exceeds the number of connections, the async
processing can quickly exhaust all available database connections, since they
are not immediately released.

We need to ensure that after each thread is completed, we close the connections for
that thread. This should resolve the issue, but we can also limit this further
by setting an explicit `max_workers` value to limit the concurrent threads being
run for seeding. When this is _not_ supplied, it will use the default value for
the concurrent threads [1].

[1] https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor

**Testing Locally**
---
You can test these changes the following ways:

Migrations:
- Run `make run-migrations` or the migrate command manually, with `--executor=parallel`
appended, and see that the output should reflect two schemas being migrated at once.

Seeding:
- You will want to create multiple tenants in order to see the full effect of the
before/after. 
- If you have ~ 100 tenants, you'll be able to reproduce the connection failure 
by just starting the server and watching the seeds run. It should run out of connections.
You should do this on `master`.
- You can monitor your connections in `psql` with the query: `SELECT sum(numbackends) FROM pg_stat_database;`
while seeds are running, to see the connections open and eventually become orphaned.
- Once you checkout this feature branch, start the server again, running seeds,
and again monitor your local database connections. They should remain low, and constant
regardless of the number of tenants you have.
- You may also optionally set `MAX_SEED_THREADS` in your `.env` file, to see how
that affects concurrency.

### DIFF
--- a/rbac/rbac/settings.py
+++ b/rbac/rbac/settings.py
@@ -348,6 +348,7 @@ ACCESS_CACHE_ENABLED = ENVIRONMENT.bool('ACCESS_CACHE_ENABLED', default=True)
 # Seeding Setup
 ROLE_SEEDING_ENABLED = ENVIRONMENT.bool('ROLE_SEEDING_ENABLED', default=True)
 GROUP_SEEDING_ENABLED = ENVIRONMENT.bool('GROUP_SEEDING_ENABLED', default=True)
+MAX_SEED_THREADS = ENVIRONMENT.int('MAX_SEED_THREADS', default=None)
 
 # disable log messages less than CRITICAL when running unit tests.
 if len(sys.argv) > 1 and sys.argv[1] == 'test':
--- a/Makefile
+++ b/Makefile
@@ -86,8 +86,8 @@ html:
 lint:
 	tox -elint
 
-reinitdb: 
-	make stop-compose 
+reinitdb:
+	make stop-compose
 	make remove-db
 	make start-db
 	sleep 5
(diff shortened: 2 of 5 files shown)


rbac/rbac/settings.py
```diff
@@ -348,6 +348,7 @@
 # Seeding Setup
 ROLE_SEEDING_ENABLED = ENVIRONMENT.bool('ROLE_SEEDING_ENABLED', default=True)
 GROUP_SEEDING_ENABLED = ENVIRONMENT.bool('GROUP_SEEDING_ENABLED', default=True)
+MAX_SEED_THREADS = ENVIRONMENT.int('MAX_SEED_THREADS', default=None)
```
j00bar, line 351: Does it make sense to give a definite default, versus leaving it open-ended?

coderbydesign: @j00bar - Maybe? I went back and forth on this yesterday. When it's `None`, it will use the default value, which is the current behavior. For now, I figured we'd just leave the behavior as is, while closing the connections explicitly. So the `max_workers` option would just be a no-op currently. 

Then if we want/need to limit the concurrency further in the future we could set `MAX_SEED_THREADS`.

I'm open to suggestions though, if we feel like we should limit it out of the gate! Initially that was the plan, since that was one way to prevent connections from being exhausted, but with closing the connections it removes that necessity. I ended up leaving it in more for future flexibility.

j00bar: Ah, okay - cool. I think that sounds fine - I just wouldn't want the default to exhaust the connections and it sounds like it won't.

PR #287, opened 2020-05-08
Add Django command for running seeds and abstract seed logic

coderbydesign:
In order to reuse the seeding logic to execute a full seeding run outside the
context of the app starting (in this case a Django command), this abstracts the
seeding for all tenants into a separate module.

We need to eventually decouple this from the app, so it can be run when our `rbac-config-sync`
job runs: https://github.com/RedHatInsights/e2e-deploy/blob/master/templates/rbac/config-sync.yaml
in order to not have seeding run for every deployment and pod. This is the first
step in getting to that point.

**Local Testing**

- to test that the current behavior is unchanged, ensure both `ROLE_SEEDING_ENABLED`
and `GROUP_SEEDING_ENABLED` are unset locally. in this case, seeds should run
when you start your local server. you should see seeds run in your console.

- to test that seeding can be disabled by default, set `ROLE_SEEDING_ENABLED`
and `GROUP_SEEDING_ENABLED` both to False and restart your local server. you
should not see seeds run in the console.

- to test the seeds can be triggered manually, from the command like run:
`./rbac/manage.py seeds` and you should see seeds run in the console.

### DIFF
--- a/rbac/management/management/commands/seeds.py
+++ b/rbac/management/management/commands/seeds.py
@@ -28,12 +28,19 @@ class Command(BaseCommand):
 
     help = 'Runs the seeding for roles and groups'
 
+    def add_arguments(self, parser):
+        """Add arguments to command."""
+        parser.add_argument('--roles-only', action='store_true')
+        parser.add_argument('--groups-only', action='store_true')
+
     def handle(self, *args, **options):
         """Handle method for command."""
-        logger.info('*** Seeding roles... ***')
-        role_seeding()
-        logger.info('*** Role seeding completed. ***\n')
+        if not options['groups_only']:
+            logger.info('*** Seeding roles... ***')
+            role_seeding()
+            logger.info('*** Role seeding completed. ***\n')
 
-        logger.info('*** Seeding groups... ***')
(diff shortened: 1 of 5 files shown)


rbac/management/management/commands/seeds.py
```diff
@@ -0,0 +1,39 @@
...
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+
+class Command(BaseCommand):
+    """Command class for running seeds."""
+
+    help = 'Runs the seeding for roles and groups'
+
+    def handle(self, *args, **options):
```
wcmitchell, line 31: Overall I like all of this change, one thought, though: Since we've separate booleans for running just the Role or Group seeds currently, would it be worth allowing an argument or something to run only one of the seed functions when calling this command?

astrozzc: Good suggestion👍 Just have to ensure the group seeds starts after the role seeding finished.

coderbydesign: Ah, great suggestion!

PR #289, opened 2020-05-15
Fix platform_default group cache invalidation

coderbydesign:
Jira: https://projects.engineering.redhat.com/browse/RHCLOUD-6522

There are triggers on all data relations which would affect a user's permissions,
to invalidate the user's policy cache. This includes adding/removing roles to/from 
a group. When this happens, the cache for all users in a group's policy is invalidated.

The problem is that for `platform_default` groups, there is no actual relation
between users and the group/policy/roles for that group. This means it effectively
bypasses the cache invalidation logic [1].

The fix for this is to conditionally invalidate all users' cache when the
`platform_default` group is updated.

[1] https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/policy/model.py#L56

### DIFF
--- a/rbac/management/cache.py
+++ b/rbac/management/cache.py
@@ -1,5 +1,6 @@
 """Redis-based caching of per-Principal per-app access policy."""
 
+import contextlib
 import json
 import logging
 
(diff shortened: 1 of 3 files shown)


rbac/management/cache.py
```diff
@@ -49,15 +50,31 @@ def get_policy(self, uuid, application):
...
         if not settings.ACCESS_CACHE_ENABLED:
             return
         try:
+            yield
+        except exceptions.RedisError:
+            logger.exception(err_msg)
+
+    def delete_policy(self, uuid):
+        """Purge the given user's policy from the cache."""
+        err_msg = "Error deleting policy for uuid %s", uuid
```
astrozzc, line 65: This would be a tuple instead of string, but seems it is working?

coderbydesign: @astrozzc it does work, yes, but I added a change in 6ea9c03 which should clean it up. 😄

PR #291, opened 2020-06-04
Add another field roles to Access model

astrozzc:
https://projects.engineering.redhat.com/browse/RHCLOUD-6572
- 

## Description of Intent of Change(s)
This is to change the Access relation to permission to 'one to many'

This is first step to make the permission unique.
After making the permission unique, it will save space for storing the
permissions and return available permissions as set easily.

## Local Testing
How can the feature be exercised? 

    Tenant.objects.create(schema_name='acct1000')
     with tenant_context(Tenant.objects.filter(schema_name='acct1000').first()):
	Access.objects.create(permission='catalog:resource:operation')
	Access.objects.create(permission='catalog:resource:operation2')
        Access.objects.create(permission='catalog:resource:operation')

Run the migration, and checked that there is only one permission catalog:resource:operation created. And when you try to create a new access object, e.g. 'approval:resource:test', a new permission will be created.

Is any special local setup required?

## Checklist
- [x] if API spec changes are required, is the spec updated?

- [x] are there any pre/post merge actions required? if so, document here.

- [x] are theses changes covered by unit tests?

- [x] if warranted, are documentation changes accounted for?

- [x] does this require migration changes?
  - [x] if yes, are they backwards compatible?

- [x] is there known, direct impact to dependent teams/components?
  -  if yes, how will this be handled?

PR #296, opened 2020-06-15
Add permission seeding

astrozzc:
Seed the permission from directory 'permissions'. Also provide
option to seed the permissions manually.

## Link(s) to Jira
https://projects.engineering.redhat.com/browse/RHCLOUD-6573

Seed the permissions which will be available to be added to the Access objects.

## Local Testing
Unit test
No

## Checklist


rbac/management/management/commands/seeds.py
```diff
@@ -30,17 +30,23 @@ class Command(BaseCommand):
...
     def add_arguments(self, parser):
         """Add arguments to command."""
+        parser.add_argument("--permissions-only", action="store_true")
         parser.add_argument("--roles-only", action="store_true")
         parser.add_argument("--groups-only", action="store_true")
 
     def handle(self, *args, **options):
         """Handle method for command."""
-        if not options["groups_only"]:
+        if options["permissions_only"]:
```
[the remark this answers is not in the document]
astrozzc: Good catch! Yea, I am changing it to `--roles|--groups|--permissions|--all`, this way we could seed any combinations of them, e.g. both `--groups --permissions`. And seed all of them by `--all`.

rbac/management/role/definer.py
```diff
@@ -84,3 +84,36 @@ def seed_roles(tenant, update=False):
...
+
+    with tenant_context(tenant):
+        for permission_file_name in permission_files:
+            permission_file_path = os.path.join(permission_directory, permission_file_name)
+            app_name = os.path.splitext(permission_file_name)[0]
+            with open(permission_file_path) as json_file:
+                data = json.load(json_file)
+                for resource, operations in data.items():
+                    for operation in operations:
+                        permission, created = Permission.objects.get_or_create(
```
coderbydesign, line 107: I think we may be want to do something here like an `update_or_create` [1] to ensure we're getting all fields set? It seems this migration [2] is only setting the `permission` field on `Permission`, so we end up with `Permission` records without the other fields populated. 

[1] https://docs.djangoproject.com/en/dev/ref/models/querysets/#update-or-create
[2] https://github.com/RedHatInsights/insights-rbac/pull/291/files#diff-abf50ea78e7ec39af811f223d8c6aa53R15

rbac/management/management/commands/seeds.py
```diff
@@ -18,29 +18,36 @@
...
-        parser.add_argument("--groups-only", action="store_true")
+        parser.add_argument("--permissions", action="store_true")
+        parser.add_argument("--roles", action="store_true")
+        parser.add_argument("--groups", action="store_true")
+        parser.add_argument("--all", action="store_true")
 
     def handle(self, *args, **options):
         """Handle method for command."""
-        if not options["groups_only"]:
+        if options["permissions"] | options["all"]:
```
coderbydesign, line 40: Is there an explicit reason to use a bitwise operator here?

rbac/management/role/definer.py
```diff
@@ -84,3 +84,36 @@ def seed_roles(tenant, update=False):
...
+    """For a tenant update or create defined permissions."""
+    permission_directory = os.path.join(settings.BASE_DIR, "management", "role", "permissions")
+    permission_files = [
+        f
+        for f in os.listdir(permission_directory)
+        if os.path.isfile(os.path.join(permission_directory, f)) and f.endswith(".json")
+    ]
+    # current_permission_ids = set()
+
+    with tenant_context(tenant):
```
coderbydesign, line 99: I think we'll want to use `with transaction.atomic():` here?

rbac/management/management/commands/seeds.py
astrozzc: Yea, to provide a convenient way "all" to support do all the seeding so we don't have to explicitly specify roles, groups, and permissions.

rbac/management/role/definer.py
astrozzc: Ah ha, good idea 👍

rbac/management/management/commands/seeds.py
astrozzc: Sorry, I miss understood, updated 👍

PR #297, opened 2020-06-16
Deploy plan

astrozzc:
## Link(s) to Jira
https://projects.engineering.redhat.com/browse/RHCLOUD-6573

To provide a compatible state for access so that code still works before migration is run

## Local Testing
Get the stable branch QA and Prod is running. Setup some Access objects without perm and permission_bridge defined. 
Then apply latest change without running migration. Setup the server, tried the /access endpoint,
which still works.

## Checklist


astrozzc:
This is step one and 2:

Steps:

	1. migrations of Access deploy together to CI: 
 	                                                                   perm            permission  			 
                 alter permission name		  =>  perm            permission_bridge 
                 add permission and copy perm  =>  perm            permission_bridge       permission
	Access filed in CI will become: perm (string), permission(string), permission_bridge(object)
	migrate to compatible mode for code in CI, but don't promote because the previous migration is together with code change, code is referencing the perm, will break before migration is done

	2. code deploy to CI to switch back to point back to permission (code in CI is pointing at perm)

	3. code and migration together all the way to prod 

	4. code deploy all the way to prod to switch from permission to perm

	5. migration deploy all the way to prod to drop permission and rename permission_bridge back to permission

PR #302, opened 2020-07-09
Use cloudwatch queues

bsquizz:
To avoid issues with cloudwatch rate limiting, `use_queues` should be set to True to batch the sending of logs.

https://github.com/kislyuk/watchtower/blob/master/watchtower/__init__.py#L55-L57

### DIFF
--- a/rbac/rbac/settings.py
+++ b/rbac/rbac/settings.py
@@ -266,7 +266,7 @@ if CW_AWS_ACCESS_KEY_ID:
         "log_group": CW_LOG_GROUP,
         "stream_name": NAMESPACE,
         "formatter": LOGGING_FORMATTER,
-        "use_queues": False,
+        "use_queues": True,
         "create_log_group": CW_CREATE_LOG_GROUP,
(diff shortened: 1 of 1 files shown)


lphiri:
@wcmitchell @coderbydesign could this is explain the issues  with cloudwatch in qa?

coderbydesign:
@lphiri it's quite possible this contributes to the resource issues, but we actually had to disable this explicitly in this commit (links to issues as to why, within): https://github.com/RedHatInsights/insights-rbac/commit/977fc6a8b49866de7093a6b2bb40f84b0ff3a61d which is part of this PR: https://github.com/RedHatInsights/insights-rbac/pull/260 in order to get CW logging working at all with gunicorn.

so there is an issue with using this library currently with RBAC either way, but we can't reenable this. I chatted about this with @bsquizz last night in Slack. apparently other apps have had similar issues.

coderbydesign:
Closing for now. We're updating how RBAC logging happens in general.

PR #304, opened 2020-07-14
Add permissions get endpoint

coderbydesign:
## Link(s) to Jira
- [Provide a way to return all permissions by account & by app, not by principal](https://projects.engineering.redhat.com/browse/RHCLOUD-6573)

We'll need an endpoint to return all permissions for an account, not by principal, and also per app scoped to the account, not principal. This new endpoint will also need query parameters to support filtering permissions by resource and operation, along with ordering support for application, resource and operation.

## Local Testing
The `Permission` model introduced in #291 supports this change. One caveat is that the table for the model is not populated by default, so you'll need to insert data in order to retrieve it from `/permissions/`

- As an org_admin I should be able to access the new endpoint /permissions/ to return all permissions for the account, which will not be scoped to a principal.

- As a *_NON_* org_admin I should *_NOT_* be able to access the new endpoint /permissions/, and receive a 403.

 - If I supply the /?application=catalog param, should see all catalog permissions for the account, not scoped to a principal.

 - Query parameters "resource" and "operation" (/access/?resource=&operation=) should also be supported, and filter permissions based off a match on the permission string, "application:resource:operation".

- Ordering on these new fields should be supported.

- The response payload should look like:
```json
{
  "meta": {
    "count": 2,
    "limit": 10,
    "offset": 0
  },
  "links": {
    "first": "/api/rbac/v1/permissions/?limit=10&offset=0",
    "next": "/api/rbac/v1/permissions/?limit=10&offset=10",
    "previous": null,
    "last": "/api/rbac/v1/permissions/?limit=10&offset=25"
  },
  "data": [
    {
      "app": "ansible-automation",
      "resource": "*",
      "operation": "*",
      "permission": "ansible-automation:*:*"
    },
    {
      "app": "approval",
      "resource": "actions",
      "operation": "read",
      "permission": "approval:actions:read"
    }
  ]
}
```

## Checklist

~- [ ] are there any pre/post merge actions required? if so, document here.~

~- [ ] if warranted, are documentation changes accounted for?~

~- [ ] does this require migration changes?~
~- [ ] if yes, are they backwards compatible?~

~- [ ] is there known, direct impact to dependent teams/components?~
~-  if yes, how will this be handled?~


wcmitchell:
Having trouble getting this to work locally, discussed with @coderbydesign and we suspect it may be a chicken and egg problem between different branches and migration paths. Will poke about a bit more to see what I can figure out. All the changes look good from a readthrough, it's just the DB state that concerns me.

coderbydesign:
@wcmitchell #307 should address the data issue we discussed. Since this is not currently being used, we shouldn't have any issues with getting this in and then running the migration to populate the fields.

coderbydesign:
This will be dependent on changes from #296 to ensure all fields of `Permission` records are populated, and a new migration to update the field names to align with the IPP. 

cc: @astrozzc

coderbydesign:
Updated based on changes from #310 from @astrozzc

PR #305, opened 2020-07-14
Use git and vault for smoke test configuration

catastrophe-brandon:
## Link(s) to Jira
- No JIRA issue

The what, why and how.

To avoid mishaps with clobbered configuration, we are switching to use vault and git for storing configuration.

## Local Testing
How can the feature be exercised?  N/A
How can the bug be exploited and fix confirmed? N/A
Is any special local setup required? N/A

## Checklist
- [ ] if API spec changes are required, is the spec updated? N/A

- [ ] are there any pre/post merge actions required? if so, document here. N/A

- [ ] are theses changes covered by unit tests? N/A

- [ ] does this require migration changes? N/A

- [ ] is there known, direct impact to dependent teams/components? N/A


catastrophe-brandon:
> You should also pass in the 'options' to execSmokeTest

Added to the bottom.

PR #306, opened 2020-07-16
Roles display name

wcmitchell:
## Link(s) to Jira

In order to allow for name changes in canned roles without drastically altering the DB and models involved, we need a `display_name` field that can be easily altered and displayed by the front end without altering the indexed `name` field.

This change adds that field, associated tests and docs, and the necessary migrations to make it usable.

## Local Testing
This change is pretty well self contained. Pulling down the branch code, allowing migrations to run, and then testing with role gets should display the new field.
NB: since we don't currently have any display_names defined, the new field defaults to the `name` field for all existing roles. The included tests ensure that non-name strings work as expected. Alternatively, altering the local testing .json config files to include display names should result in the creation of roles with alternate display names.

## Checklist

  - [?] if yes, are they backwards compatible?
    - Adds a new field that isn't used elsewhere, but is checked in test cases. Attempting to revert may result in test failures, but I don't *think* it'll affect the live system.


wcmitchell:
Closing to split between migration and functional code/tests

PR #307, opened 2020-07-16
Add migration to update Permissions

coderbydesign:
This migration [1] in #291 does not actually execute the custom `save()` method
which is an override to set the `app`, `resource` and `operation` fields when
the `Permission` objects are created [2].

This is because Django migrations will not run model methods [3].

This PR adds a data migration to populate the records created via #291, similar
to how the custom `save()` method will during creation of permissions during
seeding.

If there is no `permission` value, or if the split value is not valid, we'll skip the iteration.

[1] https://github.com/RedHatInsights/insights-rbac/pull/291/files#diff-abf50ea78e7ec39af811f223d8c6aa53R15
[2] https://github.com/RedHatInsights/insights-rbac/pull/291/files#diff-6d695aec79db3c1d2a95d1ecbd12fbe3R66-R68
[3] https://stackoverflow.com/a/37685925

## Local Testing
Assuming you have access and/or permission data, run migrations with `make run-migrations`.
You should now have fully populated `Permission` objects, whereas previously any records
would have only had `permission` populated.

## Checklist
  - [x] if yes, are they backwards compatible? - this is purely a data migration


coderbydesign:
Closing in favor of changes which will account for this in #296 per conversation with @astrozzc

PR #309, opened 2020-07-16
Role display name code

wcmitchell:
## Link(s) to Jira

In order to allow for name changes in canned roles without drastically altering the DB and models involved, we need a `display_name` field that can be easily altered and displayed by the front end without altering the indexed `name` field.

This change adds that field, associated tests and docs, and the necessary migrations to make it usable.

## Local Testing
This change change requires #308 but is otherwise is pretty well self contained. Pulling down the branch code, allowing migrations to run or ensuring that they've run from the previous PR, and then testing with role gets should display the new field.
NB: since we don't currently have any display_names defined, the new field defaults to the `name` field for all existing roles. The included tests ensure that non-name strings work as expected. Alternatively, altering the local testing .json config files to include display names should result in the creation of roles with alternate display names.

## Checklist

  - Pre merge we need to make sure that migrations including the #308 changes have been run

    - Migrations handled in #308 


wcmitchell:
Replaced by #321

PR #312, opened 2020-07-22
Fix role seed Access creation for new roles

coderbydesign:
The existing logic for setting `version_diff` during role creation [1] causes
`version_diff = defaults["version"] != role.version` to always evaluate to `False`
because we're using `defaults["version"]` to create the new role with, causing
`role.version` to be the same.

This means that when creating a new role, the `version_diff` check here [2] which
used to force `Access` objects to be created, will not be invoked.

This updates the check when setting `version_diff` to maintain the previously
functional behavior by ensuring a truthy value when a new role is being created.

Previously the expectation was that `role.version` would be a null value, thus
being able to use that in the check for `Access` creation since the value from the
config file would be a valid number value. This ensures the same behavior.

Alternately, we could change this check [3] to explicitly check against `created`,
and leave line 43 as is.

[1] https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/role/definer.py#L43
[2] https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/role/definer.py#L55
[3] https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/role/definer.py#L55

To exploit the issue on the `master` branch, create a _new_ role in an existing or new config file in `/rbac/management/role/definitions/*.json`, after you've created a tenant and the migrations and seeds have run. This only affects new roles being added during explicit seeding, where `update=True`, not during new tenant creation where `update=False`.

Run seeds as you normally would, and note in the database that `Access` records were not created for the new role. Subsequent runs of seeding, if you bump the version for that role, _will_ create the `Access` records.

Now checkout this PR's feature branch. Do the same as above where you create a net-new role in your config, and run the seeds manually. Now for that new role you should see the `Access` records.

### DIFF
--- a/rbac/management/role/definer.py
+++ b/rbac/management/role/definer.py
@@ -40,7 +40,7 @@ def _make_role(tenant, data, update=False):
         platform_default=data.get("platform_default", False),
     )
     role, created = Role.objects.get_or_create(name=name, defaults=defaults)
-    version_diff = defaults["version"] != role.version
+    version_diff = (defaults["version"] != role.version) or created
     if created:
         logger.info("Created role %s for tenant %s.", name, tenant.schema_name)
     elif version_diff:


j00bar:
It feels like having `created` and `update` as flags where both can be `True` is the fundamental issue, no? I get that it's an optimization for real-time tenant creation from API calls - but it feels like a minimal performance hit to ditch `update` as an option so long as we have `get_or_create`. Thoughts?

coderbydesign:
Closing in favor of #317. Thanks, @j00bar!

PR #321, opened 2020-08-05
Role display name code v2

wcmitchell:
## Link(s) to Jira

Code changes and migration to get things set up for display_name changes

## Local Testing
1. checkout master
2. make reinitdb
3. make serve
4. Hit an endpoint to create a tenant
5. checkout PR branch
6. Ensure role/access endpoints work 
7. make run-migrations
8. Again test role/access endpoints

## Checklist


tests/management/role/test_view.py
```diff
@@ -256,48 +286,7 @@ def test_read_role_list_success(self):
                 self.assertEqual(iterRole.get("accessCount"), 1)
                 role = iterRole
         self.assertEqual(role.get("name"), role_name)
-
-    def test_get_role_by_partial_name_by_default(self):
```
astrozzc, line 260: Are these deleted by accident?

rbac/management/migrations/0020_add_unique_to_display_name.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("management", "0019_recreate_display_name_field"),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
astrozzc, line 24: Curious, what would happen if we remove this line?

coderbydesign, line 24: Do we need both `AlterField` calls here or can the default/max/unique be set in one?

rbac/management/migrations/0020_add_unique_to_display_name.py
```diff
@@ -0,0 +1,31 @@
...
+            model_name="role", name="display_name", field=models.CharField(default="", max_length=150)
+        ),
+        migrations.RunPython(update_display_name),
```
coderbydesign, line 27: Can you refresh me on why we need this to run again? Is this because we didn't end up moving the `save` hook in the initial PR, so new roles will need to be updated?

rbac/management/role/definer.py
```diff
@@ -32,14 +32,15 @@
...
+    display_name = data.pop("display_name", name)
     access_list = data.get("access")
     defaults = dict(
         description=data.get("description", None),
         system=True,
         version=data.get("version", 1),
         platform_default=data.get("platform_default", False),
     )
-    role, created = Role.objects.get_or_create(name=name, defaults=defaults)
+    role, created = Role.objects.get_or_create(name=name, display_name=display_name, defaults=defaults)
```
coderbydesign, line 43: Do we want the `display_name` as part of the `get_or_create`? Won't this make it so if the `display_name` changes in the config it will create a new role vs update it? Then it will fail because of the name constraint (currently fails silently).

I think we may need to leave `display_name` out of the `get_or_create`, and have it as part of the `update` on line 48?
```
Role.objects.filter(name=name).update(**defaults, display_name=display_name)
```

wcmitchell: I think you're right that we'll need to add it to the update. In testing I ran into some errors with it not in get_or_create, but it's been a few days so I'll have to go back and see what exactly the behavior was. Will hop on that first thing tomorrow.

tests/management/role/test_view.py
wcmitchell: Oh, yep. Not sure if a rebase thing or a different issue, but I'll get that resolved here in a sec. Thanks!

rbac/management/migrations/0020_add_unique_to_display_name.py
wcmitchell: I'll test this, I think this was originally here to allow for the data migration to run with a default value, but I'll test just doing the data migration then adding on the uniqueness and default constraints afterwards.

rbac/management/role/definer.py
```diff
@@ -32,20 +32,22 @@
...
         description=data.get("description", None),
         system=True,
         version=data.get("version", 1),
         platform_default=data.get("platform_default", False),
     )
-    role, created = Role.objects.get_or_create(name=name, display_name=display_name, defaults=defaults)
+    role, created = Role.objects.get_or_create(name=name, defaults=defaults)
     if created:
+        if role.display_name != display_name:
+            role.display_name = display_name
```
coderbydesign, line 46: Do we need to call save on the `role` object here?

docs/source/specs/openapi.json
```diff
@@ -2399,6 +2399,10 @@
             "type": "string",
             "example": "RoleA"
           },
+          "display_name": {
```
coderbydesign, line 2402: Will we want to make this `readOnly: true`?

coderbydesign:
@wcmitchell I'm able to create/update roles via the API and set/change `display_name`. Is that expected?

wcmitchell:
I think changing things through the API is expected. I know we've gone back and forth on it a few times, and I'm not sure we've ever settled on a yea or a nay, but given that system/platform_default roles can't be affected through the API, and given that other role creation/editing is allowed for whitelisted apps via the API, I couldn't think of a good reason to make the display_name behave differently.

PR #322, opened 2020-08-06
Invalidate cache when role is removed from default group

wcmitchell:

## Link(s) to Jira
https://projects.engineering.redhat.com/browse/RHCLOUD-8126

Cache wasn't be invalidated when roles were removed from the default group. This change should resolve that.

## Local Testing
Use the reproducer script linked in the above Jira to test. Will need to change the target server and UUID for default group to your local configuration before testing.

## Checklist


astrozzc:
Tested locally, works well. I assume this would be ensured by feature tests

astrozzc:
Hmm, I am wondering how tests change the access for the platform_default group since we don't allow modifying the system and platform_default role. Hopefully we don't have another bug somewhere @catastrophe-brandon

coderbydesign:
@astrozzc you can still add/remove roles to and from the default group, you just can't modify the permissions on those system roles. Is that what you're referring to?

wcmitchell:
I'm not sure about unit testing unless we assume that the test suites will have cache access, if so then there may be a way to fold @catastrophe-brandon's testing script into a test method. I can do some digging. Not sure on the unit tests, though, will roll back through and check.

astrozzc:
> @astrozzc you can still add/remove roles to and from the default group, you just can't modify the permissions on those system roles. Is that what you're referring to?

Ah, you are right! Now I remember. I was confused by this: https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/group/view.py#L315, which is only modifying the group, not add/remove roles.

PR #324, opened 2020-08-12
Permission filters

coderbydesign:
## Link(s) to Jira
https://projects.engineering.redhat.com/browse/RHCLOUD-8443

It's been requested that the following fields (application, resource_type and verb) on the `/permissions/` endpoint support multiple, comma-separated values similar to ?application=app1,app2.

_**Note:** this does also move the `Group` view's `uuid_filter` to use this abstracted common filter for multiple values._

## Local Testing
To test this locally, use any/all of the below AC:
> I should be able to query /permissions/?application=app1,app2 and see permissions for app1 and app2.
> I should be able to query /permissions/?resource_type=res1,res1 and see permissions for res1 and res2.
> I should be able to query /permissions/?verb=read,write and see permissions for read and write.
> I should be able to combine any of the above query filters with multiple values and have them reflected in my results.

## Checklist


tests/management/permission/test_view.py
```diff
@@ -127,6 +127,53 @@ def test_get_list_is_the_only_valid_method(self):
...
+            expected_permissions = list(Permission.objects.values_list("permission", flat=True))
+
+        url = reverse("permission-list")
+        url = f"{url}?resource_type=roles,*"
+        client = APIClient()
+        response = client.get(url, **self.headers)
+        response_permissions = [p.get("permission") for p in response.data.get("data")]
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(response.data.get("data")), 4)
```
astrozzc, line 157: I am wondering if the number would cause any problem for smoke tests, where the permission data would be overwritten by actual ones, so the number might be large.

docs/source/specs/openapi.json
```diff
@@ -1767,7 +1767,7 @@
           {
             "name": "application",
             "in": "query",
-            "description": "Partial match for the application name of a permission.",
+            "description": "Partial match for the application name of a permission. You may also use a comma-separated list to match on multiple applications.",
```
astrozzc, line 1770: Does the partial match description still valid after this change? If there is a unit test to ensure this would be great!

wcmitchell:
Good points from Jay above, actual functionality seems good, though. Tried to break it in a few different ways and it always behaved as expected. LGTM once the above is addressed!

tests/management/permission/test_view.py
coderbydesign: Not sure I follow, @astrozzc? The permission data in these unit tests is isolated to the `setUp` block above, where we explicitly create 4 permissions. Smoke tests don't run within the context off the unit tests for RBAC. Let me know if I'm misunderstanding something! 😄

docs/source/specs/openapi.json
coderbydesign: Ah, good catch! No, the partial match would no longer work, here. I don't believe that was actually a requirement of the `/permissions/` endpoint, but I'm going to confirm.

coderbydesign: Confirmed from Robin that this is not an expectation/requirement. It was just a side-effect of implementation, so I'm going to remove the documentation for this support. Thanks, @astrozzc!

astrozzc: 👍great！

tests/management/permission/test_view.py
astrozzc: Ah got it! I thought it was created from the roles/permissions folder.

PR #326, opened 2020-08-18
Status type

astrozzc:
## Link(s) to Jira
https://projects.engineering.redhat.com/browse/RHCLOUD-5856

According to this https://jsonapi.org/format/#errors, the status
within error should be string instead of integer.

## Local Testing
There are unit tests to ensure this.

When getting errors, the status within error should be string (quoted integer)

No

## Checklist
- [X] if API spec changes are required, is the spec updated?

No
Yes

No
No

### DIFF
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -1972,9 +1972,6 @@
             "type": "array",
             "items": {
               "type": "object",
-              "required": [
-                "status"
-              ],
               "properties": {
                 "detail": {
                   "type": "string",
@@ -1984,6 +1981,7 @@
                   "type": "string",
                   "example": "403"
                 }
+              }
             }
           }
         }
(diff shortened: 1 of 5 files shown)


docs/source/specs/openapi.json
```diff
@@ -1972,9 +1972,15 @@
...
-                "detail": "Not Found.",
-                "status": 404
+              "properties": {
+                "detail": {
+                  "type": "string",
+                  "example": "Not found."
+                },
+                "status": {
+                  "type": "string",
+                  "example": "403"
```
wcmitchell, line 1982: Should this still be 404? It's just the spec example, so it probably doesn't matter, but wanted to make sure.

rbac/management/principal/proxy.py
```diff
@@ -132,7 +131,7 @@ def _request_principals(
         headers = {USER_ENV_HEADER: self.user_env, CLIENT_ID_HEADER: self.client_id, API_TOKEN_HEADER: self.api_token}
```
wcmitchell, line 131: Should the 200 code above (L130) also be a string?

wcmitchell: Misunderstood, only within errors! Closing this and above comment down

PR #332, opened 2020-08-27
Add checks for GUIDs to group requests

wcmitchell:

## Link(s) to Jira
- https://projects.engineering.redhat.com/browse/RHCLOUD-3699

Previously we didn't validate GUIDs before attempting to get the objects, leading to misleading 404s. This adds checks

## Local Testing
Pull down branch
make serve
Try to grab a /group/uuid/ endpoint with 'potato' or 'danceparty' as the uuid

## Checklist

### DIFF
--- a/rbac/management/group/view.py
+++ b/rbac/management/group/view.py
@@ -267,7 +267,7 @@ class GroupViewSet(
                 ]
             }
         """
-        self.validate_guid(request.path.split("/")[4])
+        self.validate_guid(kwargs.get("uuid"))
         return super().retrieve(request=request, args=args, kwargs=kwargs)
 
     def destroy(self, request, *args, **kwargs):
@@ -286,7 +286,7 @@ class GroupViewSet(
         @apiSuccessExample {json} Success-Response:
             HTTP/1.1 204 NO CONTENT
         """
-        self.validate_guid(request.path.split("/")[4])
+        self.validate_guid(kwargs.get("uuid"))
         self.protect_default_groups("delete")
         return super().destroy(request=request, args=args, kwargs=kwargs)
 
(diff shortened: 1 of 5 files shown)


wcmitchell:
I feel like there should be a smarter way to grab the UUID from the request path, but I couldn't really find a cleaner way than this. Open to suggestions!

rbac/management/group/view.py
```diff
@@ -605,6 +610,16 @@ def filters_from_params(self, valid_filters, model_name, request):
                 filters[f"{attr_filter_name}__icontains"] = param_value
         return filters
 
+    def validate_guid(self, uuid):
```
coderbydesign, line 613: This might be great to have in `utils.py`, what do you think? Seems like we could reuse this across different views. We're doing similar here: https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/group/view.py#L71-L76

We'd just need to update the signature to allow for customizing the key.

rbac/management/group/view.py
```diff
@@ -605,6 +610,16 @@ def filters_from_params(self, valid_filters, model_name, request):
...
+        """Verify GUID provided is valid before attempting to retrieve it."""
+        try:
+            UUID(uuid)
+        except ValueError:
+            key = "groups uuid validation"
+            message = f"{uuid} is not a valid UUID."
+            raise serializers.ValidationError({key: _(message)})
+        return
```
astrozzc, line 621: Nice! Overall LGTM! We could apply this to the /roles/ endpoint too! Nit: this return seems not necessary but not a big deal.

rbac/management/group/view.py
wcmitchell: Oh, yeah, that's a good idea. I'd thought about dropping the function elsewhere, but completely forgot about `utils.py`. I'll get that set up in a few

rbac/management/group/view.py
wcmitchell: Oh, yep, with you on the return. That was in there from stubbing things out. I'll pull that out real quick. After moving the function to utils.py as above, I can also pull it in to /roles/

PR #333, opened 2020-08-28
Return options for permission fields

astrozzc:
Return available options for permission fields so UI could call this api
to return lists to customer.

Also provide filters for the returned list.

## Link(s) to Jira
- https://projects.engineering.redhat.com/browse/RHCLOUD-8472

UI team wants to be able to list available options for fields of permission. 

## Local Testing
Setup local server and use curl:
e.g. curl -v GET "http://0.0.0.0:8080/r/insights/platform/rbac/v1/permissions/get_options/?field=application&verb=unlink" | python -m json.tool
Unit tests
No

## Checklist

### DIFF
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -1846,6 +1846,100 @@
           }
         }
       }
+    },
+    "/permissions/options/": {
+      "get": {
+        "tags": [
+          "Permission"
+        ],
+        "summary": "List the available options for fields of permissions for a tenant",
+        "description": "By default, options of application is returned. And could be resource_type or verb on demand.",
+        "operationId": "listPermissionOptions",
+        "parameters": [
+          {
+            "$ref": "#/components/parameters/QueryLimit"
+          },
+          {
+            "$ref": "#/components/parameters/QueryOffset"
+          },
+          {
+            "name": "field",
+            "in": "query",
+            "description": "specify which fields of permission to display",
+            "required": true,
+            "schema": {
+              "type": "string",
+              "enum": [
+                "application",
+                "resource_type",
+                "verb"
+              ]
+            }
+          },
+          {
+            "name": "application",
+            "in": "query",
+            "description": "Filter returned options based on application. You may also use a comma-separated list to filter on multiple applications.",
+            "required": false,
+            "schema": {
+              "type": "string"
+            }
+          },
+          {
+            "name": "resource_type",
+            "in": "query",
+            "description": "Filter returned options based on resource_type. You may also use a comma-separated list to filter on multiple resource_types.",
+            "required": false,
+            "schema": {
+              "type": "string"
+            }
+          },
+          {
+            "name": "verb",
+            "in": "query",
+            "description": "Filter returned options based on verb. You may also use a comma-separated list to filter on multiple verbs.",
+            "required": false,
+            "schema": {
+              "type": "string"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "A list of options for field of permission",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/PermissionOptionsPagination"
+                }
+              }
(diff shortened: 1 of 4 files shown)


rbac/management/permission/view.py
```diff
@@ -93,3 +98,41 @@ def list(self, request, *args, **kwargs):
...
+
+            HTTP/1.1 200 OK
+        {
+          "data": [
+            catalog, approval
+          ]
+        }
+        """
+        filters = {}
+        query_field = request.query_params.get(QUERY_FIELD, "application")
```
wcmitchell, line 130: This works, but can only handle returning a single field at a time. Is there any possibility that we'd need to return multiple fields with a request like /permissions/?field=application,verb&verb=read or something? Not sure if this is a viable use case or not.

coderbydesign: Yeah, I was wondering the same with regard to having multiple values like `?field=application,verb`.

I also feel like we should have some validation on this to prevent 500s when you supply an unsupported `field` value?

docs/source/specs/openapi.json
```diff
@@ -1837,6 +1837,100 @@
...
+              "enum": [
+                "application",
+                "resource_type",
+                "verb"
+              ],
+              "default": "application"
+            }
+          },
+          {
+            "name": "application",
```
coderbydesign, line 1866: Curious, were these query filters for `application`, `resource_type` and `verb` requested when you talked with Robin? I feel like it's a bit strange to say, "give me the verb options, but only verb options that equal 'a'". Especially when we have those filters on `/permissions/`, but maybe I'm missing a use-case?

astrozzc: What they want is returning all options, and also be able to filter for example "give me the verb options, but only resource_type that equal to 'a'" or "give me the resource_type options, but only application that equal to 'catalog'". And yea, the "give me the verb options, but only verb options that equal 'a'" is not our goal, just a side effect and we just not block users to use it.

rbac/management/permission/view.py
astrozzc: Yea, I would add a validation to the field. Multiple fields would be nice to have but temporary I can't think of a use case for that. Maybe we could leave that as a future enhancement?  

At first what they want is returning options for all the fields (application, resource_type, verb). Then I looked at their use case, they just want to that when customer click e.g. verb, get all the verbs and pop the list (this PR will make that possible), then customer could select one verb and then return all the permissions that have the verb (currently supported).
Then I discussed with Robin, maybe it is better to support when customer click e.g. application, get all application, then, customer select one of them, e.g., 'catalog', then when they click the resource_type, then they would get all the resource_type that belong to 'catalog' pop up.
This way, 
1. UI is satisfied for getting one field options;
2. we don't have to return all the fields, especially, when the available options become large, we could reduce the payload.

docs/source/specs/openapi.json
```diff
@@ -1846,6 +1846,100 @@
...
+                "application",
+                "resource_type",
+                "verb"
+              ]
+            }
+          },
+          {
+            "name": "application",
+            "in": "query",
+            "description": "Exact match for the application name of a permission. You may also use a comma-separated list to match on multiple applications.",
```
coderbydesign, line 1882: I don't think these descriptions are accurate for `application`, `resource_type` and `verb` here? It's not the match of the app name, it's the option filter key? Also doesn't support multiple values? Assuming it's just because it's a copy/paste from the permission entry? 😄

coderbydesign (commented):
Thanks for the changes, @astrozzc! Functionally this looks good! Just had a comment regarding the spec.

docs/source/specs/openapi.json
astrozzc: Yea, good catch! I changed that description. But it does support multiple values 😁 I just added a unit test to ensure that "test_return_options_with_comma_separated_filter"

coderbydesign (approved):

PR #336, opened 2020-09-09
Add cross account request model

astrozzc:
## Link(s) to Jira
- https://projects.engineering.redhat.com/browse/RHCLOUD-8875

This is one step to support TAM cross account access.

## Local Testing
Unit tests are added

## Checklist


rbac/management/cross_account_request/model.py
```diff
@@ -0,0 +1,56 @@
...
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+STATUS_LIST = ["requested", "cancelled", "approved", "denied"]
+
+
+class CrossAccountRequest(models.Model):
+    """Cross account request."""
+
+    request_id = models.UUIDField(default=uuid4, editable=False, unique=True, null=False, primary_key=True)
+    target_account = models.CharField(max_length=15)
+    requestor = models.CharField(max_length=150)
```
coderbydesign, line 36: Is `requestor` meant to be the TAM, as a request being initiated by the employee? Will we need to tie this to a principal?

rbac/management/cross_account_request/model.py
```diff
@@ -0,0 +1,56 @@
...
+    description = models.TextField(null=True)
```
coderbydesign, line 37: What will `description` be used for here?

rbac/management/cross_account_request/model.py
```diff
@@ -0,0 +1,56 @@
...
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+STATUS_LIST = ["requested", "cancelled", "approved", "denied"]
+
+
+class CrossAccountRequest(models.Model):
+    """Cross account request."""
+
+    request_id = models.UUIDField(default=uuid4, editable=False, unique=True, null=False, primary_key=True)
+    target_account = models.CharField(max_length=15)
```
coderbydesign, line 35: This would be the account the TAM is requesting access to, correct?

rbac/management/cross_account_request/model.py
```diff
@@ -0,0 +1,56 @@
...
+    requestor = models.CharField(max_length=150)
+    description = models.TextField(null=True)
+    created = models.DateTimeField(default=timezone.now)
+    startDate = models.DateTimeField(default=timezone.now)
+    endDate = models.DateTimeField()
+    modified = AutoDateTimeField(default=timezone.now)
+    status = models.CharField(max_length=10, default="requested")
```
coderbydesign, line 42: Should we set this up to be an enum? [Here's an example](https://medium.com/@bencleary/using-enums-as-django-model-choices-96c4cbb78b2e) of how we might want to enforce the enum for valid values on the model automatically.

coderbydesign (commented):
Looks like a solid start! Had a couple of questions. 😄

rbac/management/cross_account_request/model.py
astrozzc: Yes. Updated using requestor_account and requestor principal

rbac/management/cross_account_request/model.py
astrozzc: As discussed, this is not necessary currently, just for storing some info such as the request reason. But we could add this any time.

rbac/management/cross_account_request/model.py
astrozzc: Yes

wcmitchell (approved):
LGTM

rbac/api/cross_access/model.py
```diff
@@ -0,0 +1,62 @@
...
+#
+
+"""Models to store cross account access request."""
+from uuid import uuid4
+
+from django.db import models
+from django.utils import timezone
+from management.rbac_fields import AutoDateTimeField
+
+STATUS_LIST = ["requested", "cancelled", "approved", "denied"]
```
astrozzc, line 25: change "requested" to "pending"

coderbydesign, line 25: Would we expectation be that the "expired" state be implicit based on the start/end date, and not necessarily the status state in the db?

catastrophe-brandon: I guess that would incur the effort of calculating whether or not a given request has expired whenever a request is made. As long as there are a small number of requests at a given time, I don't expect it to be a performance burden. In the integration tests we'll probably want to figure out how to create a given request and force it to expire. On that note, will there be any validation of the start/end date at the time the request is created?

rbac/api/cross_access/model.py
```diff
@@ -0,0 +1,62 @@
...
+    first_name = models.CharField(max_length=150)
+    last_name = models.CharField(max_length=150)
+    email = models.CharField(max_length=150)
+
+
+class CrossAccountRequest(models.Model):
+    """Cross account access request."""
+
+    request_id = models.UUIDField(default=uuid4, editable=False, unique=True, null=False, primary_key=True)
+    target_account = models.CharField(max_length=15)
```
catastrophe-brandon, line 42: Would it make sense to also have `null=False` for this field? Is there a situation where the target account would be null?

rbac/api/cross_access/model.py
```diff
@@ -0,0 +1,62 @@
...
+    endDate = models.DateTimeField()
+    modified = AutoDateTimeField(default=timezone.now)
+    status = models.CharField(max_length=10, default="requested")
+
+    def validate_input_value(self):
+        """Validate status is valid, and date is valid."""
+        if self.status not in STATUS_LIST:
+            raise Exception(f'Unknown status "{self.status}" specified, {STATUS_LIST} are valid inputs.')
+
+        if self.startDate >= self.endDate:
```
catastrophe-brandon, line 55: If the start date and end date are both in the past it looks like they will be allowed.

tests/api/cross_access/test_model.py
```diff
@@ -0,0 +1,122 @@
...
+            target_account="123456", employee=self.employee, endDate=self.ref_time + timedelta(10)
+        )
+
+    def tearDown(self):
+        """Tear down cross account request model tests."""
+        Employee.objects.all().delete()
+        CrossAccountRequest.objects.all().delete()
+
+    def test_request_creation_success(self):
+        """Test the creation of cross account request."""
```
catastrophe-brandon, line 54: When a request is created, is there any validation to confirm that the creator is an internal user, or is that done completely by 3scale?

Also, this is probably my unfamiliarity with Django, but is there any referential integrity we can leverage to make sure that the target account is a real account that exists? If that validation is in-place, then I probably haven't come across it yet ;)

rbac/api/cross_access/model.py
coderbydesign: Maybe just a start/end date off now()? I was thinking about the implicit state as well. I suppose that would be a possibility to have an internal check on the state when a request object is requested, to update it before the response if it's expired?

astrozzc: yea, I was thinking about the "expired" state as implicit based on the end date. That does require extra calculation. 
It would be useful to make a scheduler to expire it (update the db status) on end date then, that way it won't introduce extra calculation when listing request objects.

rbac/api/cross_access/model.py
astrozzc: Good call, yea, we could add that.

rbac/api/cross_access/model.py
astrozzc: Right, will add check on that condition

tests/api/cross_access/test_model.py
astrozzc: That's done by 3scale, it would pass the identity of the creator.

No, me too, I haven't come across any account validation

rbac/api/cross_access/model.py
astrozzc: Oh, the default is null=False and blank=False for django model

rbac/api/cross_access/model.py
```diff
@@ -0,0 +1,72 @@
...
+from django.core.exceptions import ValidationError
+from django.db import models
+from django.utils import timezone
+from management.rbac_fields import AutoDateTimeField
+
+
+STATUS_LIST = ["pending", "cancelled", "approved", "denied", "expired"]
+
+
+class Employee(models.Model):
```
coderbydesign, line 31: @lphiri any concerns with storing PII for employees on these requests?

coderbydesign (approved):

rbac/api/cross_access/model.py
lphiri: @coderbydesign checking - email sent out.

lphiri:
@coderbydesign @astrozzc  we should look into cleaning up once someone is no longer an employee, same way we do for normal users.

PR #337, opened 2020-09-14
Add unique display_name constraint to role model

wcmitchell:

## Link(s) to Jira
- Quick fix, no Jira

We created a migration to ensure the display_name field is unique, but didn't actually add the constraint to the model. This fixes that.

## Local Testing
NA, this doesn't change any functionality, as that's handled by the migration. This is just ensuring that the model and migration are in line

## Checklist

### DIFF
--- a/rbac/management/role/model.py
+++ b/rbac/management/role/model.py
@@ -72,7 +72,7 @@ class Access(models.Model):
     """An access object."""
 
     perm = models.TextField(null=False)
-    permission = models.TextField(null=False,  default="*:*:*")
+    permission = models.TextField(null=False, default="*:*:*")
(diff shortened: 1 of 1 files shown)


astrozzc (approved):
Tested locally, which removes the extra make migrations. LGTM

rbac/management/role/model.py
```diff
@@ -72,7 +72,7 @@ class Access(models.Model):
     """An access object."""
 
     perm = models.TextField(null=False)
-    permission = models.TextField(null=False)
+    permission = models.TextField(null=False,  default="*:*:*")
```
coderbydesign, line 75: Looks like the linter is complaining about the whitespace here.
```suggestion
    permission = models.TextField(null=False, default="*:*:*")
```

coderbydesign (approved):

PR #340, opened 2020-09-18
Return only org admins within an account

astrozzc:
This is to support return org admins only within an account.

## Link(s) to Jira
JIRA: https://issues.redhat.com/browse/RHCLOUD-8988

Add ability to list org admins only instead of getting a bunch of users back and then filter by org_admin flag.

## Local Testing
Unit test
Local BOP server to do the query 

## Checklist

### DIFF
--- a/rbac/management/principal/view.py
+++ b/rbac/management/principal/view.py
@@ -31,6 +31,8 @@ SORTORDER_KEY = "sort_order"
 VALID_SORTORDER_VALUE = ["asc", "desc"]
 STATUS_KEY = "status"
 VALID_STATUS_VALUE = ["enabled", "disabled", "all"]
+ADMIN_ONLY_KEY = "admin_only"
+VALID_ADMIN_ONLY_VALUE = ["true", "false"]
 
 
 class PrincipalView(APIView):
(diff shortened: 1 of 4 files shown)


rbac/management/principal/view.py
```diff
@@ -116,7 +119,9 @@ def get(self, request):
                 user.account, email=email, limit=limit, offset=offset, sort_order=sort_order
             )
         else:
-            resp = proxy.request_principals(user.account, limit=limit, offset=offset, sort_order=sort_order)
+            resp = proxy.request_principals(
```
coderbydesign, line 122: Just looking at this and #342 and curious if at some point we might want to make an effort to refactor some of this to not cause the signature of this to continue to grow. Just wanted to point that out since I'm sure you'll have to rebase one of these. :)

coderbydesign (approved):
Minor comment, but 👍

wcmitchell (approved):
👍

rbac/management/principal/view.py
astrozzc: 👍 I will merge the other one and rebase this one to refactor it

PR #341, opened 2020-09-21
Add initial restricted internal API access

coderbydesign:
This change will introduce a prefix `/_private/` into the RBAC API, which will
be gated by the new `InternalIdentityHeaderMiddleware`, and will require that
the identity header be one from Turnpike.

We will need to setup the config in Turnpike similar to:
```
- name: rbac
  route: /api/rbac
  origin: http://rbac.svc.cluster.local:8080/_private
  ...
  ...
```

which will expose the `_private` routes to internal users. Anything coming through
the 3scale gateway on https://cloud.redhat.com/api/rbac/v1/* will continue to be
served through the current middlewares, and any attempt to request `_private`
through 3scale or directly will fail with a 403.

We will also short-circuit the current `IdentityHeaderMiddleware` during `process_request`
and `process_response` if the path matches a private URL, and ensure the `InternalIdentityHeaderMiddleware`
check is run (which will no-op `/api/` paths).

In addition to this setup, we've added two private routes:
```
/_private/api/tenant/unmodified/
/_private/api/tenant/<str:tenant_schema_name>/
```

The former will return the `schema_name` of all tenants which meet the following
criteria, deeming them "unmodified":

- tenant only has one group
- that one group is a system group (meaning the platform default that has _not_ been modified)
- tenant only has system roles (meaning they haven't created any new roles)
- the total number of groups == 1
- the total number of roles == the number of system roles

The latter will allow for a system endpoint in order to delete a schema. This will
ultimately help to prune tenants which haven't done anything with their RBAC
data, and thus can afford to be deleted and recreated if they access RBAC again,
helping to clean the ~30k schemas we currently manage in production now that
[this ticket](https://issues.redhat.com/browse/RHCLOUD-8796) has been fulfilled.

## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-8800

We need a way to access internal APIs to perform queries or actions against RBAC
which should be limited to employees. This will allow for these endpoints to
site behind Turnpike, our internal SSO gateway. Initially we'll be using this to
view/prune tenants which have not made updates to their default schema.

## Local Testing
Try to access `/_private/api/tenant/unmodified/` with the default external identity
set in `dev_middleware.py` and you should see that you get a 403.

Modify your `dev_middleware.py` file to include an `associate` key/value, and change
the `type` to `Associate`, mocking the identity from Turnpike, and you should now be 
able to access these endpoints.

In order to use a "destructive" endpoint, you'll need to set a valid timestamp (which will be set as UTC)
via `RBAC_DESTRUCTIVE_ENABLED` such as "10/01/2020" which
will set the destructive operations allowance to expire on 10/01/2020.

## Checklist
We'll need to setup Turnpike as noted above.


lphiri:
@coderbydesign - general comment on logging, I see we are using ".info" even for "audit" type logs.  Are info logs always enabled?

coderbydesign:
@lphiri info logs are enabled by default, yes.

tests/identity_request.py
```diff
@@ -88,23 +88,44 @@ def _create_customer(cls, account, create_tenant=True):
...
-        cls, customer_data, user_data, create_customer=True, create_tenant=False, is_org_admin=True
+        cls, customer_data, user_data, create_customer=True, create_tenant=False, is_org_admin=True, is_internal=False
     ):
         """Create the request context for a user."""
         customer = customer_data
         account = customer.get("account_id")
         if create_customer:
             cls.customer = cls._create_customer(account, create_tenant=create_tenant)
+
         identity = {
```
astrozzc, line 99: this seems could be removed since you have replaced it with _build_identity

rbac/internal/views.py
```diff
@@ -0,0 +1,78 @@
...
+    only_system_roles = role_qs and Role.objects.count() == role_qs.count()
+    return only_system_group and only_system_roles
+
+
+def list_unmodified_tenants(request):
+    """List unmodified tenants."""
+    logger.info(f"Unmodified tenants requested by: {request.user.username}")
+    tenant_qs = Tenant.objects.all()
+    to_return = []
+    for tenant_obj in tenant_qs:
```
astrozzc, line 56: I am wondering how long will this list take in prod. There is no progress bar for it so maybe just wait 😁

astrozzc (approved):
LGTM overall

rbac/internal/views.py
coderbydesign: I just updated a few of these queries to just return the counts, so hopefully that will also help.

tests/identity_request.py
coderbydesign: Good catch, thanks! It's removed, now. 😄

rbac/internal/views.py
```diff
@@ -0,0 +1,78 @@
...
+logger = logging.getLogger(__name__)
+
+
+def destructive_ok():
+    """Determine if it's ok to run destructive operations."""
+    now = datetime.datetime.utcnow().replace(tzinfo=pytz.UTC)
+    return now < settings.INTERNAL_DESTRUCTIVE_API_OK_UNTIL
+
+
+def tenant_is_unmodified():
```
j00bar, line 42: This _could_ be more database efficient, but it's not really a big deal... Here you've got 4 db queries regardless of whether it's a modified or unmodified tenant.

rbac/rbac/settings.py
```diff
@@ -315,3 +319,13 @@
 # Optionally log all DB queries
 if ENVIRONMENT.bool("LOG_DATABASE_QUERIES", default=False):
     LOGGING["loggers"]["django.db.backends"] = {"handlers": ["console"], "level": "DEBUG", "propagate": False}
+
+# Internal API Configuration
+INTERNAL_API_PATH_PREFIXES = ["/_private/"]
+
+try:
+    INTERNAL_DESTRUCTIVE_API_OK_UNTIL = parse_dt(
+        os.environ.get("RBAC_DESTRUCTIVE_ENABLED", "not-a-real-time"), tzinfos={"UTC": pytz.UTC}
```
j00bar, line 328: You may want to consider ensuring that this is a tz-aware datetime coming out of here, versus a naive one.

j00bar (approved):
Some minor suggestions inline, but otherwise...

Beautiful

rbac/rbac/settings.py
coderbydesign: Ah, for some reason I thought I'd tested this in a way that did ensure that, but that was only if you set the tz in the config. Pushed a commit that should resolve that, thanks!

rbac/internal/views.py
coderbydesign: Just pushed a change to short-circuit the counts as they check.

astrozzc (approved):
LGTM

j00bar (approved):

wcmitchell (approved):
LGTM!

PR #342, opened 2020-09-21
Return users with desired status

astrozzc:
This is to support returning users with desired status. Default is to return
only active users.

## Link(s) to Jira
- JIRA: https://issues.redhat.com/browse/RHCLOUD-8989

Provide FE options to return users with desired status

## Local Testing
Unit test available

Set up against local BOP which works

## Checklist


coderbydesign (approved):

wcmitchell (approved):
👍

PR #346, opened 2020-09-24
Add `?exclude_globals=true|false` to `/permissions/` GET endpoint

coderbydesign:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-8967

The experience team needs a way to filter out any permission with a "global/*"
value in any or all of the permission fields (`application`, `resource_type`, verb),
in order to not allow the options for UI role creation to have these values, but
rather _only_ granular permissions.

This sets up a query parameter `?exclude_globals` on the `/permissions/` GET
endpoint, with valid string values of `true` and `false`. By default, this will
be set to `false` and return all permissions.

If set to `true`, any permission such as:
```
*:*:*
*:*:baz
*:bar:*
*:bar:baz
foo:*:*
foo:*:baz
foo:bar:*
```
would not be returned, only those with granular values such as:
```
foo:bar:baz
```

## Local Testing
**Preserving Existing Behavior as the Default**
This should continue to return _all_ permissions: 
```
http://localhost:8000/api/rbac/v1/permissions/
```

**Exclude Globals**
This should return only permissions without global values: 
```
http://localhost:8000/api/rbac/v1/permissions/?exclude_globals=true
```

**Handle Invalid Values**
This should return a 400 if `exclude_globals` is not `true` or `false`: 
```
http://localhost:8000/api/rbac/v1/permissions/?exclude_globals=foo
```

**Combination of Filters:**
This should allow more granular filtering by combining filters:
```
http://localhost:8000/api/rbac/v1/permissions/?application=approval_local_test&exclude_globals=true
```

## Checklist


wcmitchell (approved):
👍

astrozzc (approved):
LGTM! Not sure why they choose exclude_globals instead of exclude_wildcards or sth. The name is a little bit confusion for me at first.

PR #347, opened 2020-09-25
Get rid of using permission field in access

astrozzc:
## Link(s) to Jira
- JIRA: https://issues.redhat.com/browse/RHCLOUD-8358

Currently, perm and permission are synced. Switch to user perm only and later
we could alter the permission field to object without any impact to service.

## Local Testing
Unit test
Setup server locally, reinitdb, run seeds and call /roles, /access endpoints and everything works as expected
curl -v GET "http://0.0.0.0:8080/r/insights/platform/rbac/v1/roles/{uuid}/" | python -m json.tool
curl -v GET "http://0.0.0.0:8080/r/insights/platform/rbac/v1/access/?application=" | python -m json.tool

## Checklist


rbac/management/role/definer.py
```diff
@@ -56,6 +56,7 @@ def _make_role(tenant, data):
             return role
     for access_item in access_list:
         resource_def_list = access_item.pop("resourceDefinitions", [])
+        access_item["perm"] = access_item.pop("permission")
```
coderbydesign, line 59: So this is just ensuring that we store the the permission on the `perm` value, correct? This will drop the `access_item["permission"]` value from being stored I believe, is that a concern?

rbac/management/role/model.py
```diff
@@ -83,14 +83,7 @@ def permission_application(self):
 
     def split_permission(self):
         """Split the permission."""
-        return self.permission.split(":")
-
-    def save(self, *args, **kwargs):
```
coderbydesign, line 88: Same with this, pulling this out and not updating it to store the `permission` field will leave the `permission` field with the default value of `*:*:*`.

rbac/management/role/definer.py
astrozzc: Yea, the access object will created with perm and without permission. We are going to drop the permission field in the next step 2, and create a new permission field, then in step 3, restore the value from this perm value.
Currently, the perm and permission are synced so it is fine to drop the permission value.

rbac/management/role/model.py
astrozzc: yea, this permission field will not be used any more, we are going to delete this permission field, and recreate it in the following PR

coderbydesign (approved):
Thanks for addressing my questions!

PR #351, opened 2020-09-28
Exclude the public schema from the Tenant query on internal endpoint

coderbydesign:
Locally, the code works as is. Deployed however, the `public` schema is returned
in `Tenant.objects.all()`, which causes subsequent queries using `management` models
to fail.

I'm not entirely sure why the `public` schema isn't returned locally as it is deployed,
but we should be excluding this anyway.


astrozzc (approved):
Oh, wow, I forgot about this too!!!

wcmitchell (approved):
Odd behavior, but makes sense. 👍

PR #356, opened 2020-10-02
Trigger/monitor migrations from internal API endpoint

coderbydesign:
## Link(s) to Jira

This sets up a `POST` endpoint on `/_private/api/migrations/run/` 
which will trigger migrations to be run in a background worker.

This will only be accessible via Turnpike/internal SSO. By default, we'll only be
able to run `migrate_schemas` without running the parallel executor, otherwise
the celery worker will complain about spawning more processes.

The output for this task will end up in the worker logs, which in a deployed environment
will help with isolated the output for migrations. 

This also sets up a `GET` endpoint on `/_private/api/migrations/progress/?migration_name=&app=`
which will allow us to get the status of a particular migration across all tenants.
By default the `?app=` will be `management`, but you can optionally supply the `app`
param for checking against `api` or other Django apps' migrations.

Given that we know what the last applied migration should be for an app during
a release, we can use this to supply that migration name to get back the completion
progress during a migration run, which will look like:

```
http://localhost:8000/_private/api/migrations/progress/?migration_name=0004_crossaccountrequest&app=api
```

```
{
  "migration_name": "0004_crossaccountrequest",
  "app_name": "api",
  "tenants_completed_count": 3,
  "total_tenants_count": 6,
  "incomplete_tenants": ["acct1", "acct2", "acct3"],
  "percent_completed": 50
}
```

## Local Testing
You'll need to ensure you have a celery process running locally to process the
queue. You can do so either with the `rbac-worker` service in the Docker
compose file if you run `docker-compose up` to run RBAC, or manually
by running similar to the following, configured to connect to your local redis:

```
$: pipenv shell
$: cd rbac/
$: celery worker -A rbac.celery --broker=redis://127.0.0.1:6379/0 --loglevel=INFO
```
Then you can trigger the process with:
```
http://localhost:8000/_private/api/migrations/run/
```

To check the progress of a migration, query a value from the `django_migrations` table
in the `name` column:

```
http://localhost:8000/_private/api/migrations/progress/?migration_name=0004_crossaccountrequest&app=api
```

TODO:
- [x] tests
- [ ] do we want migrations put behind the "destructive" gate? I lean towards no since
      it's idempotent, but maybe we should standardize on anything which alters the db
      should be, and only GETs are not?
- [x] add a monitoring endpoint to determine the progress of migrations

## Checklist

- [x] Input Validation

- [x] Output Encoding

- [x] Session Management

- [x] Access Control

- [x] Cryptographic Practices

- [x] Data Protection

- [x] Communication Security

- [x] System Configuration

- [x] Database Security

- [x] File Management

- [x] Memory Management

### DIFF
--- a/rbac/internal/views.py
+++ b/rbac/internal/views.py
@@ -84,7 +84,7 @@ def tenant_view(request, tenant_schema_name):
                 return HttpResponse(status=204)
             else:
                 return HttpResponse("Tenant cannot be deleted.", status=400)
-    return HttpResponse(status=405)
+    return HttpResponse(f'Method "{request.method}" not allowed.', status=405)
 
 
 def run_migrations(request):
@@ -93,7 +93,7 @@ def run_migrations(request):
         logger.info(f"Running migrations: {request.method} {request.user.username}")
         run_migrations_in_worker.delay()
         return HttpResponse("Migrations are running in a background worker.", status=202)
-    return HttpResponse(status=405)
+    return HttpResponse(f'Method "{request.method}" not allowed.', status=405)
 
 
 def migration_progress(request):
@@ -126,4 +126,4 @@ def migration_progress(request):
         }
 
         return HttpResponse(json.dumps(payload), content_type="application/json")
-    return HttpResponse(status=405)
+    return HttpResponse(f'Method "{request.method}" not allowed.', status=405)
--- a/rbac/internal/views.py
+++ b/rbac/internal/views.py
@@ -100,23 +100,30 @@ def migration_progress(request):
     """View method for checking migration progress."""
     if request.method == "GET":
         migration_name = request.GET.get("migration_name")
-        app = request.GET.get("app", "management")
+        app_name = request.GET.get("app", "management")
         if not migration_name:
             return HttpResponse("Please specify a migration name in the `?migration_name=` param.", status=400)
-        tenants_completed = 0
+        tenants_completed_count = 0
+        incomplete_tenants = []
         tenant_qs = Tenant.objects.exclude(schema_name="public")
         tenant_count = tenant_qs.count()
         for idx, tenant in enumerate(list(tenant_qs)):
             with tenant_context(tenant):
-                migrations_have_run = MigrationRecorder.Migration.objects.filter(name=migration_name, app=app).exists()
+                migrations_have_run = MigrationRecorder.Migration.objects.filter(
+                    name=migration_name, app=app_name
+                ).exists()
                 if migrations_have_run:
-                    tenants_completed += 1
+                    tenants_completed_count += 1
+                else:
+                    incomplete_tenants.append(tenant.schema_name)
         payload = {
             "migration_name": migration_name,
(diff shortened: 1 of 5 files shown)


rbac/internal/views.py
```diff
@@ -83,3 +85,45 @@ def tenant_view(request, tenant_schema_name):
...
     return HttpResponse(status=405)
+
+
+def run_migrations(request):
+    """View method for running migrations."""
+    if request.method == "POST":
+        logger.info(f"Running migrations: {request.method} {request.user.username}")
+        run_migrations_in_worker.delay()
+        return HttpResponse("Migrations are running in a background worker.", status=202)
+    return HttpResponse(status=405)
```
astrozzc, line 96: Should we put a message warning about supported method?

rbac/internal/views.py
```diff
@@ -83,3 +85,45 @@ def tenant_view(request, tenant_schema_name):
...
+            "migration_name": migration_name,
+            "app_name": app_name,
+            "tenants_completed_count": tenants_completed_count,
+            "total_tenants_count": tenant_count,
+            "incomplete_tenants": incomplete_tenants,
+            "percent_completed": int((tenants_completed_count / tenant_count) * 100),
+        }
+
+        return HttpResponse(json.dumps(payload), content_type="application/json")
+    return HttpResponse(status=405)
```
astrozzc, line 129: Same here, just saying the method other than get is not allowed/supported?

rbac/management/tasks.py
```diff
@@ -18,10 +18,17 @@
...
 @shared_task
 def principal_cleanup():
     """Celery task to clean up principals no longer existing."""
     clean_tenants_principals()
+
+
+@shared_task
+def run_migrations_in_worker():
+    """Celery task to run migrations."""
+    call_command("migrate_schemas")
```
astrozzc, line 34: Curious, is there a way to run this with --executor=parallel?

astrozzc (approved):
Nice!! Overall LGTM

rbac/management/tasks.py
coderbydesign: Not that I'm aware of. That's what I was referring to yesterday where we can't run a subprocess in the celery worker. Worth digging into a bit more, but if we have a smaller list of schemas and a way to monitor them, I think it's less important to run them in parallel.

wcmitchell, line 34: Do we need any error checking/handling on this, or will Django handle, say, a failed migration for us?

wcmitchell (approved):
One question, but everything looks good overall. 👍

rbac/management/tasks.py
coderbydesign: Not really, because it kicks it off in a background process. However, if a migration fails, it should continue with the next, but that's something that would be baked into django-tenant-schemas I believe. We will get the logs regardless of what happens, in our worker pod though. So really I think this would be the same as running them in a process the way we do now, without our own error handling, but rather letting the library do that. Open to suggestions if you think we should try to do more, though!

PR #357, opened 2020-10-05
Add internal endpoint for triggering seeds

coderbydesign:
## Link(s) to Jira

Adds a `POST` endpoint at `/_private/api/seeds/run/` with optional params for
specifying which seeds to run with `?seed_types=roles,permissions,groups`.

When this is not specified, it will default to all, similar to the underlying
Django command.

If an unsupported seed type is presented, it will 400 the request in order to
help prevent seeds from failing silently in a background process.

## Local Testing
Ensure you have a celery worker running locally, and run a `POST` request
against `/_private/api/seeds/run/` with/without `?seed_types=` populated.

## Checklist

- [x] Input Validation

- [x] Output Encoding

- [ ] Session Management

- [ ] Access Control

- [ ] Cryptographic Practices

- [ ] Data Protection

- [ ] Communication Security

- [ ] System Configuration

- [ ] Database Security

- [ ] File Management

- [ ] Memory Management

### DIFF
--- a/rbac/internal/views.py
+++ b/rbac/internal/views.py
@@ -26,7 +26,7 @@ from django.db.migrations.recorder import MigrationRecorder
 from django.http import HttpResponse
 from django.shortcuts import get_object_or_404
 from management.models import Group, Role
-from management.tasks import run_migrations_in_worker
+from management.tasks import run_migrations_in_worker, run_seeds_in_worker
 from tenant_schemas.utils import tenant_context
 
 from api.models import Tenant
@@ -53,7 +53,10 @@ def tenant_is_unmodified():
 
 
 def list_unmodified_tenants(request):
-    """List unmodified tenants."""
+    """List unmodified tenants.
+
+    GET /_private/api/tenant/unmodified/
+    """
     logger.info(f"Unmodified tenants requested by: {request.user.username}")
     tenant_qs = Tenant.objects.exclude(schema_name="public")
     to_return = []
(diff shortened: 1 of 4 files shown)


rbac/internal/views.py
```diff
@@ -127,3 +139,24 @@ def migration_progress(request):
...
+    """View method for running seeds.
+
+    POST /_private/api/seeds/run/?seed_types=permissions,rolese,groups
+    """
+    if request.method == "POST":
+        args = {}
+        option_key = "seed_types"
+        valid_values = ["permissions", "roles", "groups"]
+        seed_types_param = request.GET.get(option_key)
+        if seed_types_param:
```
wcmitchell, line 154: May be able to use https://github.com/RedHatInsights/insights-rbac/blob/449c39e0fc9de547165b70a502d362a6784cfa15/rbac/management/utils.py#L155 or something similar here?

wcmitchell (approved):
Looks good! One comment on a potential DRY opportunity, but nothing blocking.

astrozzc (approved):

rbac/internal/views.py
coderbydesign: Good suggestion! Unfortunately I don't think `validate_and_get_key` will work out of the box in this case because a) the default value should actually be empty, in which case it would fail unless I pass in "groups,permissions,roles" as the default value, but also b) because it expects (and returns) one value, vs doing the "optional multiple values" with splitting a string like "groups,roles".

PR #377, opened 2020-11-06
Move cache of tenant from memory to redis

astrozzc:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-10094

The cache of tenant is in memory causes problem when we prune
the tenants. The tenant persists within the server until we
restart it. Therefore, moving the cache into redis so we could prune
it when we prune the tenant.
    
## Local Testing
Unit test
setting up the server and call 
tried the curl -X DELETE "http://0.0.0.0:8080/_private/api/tenant/acct10001/" and everything works fine

## Checklist

- [ ] Input Validation

- [ ] Output Encoding

- [ ] Session Management

- [ ] Access Control

- [ ] Cryptographic Practices

- [ ] Data Protection

- [ ] Communication Security

- [ ] System Configuration

- [ ] Database Security

- [ ] File Management

- [ ] Memory Management


rbac/management/cache.py
```diff
@@ -34,6 +34,96 @@ def connection(self):
...
+        return None
+
+    def delete_cached(self, index, obj_name):
+        """Delete cache from redis."""
+        err_msg = f"Error deleting {obj_name} for {index}"
+        with self.delete_handler(err_msg):
+            logger.info(f"Deleting {obj_name} cache for {index}")
+            self.connection.delete(self.key_for(index))
+
+    def set_cache(self, pipe, index, item):
```
coderbydesign, line 66: Should this be the default behavior, or should we leave the existing logic for `AccessCache` in that class/method? Just wondering if this is "general" enough to be reused, or if the classes should implement it themselves since we have different methods to set (`set` vs `hset`)?

rbac/management/cache.py
```diff
@@ -34,6 +34,96 @@ def connection(self):
...
+        except exceptions.RedisError:
+            logger.exception(err_msg)
+
+    def get_from_redis(self, index):
+        """Get object from redis based on index."""
+        obj = self.connection.hget(*(self.key_for(index[0]), index[1]))
+        if obj:
+            return json.loads(obj)
+
+    def get_cached(self, index, error_message):
```
coderbydesign, line 51: The `index` naming in this call chain confused me for a bit, because it doesn't appear to be the actual index value, but rather the key?

rbac/management/cache.py
```diff
@@ -34,6 +34,96 @@ def connection(self):
...
+    def delete_handler(self, err_msg):
+        """Handle delete events."""
+        try:
+            yield
+        except exceptions.RedisError:
+            logger.exception(err_msg)
+
+    def get_from_redis(self, index):
+        """Get object from redis based on index."""
+        obj = self.connection.hget(*(self.key_for(index[0]), index[1]))
```
coderbydesign, line 47: This wasn't terribly clear to me, either, what the `index[0]` / `index[1]` values are, and what the expected interface is for them. Again, maybe a case of asking whether or not we should be assuming `h*` are the methods to use by default? Just throwing it out there for consideration. 😄

coderbydesign (commented):
Functionally looks good, just had a few questions!

rbac/management/cache.py
astrozzc: 👍I am moving the implementation out

rbac/management/cache.py
astrozzc: Agree, changing it to key

rbac/management/cache.py
```diff
@@ -34,41 +34,122 @@ def connection(self):
...
+    def delete_cached(self, key, obj_name):
+        """Delete cache from redis."""
+        err_msg = f"Error deleting {obj_name} for {key}"
+        with self.delete_handler(err_msg):
+            logger.info(f"Deleting {obj_name} cache for {key}")
+            self.connection.delete(self.key_for(key))
+
+    def set_cache(self, pipe, key, item):
+        """Set cache to redis."""
+        raise Exception("Please override the set_cache method.")
```
coderbydesign, line 66: Love the pattern, but a small nit is that we may want to use `NotImplementedError` in these cases?

coderbydesign (approved):
Awesome!

rbac/management/cache.py
astrozzc: Thanks! This is what I was looking for

wcmitchell (approved):
👍

PR #379, opened 2020-11-18
Verify permission exists on role creation

coderbydesign:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-10250

We currently check against our allow-list to ensure permissions of a role being
created have their `application` field defined in list.

However, iff that check passes, you are still able to create a permission for any
`resource_type` or `verb`, which may or may not be a valid permission. For example,
given the following allow-list:

```
ROLE_CREATE_ALLOW_LIST = "cost-management"
```

This role creation would succeed:
```
{"name": "Custom Role", "access": [{"permission": "cost-management:foo:bar", "resourceDefinitions": []}]}
```

And this role creation would fail:
```
{"name": "Custom Role", "access": [{"permission": "rbac:foo:bar", "resourceDefinitions": []}]}
```

Assuming we don't have a permission defined as `cost-management:foo:bar`, this
would not be what we want. The changes in this PR will enforce the `resource_type`
and `verb` check, to ensure a permission in the DB defined with `cost-management:foo:bar`
exists first.

## Local Testing
- on `master`, try to create a role with a permission for an app which exists in
  your local `ROLE_CREATE_ALLOW_LIST`, with a random `resource_type` and `verb`
- note that this currently works
- checkout this feature branch
- run the same POST as you did on `master`
- this should now fail with a 400 and helpful message
- now send another POST, this time with a permission string you _do_ have locally
  in your db
- this should create the role

## Checklist

- [x] Input Validation

- [ ] Output Encoding

- [ ] Session Management

- [ ] Access Control

- [ ] Cryptographic Practices

- [ ] Data Protection

- [ ] Communication Security

- [ ] System Configuration

- [ ] Database Security

- [ ] File Management

- [ ] Memory Management

### DIFF
--- a/rbac/internal/views.py
+++ b/rbac/internal/views.py
@@ -150,7 +150,7 @@ def migration_progress(request):
 def run_seeds(request):
     """View method for running seeds.
 
-    POST /_private/api/seeds/run/?seed_types=permissions,rolese,groups
+    POST /_private/api/seeds/run/?seed_types=permissions,roles,groups
     """
     if request.method == "POST":
         args = {}
--- a/rbac/management/role/view.py
+++ b/rbac/management/role/view.py
@@ -27,6 +27,7 @@ from django.http import Http404
 from django.utils.translation import gettext as _
 from django_filters import rest_framework as filters
 from management.filters import CommonFilters
+from management.models import Permission
 from management.permissions import RoleAccessPermission
 from management.querysets import get_role_queryset
 from management.role.serializer import AccessSerializer, RoleDynamicSerializer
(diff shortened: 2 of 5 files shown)


astrozzc (commented):
Overall LGTM, it might be better to add this for the update as well 😉

wcmitchell (commented):
All seems good! Agree with Jay that we probably want the same check in update, though

wcmitchell (approved):
👍

astrozzc (approved):

PR #388, opened 2020-12-01
WIP - Clowder integration

bennyturns:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-10213

Integrate RBAC with Clowder operator / ephemeral envs.

## Local Testing
How can the feature be exercised?    Normal automated tests in stage / prod, smoke tests in eph cluster.
Is any special local setup required?   Test using CRC, minikube, or the eph cluster.

## Checklist

- [ ] Input Validation

- [ ] Output Encoding

- [ ] Session Management

- [ ] Access Control

- [ ] Cryptographic Practices

- [ ] Data Protection

- [ ] Communication Security

- [ ] System Configuration

- [ ] Database Security

- [ ] File Management

- [ ] Memory Management

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,8 @@ ENV NODEJS_VERSION=10 \
     PATH=$HOME/.local/bin/:$HOME/node_modules/.bin/:$HOME/.npm-global/bin/:$PATH \
     LC_ALL=en_US.UTF-8 \
     LANG=en_US.UTF-8 \
-    PIP_NO_CACHE_DIR=off
+    PIP_NO_CACHE_DIR=off \
+    ENABLE_PIPENV=true
 
 ENV SUMMARY="Insights RBAC is a role based access control web server" \
     DESCRIPTION="Insights RBAC is a role based access control web server"
--- a/Dockerfile.new
+++ b/Dockerfile.new
@@ -1,0 +1,11 @@
+FROM registry.redhat.io/ubi8/python-36
+USER 0
+COPY Pipfile Pipfile
+COPY Pipfile.lock Pipfile.lock
+COPY run_server.sh run_server.sh
+RUN yum install -y git gcc python3-devel && pip3 install pipenv pip pipenv-to-requirements && pip3 install -U pip && pipenv run pipenv_to_requirements -f \
+    && pip3 install -r requirements.txt && yum remove -y nodejs gcc python3-devel atlas && yum clean all
+COPY rbac rbac
+COPY scripts scripts
+COPY openshift/s2i/bin/run run
+ENTRYPOINT ["./run"]
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "12a3e1adac98a22dd4826c5f73f03c3d313d4b79238ebdd765036288c8290418"
+            "sha256": "e6f8899039b8e62f68acf442f44097c6ecbd987eabbcadd467d17824480d5753"
         },
         "pipfile-spec": 6,
         "requires": {
@@ -21,8 +21,14 @@
                 "sha256:70cdb10628468ff14e57ec2f751c7aa9e48e7e3651cfd62d431213c0c4e58f21",
                 "sha256:aa7f313fb887c91f15474c1229907a04dac0b8135822d6603437803424c0aa59"
             ],
-            "markers": "python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3, 3.4'",
             "version": "==2.6.1"
+        },
+        "app-common-python": {
+            "hashes": [
+                "sha256:c0a0ff2ecf58f28f9252349ca32b7b97086163fdbb196a29d05561f4caad32a1"
+            ],
+            "index": "pypi",
+            "version": "==0.1.2"
         },
         "billiard": {
             "hashes": [
@@ -56,10 +62,10 @@
         },
         "certifi": {
             "hashes": [
-                "sha256:5930595817496dd21bb8dc35dad090f1c2cd0adfaf21204bf6732ca5d8ee34d3",
-                "sha256:8fc0819f1f30ba15bdb34cceffb9ef04d99f420f68eb75d901e9560b8749fc41"
+                "sha256:1f422849db327d534e3d0c5f02a263458c3955ec0aae4ff09b95f195c59f4edd",
+                "sha256:f05def092c44fbf25834a51509ef6e631dc19765ab8a57b4e7ab85531f0a9cf4"
             ],
-            "version": "==2020.6.20"
+            "version": "==2020.11.8"
         },
         "chardet": {
             "hashes": [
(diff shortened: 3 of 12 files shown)


Dockerfile.new
```diff
@@ -0,0 +1,11 @@
...
+USER 0
+COPY Pipfile Pipfile
+COPY Pipfile.lock Pipfile.lock
+COPY run_server.sh run_server.sh
+RUN yum install -y git gcc python3-devel && pip3 install pipenv pip pipenv-to-requirements && pip3 install -U pip && pipenv run pipenv_to_requirements -f \
+    && pip3 install -r requirements.txt && yum remove -y nodejs gcc python3-devel atlas && yum clean all
+COPY rbac rbac
+COPY scripts scripts
+COPY openshift/s2i/bin/run run
+ENTRYPOINT ["./run"]
```
coderbydesign, line 11: Will this Dockerfile be expected to be used for the app service as well as the scheduler/worker? Asking because we can't run this entrypoint for the scheduler/worker, those have different startup commands: https://github.com/RedHatInsights/e2e-deploy/blob/master/templates/rbac/worker.yaml#L49-L54

Sorry if I'm missing something. 😄

Dockerfile
```diff
@@ -10,7 +10,8 @@ ENV NODEJS_VERSION=10 \
     PATH=$HOME/.local/bin/:$HOME/node_modules/.bin/:$HOME/.npm-global/bin/:$PATH \
     LC_ALL=en_US.UTF-8 \
     LANG=en_US.UTF-8 \
-    PIP_NO_CACHE_DIR=off
+    PIP_NO_CACHE_DIR=off \
+    ENABLE_PIPENV=true
```
coderbydesign, line 14: Curious what this change is related to? Will clowder be using Dockerfile.new solely? We do have some UI components in the app currently for viewing the DRF browsable API.

Dockerfile.new
coderbydesign: I do see you have those in the clowder files below, but they should be run off the base image without the ./run script being run, I believe.

openshift/s2i/bin/run
```diff
@@ -97,7 +98,7 @@ if is_gunicorn_installed; then
     export WEB_CONCURRENCY=${WEB_CONCURRENCY:-$(get_default_web_concurrency)}
 
     echo "---> Serving application with gunicorn ($APP_MODULE) ..."
-    exec gunicorn "$APP_MODULE" --chdir=$APP_HOME --bind=0.0.0.0:8080 --config "$APP_CONFIG" --preload
+    exec gunicorn "$APP_MODULE" --chdir=$APP_HOME --bind=0.0.0.0:8000 --config "$APP_CONFIG" --preload
```
coderbydesign, line 101: These port changes I assume will mean we'll need to make some adjustments to our CI/QA config?

rbac/rbac/settings.py
```diff
@@ -213,10 +214,18 @@
     "ORDERING_PARAM": "order_by",
 }
 
-CW_AWS_ACCESS_KEY_ID = ENVIRONMENT.get_value("CW_AWS_ACCESS_KEY_ID", default=None)
-CW_AWS_SECRET_ACCESS_KEY = ENVIRONMENT.get_value("CW_AWS_SECRET_ACCESS_KEY", default=None)
-CW_AWS_REGION = ENVIRONMENT.get_value("CW_AWS_REGION", default="us-east-1")
-CW_LOG_GROUP = ENVIRONMENT.get_value("CW_LOG_GROUP", default="platform-dev")
+# CW settings
+if ENVIRONMENT.bool("CLOWDER_ENABLED", default=False):
```
coderbydesign, line 218: Just a note on the CW integration with RBAC. We currently have this disabled, by way of having null values for the keys in our environments, because we ran into issues where we weren't able to uses parallel processing, and the synchronous CW delivery was killing our pods. https://github.com/RedHatInsights/insights-rbac/pull/302#issuecomment-656615590

tmp/rbac.yml
```diff
@@ -0,0 +1,362 @@
+apiVersion: cloud.redhat.com/v1alpha1
```
coderbydesign, line 1: Curious what this tmp file is used for?

coderbydesign (commented):
Thanks for the PR @bennyturns! Left a handful of comments/questions, mostly just to understand the context. Let me know if chatting would be easier. 😄

Dockerfile
psav: So, @bennyturns and I were working together on this at the end, originally @bennyturns was using your setup, which was based on s2i I believe, but @kylape asked me to see if we could put together a new raw Dockerfile, so that we were not based off of s2i anymore. I would imagine this change can be ignored, but would need to ask @bennyturns.

Dockerfile.new
psav: I think the `command` overrides the entrypoint, but yes, we can change this so that the single image has no `command/entrypoint` per se, and we use the command in ClowderApp instead to invoke. Would you be OK with that?

openshift/s2i/bin/run
psav: Actually, there is a little more work to do on these, these port changes should actually be driven by the `cdappconfig.json` file which is mounted inside the container. Usually apps will run up a service and query the JSON document, failing back on an env var if clowder isn't enabled. We can emulate the same functionality here if you wish?

rbac/rbac/settings.py
psav: We could put another environment variable on the template here to simply disable the logging altogether?

tmp/rbac.yml
psav: This is the `ClowdApp` that would be used to deploy RBAC

Dockerfile.new
coderbydesign: Seems like that should work! That way we can reuse the one image for all the services, which is the goal?

openshift/s2i/bin/run
coderbydesign: That sounds like it would be a great approach. 👍

rbac/rbac/settings.py
coderbydesign: I think that would be great! That would clear up some ambiguity around whether or not we'd got it disabled. 😄

tmp/rbac.yml
coderbydesign: 👍

Pipfile.lock
```diff
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "12a3e1adac98a22dd4826c5f73f03c3d313d4b79238ebdd765036288c8290418"
+            "sha256": "e6f8899039b8e62f68acf442f44097c6ecbd987eabbcadd467d17824480d5753"
```
kylape, line 4: why does the lock file have such a large diff?

tmp/rbac.yml
kylape: There are two other ClowdApp yaml docs in this PR.  Can you explain what distinguishes one from another?

Dockerfile
bennyturns: I will remove this change.

PR #390, opened 2020-12-02
Add patch support to role detail

coderbydesign:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-10938

This change allows for a `PATCH` request to be made against `/roles/<uuid>/` without
needing to supply the entire `Role` instance payload, along with the `access` object
required in the `PUT`.

Instead, you can now supply a `PATCH` with any/all of the following:
```json
{
  "name": "Name of a role",
  "display_name": "Display name of a role",
  "description": "Description of a role"
}
```

This will then return the entire role object, the same as a `GET` or `PUT` would.

This makes use of the `partial_update` method, which skips the access validation,
validates the UUID and then updates the role using a new serializer `RolePatchSerializer`
which is only used on `PATCH` requests to `/roles/<uuid>`. This serializer inherits
from the `RoleSerializer` but overrides the `access` to be optional, and does not
validate `access` on update, explicitly setting the `name`, `display_name` and
`description` fields only.

## Local Testing
To test this locally:
- submit the following payload as a `PUT` on a role and see how validation fails
  because you currently need to send the entire payload:
  ```
    {
      "name": "Name of a role",
      "display_name": "Display name of a role",
      "description": "Description of a role"
    }
  ```
- submit the same payload as a `PATCH` and see that it succeeds and returns the role

## Checklist

- [ ] Input Validation

- [ ] Output Encoding

- [ ] Session Management

- [ ] Access Control

- [ ] Cryptographic Practices

- [ ] Data Protection

- [ ] Communication Security

- [ ] System Configuration

- [ ] Database Security

- [ ] File Management

- [ ] Memory Management


catastrophe-brandon (approved):

astrozzc (approved):
Tested locally, everything works great!

wcmitchell (approved):
👍

PR #394, opened 2020-12-08
Add management models to public schema

coderbydesign:
## Description of Intent of Change(s)
This will setup the tables in the `public` schema which is what #389 was set out to do, but without duplication of migrations and models.

**eed802e**
To support the upcoming TAM cross-request access features as well as moving away
from Django Tenant Schemas, this will sync the `management` app models with the
public schema structure moving forward.

Initially, we'll be seeding the `system` data in this schema in order to populate
the front-end with the roles and permissions available for cross-access requests,
which will ultimately be tied via a many-to-many relation to cross-access request
object.

Eventually, these models will be kept in sync with individual tenant schema
data, to migrate away from a multi-schema tenant architecture.

**b21359f**
- sets up a data migration to create a `public` Tenant
- remove the checks on `schema_name != 'public'` when running seeds and the principal
  cleanup task. we should manage these for the public schema now as well, now that
  the models exist in the `public` schema

## Local Testing
- from `master`, ensure that you have at least one tenant created first
- note that the `django_migrations` table in the `public` schema has `management_` tables, the same as your tenant schema(s)
- checkout this feature branch
- run `./rbac/manage.py migrate_schemas --fake management zero --schema public` to remove the `djanto_migrations` records for `management_` from the `public` schema
- verify the above worked in the `public` schema, but left your tenant schema(s) unaffected
- run `make run-migrations`
- verify that the public schema now has the `management` tables
- run `./rbac/manage.py seeds`
- verify that seeds have been run against the public schema

## Checklist
- [ ] if API spec changes are required, is the spec updated?

- [x] are there any pre/post merge actions required? if so, document here.
  - we'll need to run the 3 scripts as outlined in the local testing section, in order, in all envs

- [ ] are theses changes covered by unit tests?

- [ ] if warranted, are documentation changes accounted for?

- [ ] does this require migration changes?
  - [ ] if yes, are they backwards compatible?

- [ ] is there known, direct impact to dependent teams/components?
  -  if yes, how will this be handled?

## Secure Coding Practices Checklist Link
- https://github.com/RedHatInsights/secure-coding-checklist

## Secure Coding Practices Checklist
- [ ] Input Validation

- [ ] Output Encoding

- [ ] Authentication and Password Management

- [ ] Session Management

- [ ] Access Control

- [ ] Cryptographic Practices

- [ ] Error Handling and Logging

- [ ] Data Protection

- [ ] Communication Security

- [ ] System Configuration

- [ ] Database Security

- [ ] File Management

[...]

PR #395, opened 2020-12-10
Associate the roles with cross account requests

astrozzc:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-10911

Associate the roles with cross account requests
Add a many to many relation between roles and cross account requests.

## Local Testing
Unit test of model change is added.

## Checklist

https://github.com/RedHatInsights/insights-rbac/pull/394

- [ ] Input Validation

- [ ] Output Encoding

- [ ] Session Management

- [ ] Access Control

- [ ] Cryptographic Practices

- [ ] Data Protection

- [ ] Communication Security

- [ ] System Configuration

- [ ] Database Security

- [ ] File Management

- [ ] Memory Management


rbac/api/cross_access/model.py
```diff
@@ -39,6 +39,7 @@ class CrossAccountRequest(models.Model):
     end_date = models.DateTimeField(null=False, blank=False, default=None)
     modified = AutoDateTimeField(default=timezone.now)
     status = models.CharField(max_length=10, default="pending")
+    roles = models.ManyToManyField("management.Role", related_name="cross_requests")
```
coderbydesign, line 42: Just for clarity, since the naming of this model is a bit confusing, maybe we should consider something like `related_name="cross_account_requests"` for the relation?

coderbydesign (commented):
I do see one thing that _could_ be concerning. Since you can access the `api` models from within a tenant's schema, you can still do something like this while you're in a tenant's schema (not the public schema):
```
from api.models import CrossAccountRequest
from management.models import Role

role = Role.objects.first() # let's say this role has an id of 1 on schema 10001
requests = role.cross_requests # this will return the cross account requests for role 1 on the public schema, since the join table has the id of the CAR and the role (in this case 1). 
```
So it will return the role with the same id in the tenat's schema, even though it's tied to a role in the public schema with ID of 1. Does that make sense? I feel like ideally we'd not be able to access those relations from within a tenant's schema, or enforce some sort of validation, otherwise it's theoretically possible to grab the wrong data for a CrossAccountRequest.

wcmitchell (approved):
Pulled this down with an existing custom local env, everything seemed stable and working properly. Wasn't entirely sure how best to test the new model, but unit tests are passing, so tentative approval here

PR #405, opened 2021-01-18
Fix QA build

coderbydesign:
Move fix from #404 into `stable` 

cc: @bennyturns


wcmitchell (approved):
👍 Fixed issues in build phase

PR #415, opened 2021-01-28
Add ability to remove permissions and roles in seeds

coderbydesign:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-8338
- https://issues.redhat.com/browse/RHCLOUD-11705

When we run seeds, we should reconcile permissions and roles, in order to allow for permissions and `system` roles to be removed from the config and have them deleted from RBAC.

Currently, when role or permission is removed from the config, it is not removed from the database, which means we often need to run one-off scripts to remove them.

This ensures we remove any permissions which don't exist in the database after running the seeds, which would mean there are permissions which aren't reflected in the config (source of truth). The same is true for roles, where any `system` roles will be removed if they don't exist in the config files but do exist in the database.

**Note:** _We may want to consider what happens if there's an iteration failure, leading to a record which _is_ in the database not being added to the `set`, thus removing it._

This will only be run as log output, currently. This will allow us to identify any permissions/roles which would _currently_ be removed, to ensure we can add them to the config files if necessary, before running this destructively. We'll follow-up with a PR to actually start running the deletes. Optionally, we could control this with an env variable if we want to.

## Local Testing
- run seeds on a new db
- manually add two roles, one `system` and one non-`system` in the db
- manually add a new permission in the db
- run seeds 
- confirm the new permission is marked for deletion
- confirm the new `system` role is marked for deletion, but not the non-`system` role

## Checklist

  - run seeds

- [ ] Input Validation

- [ ] Output Encoding

- [ ] Session Management

- [ ] Access Control

- [ ] Cryptographic Practices

- [ ] Data Protection

- [ ] Communication Security

- [ ] System Configuration

- [ ] Database Security

- [ ] File Management

- [ ] Memory Management


rbac/management/role/definer.py
```diff
@@ -83,14 +83,23 @@ def seed_roles(tenant):
...
                 with open(role_file_path) as json_file:
                     data = json.load(json_file)
                     role_list = data.get("roles")
-                    _update_or_create_roles(tenant, role_list)
+                    role = _update_or_create_roles(tenant, role_list)
+                    current_role_ids.add(role.id)
+
+        roles_to_delete = Role.objects.filter(system=True).exclude(id__in=current_role_ids)
+        logger.info(
+            f"[READ-ONLY] Deleting the following '{roles_to_delete.count()}' roles(s): {roles_to_delete.values()}"
```
wcmitchell, line 99: I think this is the right idea, but the log message may be unclear to someone seeing it go by. Maybe worth changing `Deleting the following roles` to something like `The following {count} roles eligible for removal` or altering the `[READ-ONLY]` to WIP or [TESTING]?

rbac/management/role/definer.py
```diff
@@ -131,16 +140,16 @@ def seed_permissions(tenant):
...
                             )
-
-            # TODO: Remove permissions that are no longer exist, could enable later after enforcing
-            # all available permission from the permissions files (there might be some custom ones for
-            # Costmanagement or Remediations currently)
-            # Permission.objects.exclude(id__in=current_permission_ids).delete()
-            # Override delete methods for Permission to remove the related Access objects
+        perms_to_delete = Permission.objects.exclude(id__in=current_permission_ids)
+        logger.info(
+            f"[READ-ONLY] Deleting the following '{perms_to_delete.count()}' permission(s): {perms_to_delete.values()}"
```
wcmitchell, line 151: Same here as above, not sure if [READ ONLY] fully indicates that the operation isn't actually being performed at the time.

wcmitchell (approved):
A few questions on wording of log messages, but nothing worth holding up merging.

rbac/management/role/definer.py
coderbydesign: Good call. It's confusing, I'll update! Thanks!

coderbydesign:
When @astrozzc is back, I'd like to chat about potential issues around the fact that we create permissions during role seeding, and how this affects that. I believe we should be ok, since a permission created during role seeding _should_ still exist in the permission file. If it doesn't, we'll want to add it.

rbac/management/role/definer.py
```diff
@@ -67,7 +67,7 @@ def _update_or_create_roles(tenant, roles):
     """Update or create roles from list."""
     for role_json in roles:
         try:
-            _make_role(tenant, role_json)
```
astrozzc, line 70: Where will this return? Will it break the for loop?

coderbydesign: Nice catch, Jay! I added a new commit that ensures we collect all IDs.

astrozzc (approved):
LGTM

PR #424, opened 2021-02-17
Permission return

astrozzc:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-11904

## Description of Intent of Change(s)
- For cross account principals. Do not return the platform default permissions for them.

- When listing groups for cross account principal, do not return platform group since
they will not inherit the platform default group.

- Only canned roles could be assigned to cross account request principal: when cross account request is created, we verify that the assigned roles are system roles.

- Add a verification in the middleware to check the cross account requests.
Verify if the user is internal and email address ends with @redhat.com.

## Local Testing
Unit tests have covered most cases.

## Checklist
- [ ] if API spec changes are required, is the spec updated?

- [ ] are there any pre/post merge actions required? if so, document here.

- [ ] are theses changes covered by unit tests?

- [ ] if warranted, are documentation changes accounted for?

- [ ] does this require migration changes?
  - [ ] if yes, are they backwards compatible?

- [ ] is there known, direct impact to dependent teams/components?
  -  if yes, how will this be handled?

## Secure Coding Practices Checklist Link
- https://github.com/RedHatInsights/secure-coding-checklist

## Secure Coding Practices Checklist
- [ ] Input Validation

- [ ] Output Encoding

- [ ] Authentication and Password Management

- [ ] Session Management

- [ ] Access Control

- [ ] Cryptographic Practices

- [ ] Error Handling and Logging

- [ ] Data Protection

- [ ] Communication Security

- [ ] System Configuration

- [ ] Database Security

- [ ] File Management

- [ ] Memory Management

- [ ] General Coding Practices

PR #426, opened 2021-02-23
Add update endpoints and tests for cross-account API

wcmitchell:

## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-8882

Adds PUT and PATCH handlers for updating cross-account requests

## Local Testing
tox will run the unit tests written to try the new endpoints

Outside of that, pulling down the branch and running `make serve` as usual should allow for creation and updating of cross account requests. PUT or PATCH requests to [...]/rbac/v1/cross-account-requests/[uuid]/ will update allowed fields.

NB: Updating an expired request will instead create a new request for the requested account/roles/timeline

## Checklist
- Spec update to come

- [x] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [x] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/api/cross_access/model.py
+++ b/rbac/api/cross_access/model.py
@@ -53,10 +53,7 @@ class CrossAccountRequest(models.Model):
 
         [self.validate_date(date) for date in [self.start_date, self.end_date]]
 
-        if (
-            isinstance(self.end_date, datetime.datetime)
-            and isinstance(self.start_date, datetime.datetime)
-        ):
+        if isinstance(self.end_date, datetime.datetime) and isinstance(self.start_date, datetime.datetime):
             if self.start_date.date() > (datetime.datetime.now() + datetime.timedelta(60)).date():
                 raise ValidationError("Start date must be within 60 days of today.")
 
--- a/rbac/api/cross_access/model.py
+++ b/rbac/api/cross_access/model.py
@@ -56,9 +56,15 @@ class CrossAccountRequest(models.Model):
         if (
             isinstance(self.end_date, datetime.datetime)
             and isinstance(self.start_date, datetime.datetime)
-            and self.start_date.date() > self.end_date.date()
         ):
-            raise ValidationError("Start date must be earlier than end date.")
+            if self.start_date.date() > (datetime.datetime.now() + datetime.timedelta(60)).date():
+                raise ValidationError("Start date must be within 60 days of today.")
+
+            if self.start_date > self.end_date:
+                raise ValidationError("Start date must be earlier than end date.")
+
+            if self.end_date - self.start_date > datetime.timedelta(365):
+                raise ValidationError("Access duration may not be longer than one year.")
 
     def save(self, *args, **kwargs):
         """Override save method to validate some input."""
--- a/rbac/api/cross_access/serializer.py
+++ b/rbac/api/cross_access/serializer.py
@@ -95,3 +95,16 @@ class CrossAccountRequestDetailSerializer(serializers.ModelSerializer):
         for role in roles:
             request.roles.add(role)
         return request
+
+    def update(self, instance, validated_data):
+        """Override the update method to associate the roles to cross account request after it is updated."""
+        role_data = validated_data.pop("roles")
+        display_names = [role["display_name"] for role in role_data]
+        for field in validated_data:
+            setattr(instance, field, validated_data.get(field))
+
+        roles = Role.objects.filter(display_name__in=display_names)
+        for role in roles:
+            instance.roles.add(role)
+        instance.save()
+        return instance
--- a/rbac/api/cross_access/serializer.py
+++ b/rbac/api/cross_access/serializer.py
@@ -67,8 +67,8 @@ class CrossAccountRequestDetailSerializer(serializers.ModelSerializer):
     request_id = serializers.UUIDField(read_only=True)
     target_account = serializers.CharField(max_length=15)
     user_id = serializers.CharField(max_length=15)
-    start_date = serializers.DateTimeField(format="%m/%d/%Y")
-    end_date = serializers.DateTimeField(format="%m/%d/%Y")
+    start_date = serializers.DateTimeField(format="%m/%d/%Y", input_formats=["%m/%d/%Y"])
+    end_date = serializers.DateTimeField(format="%m/%d/%Y", input_formats=["%m/%d/%Y"])
     created = serializers.DateTimeField(format="%m/%d/%Y", read_only=True)
     status = serializers.CharField(max_length=10, read_only=True)
     roles = RoleSerializer(many=True)
@@ -98,13 +98,15 @@ class CrossAccountRequestDetailSerializer(serializers.ModelSerializer):
 
     def update(self, instance, validated_data):
         """Override the update method to associate the roles to cross account request after it is updated."""
-        role_data = validated_data.pop("roles")
-        display_names = [role["display_name"] for role in role_data]
+        if "roles" in validated_data:
+            role_data = validated_data.pop("roles")
+            display_names = [role["display_name"] for role in role_data]
+            roles = Role.objects.filter(display_name__in=display_names)
+            instance.roles.clear()
+            instance.roles.add(*roles)
+
         for field in validated_data:
             setattr(instance, field, validated_data.get(field))
 
-        roles = Role.objects.filter(display_name__in=display_names)
-        for role in roles:
-            instance.roles.add(role)
         instance.save()
         return instance
--- a/rbac/api/cross_access/serializer.py
+++ b/rbac/api/cross_access/serializer.py
@@ -16,6 +16,7 @@
 #
 
 """Serializer for CrossAccountRequest."""
+from django.db import transaction
 from management.models import Role
 from management.permission.serializer import PermissionSerializer
 from rest_framework import serializers
@@ -96,6 +97,7 @@ class CrossAccountRequestDetailSerializer(serializers.ModelSerializer):
             request.roles.add(role)
         return request
 
+    @transaction.atomic
     def update(self, instance, validated_data):
         """Override the update method to associate the roles to cross account request after it is updated."""
         if "roles" in validated_data:
--- a/rbac/api/cross_access/view.py
+++ b/rbac/api/cross_access/view.py
@@ -18,12 +18,16 @@
 """View for cross access request."""
 from datetime import datetime, timedelta
 
+import pytz
+from django.db.models.query import QuerySet
 from django.utils import timezone
+from django.utils.translation import gettext as _
 from django_filters import rest_framework as filters
 from management.models import Principal, Role
 from management.principal.proxy import PrincipalProxy
-from management.utils import validate_and_get_key, validate_limit_and_offset
-from rest_framework import mixins, viewsets
+from management.utils import validate_and_get_key, validate_limit_and_offset, validate_uuid
+from rest_framework import mixins, serializers, viewsets
+from rest_framework.response import Response
 from rest_framework.serializers import ValidationError
 from tenant_schemas.utils import tenant_context
 
@@ -37,6 +41,7 @@ ACCOUNT = "target_account"
 USER_ID = "user_id"
 VALID_QUERY_BY_KEY = [ACCOUNT, USER_ID]
 PARAMS_FOR_CREATION = ["target_account", "start_date", "end_date", "roles"]
+VALID_PATCH_FIELDS = ["start_date", "end_date", "roles", "status"]
 
 PROXY = PrincipalProxy()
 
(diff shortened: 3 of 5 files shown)


rbac/api/cross_access/serializer.py
```diff
@@ -95,3 +95,16 @@ def create(self, validated_data):
...
+
+    def update(self, instance, validated_data):
+        """Override the update method to associate the roles to cross account request after it is updated."""
+        role_data = validated_data.pop("roles")
+        display_names = [role["display_name"] for role in role_data]
+        for field in validated_data:
+            setattr(instance, field, validated_data.get(field))
+
+        roles = Role.objects.filter(display_name__in=display_names)
+        for role in roles:
```
coderbydesign, line 107: I think you could also do `instance.roles.add(*roles)` here?

rbac/api/cross_access/view.py
```diff
@@ -107,6 +116,54 @@ def list(self, request, *args, **kwargs):
...
+        payload = request.data
+        for field in payload:
+            if field not in VALID_PATCH_FIELDS:
+                key = "cross-account request"
+                message = f"Field '{field}' is not supported. Please use one or more of: {VALID_PATCH_FIELDS}."
+                error = {key: [_(message)]}
+                raise serializers.ValidationError(error)
+
+        with tenant_context(Tenant.objects.get(schema_name="public")):
+            current = self.get_object()
```
coderbydesign, line 131: From my understanding, this will return the object based on the pk?

rbac/api/cross_access/view.py
```diff
@@ -107,6 +116,54 @@ def list(self, request, *args, **kwargs):
             return self.replace_user_id_with_info(result)
         return result
 
+    def partial_update(self, request, *args, **kwargs):
```
coderbydesign, line 119: Should we do anything if you don't provide any update data at all?

rbac/api/cross_access/view.py
```diff
@@ -211,3 +283,52 @@ def create_principal(self, target_account, user_id):
...
+            if start_date and isinstance(start_date, str):
+                start_date = pytz.utc.localize(datetime.strptime(start_date, "%m/%d/%Y"))
+            if end_date and isinstance(end_date, str):
+                end_date = pytz.utc.localize(datetime.strptime(end_date, "%m/%d/%Y"))
+        except ValueError:
+            raise self.throw_validation_error(
+                "cross-account-update", "start_date or end_date does not match format: '%m/%d/%Y'."
+            )
+
+        if start_date and start_date > pytz.utc.localize(datetime.now() + timedelta(60)):
```
coderbydesign, line 304: Is it possible to reuse some of the create logic for this? Maybe this is what you were referring to last week about DRY-ing things up eventually, actually? :)

rbac/api/cross_access/view.py
```diff
@@ -211,3 +283,52 @@ def create_principal(self, target_account, user_id):
...
+            raise self.throw_validation_error("cross-account-update", "Start date must be within 60 days of today.")
```
coderbydesign, line 305: I wonder...will we get into a situation where you could potentially perpetually update a request to bump it out 60 days? Probably an edge-case that may not be a concern, but I just thought about that.

coderbydesign (commented):
Still looking through this, but left a few notes so far. I'm going to pull it down and run through things locally as well.

rbac/api/cross_access/view.py
```diff
@@ -159,6 +216,21 @@ def throw_validation_error(self, source, message):
         error = {source: [message]}
         raise ValidationError(error)
 
+    def update_cross_account_request(self, car, update_data):
+        """Update cross-account-request with list of role names."""
+        if car.status == "pending" and update_data.get("status").lower() == "approved":
```
coderbydesign, line 221: This will throw a 500 if you don't supply `status` in the `PUT` payload.

rbac/api/cross_access/view.py
```diff
@@ -211,3 +283,52 @@ def create_principal(self, target_account, user_id):
...
+
+        if roles:
+            with tenant_context(Tenant.objects.get(schema_name="public")):
+                for role in roles:
+                    try:
+                        Role.objects.get(display_name=role)
+                    except Role.DoesNotExist:
+                        raise self.throw_validation_error("cross-account-update", f"Role '{role}' does not exist.")
+
+        if target_account:
```
coderbydesign, line 323: Do we want to allow the `target_account` to be updated?

rbac/api/cross_access/view.py
```diff
@@ -159,6 +216,21 @@ def throw_validation_error(self, source, message):
         error = {source: [message]}
         raise ValidationError(error)
 
+    def update_cross_account_request(self, car, update_data):
+        """Update cross-account-request with list of role names."""
+        if car.status == "pending" and update_data.get("status").lower() == "approved":
+            # create_principal_for_approved_request()
+            pass
+        for field in update_data:
```
coderbydesign, line 224: Should we maybe wrap this in a transaction in case any of the updates fail, we won't wipe out the roles first and commit that?

rbac/api/cross_access/view.py
```diff
@@ -107,6 +116,54 @@ def list(self, request, *args, **kwargs):
...
+    def update(self, request, *args, **kwargs):
+        """Update a cross-account request."""
+        validate_uuid(kwargs.get("pk"), "cross-account request uuid validation")
+
+        with tenant_context(Tenant.objects.get(schema_name="public")):
+            current = self.get_object()
+
+            self.validate_and_get_input_for_creation(request.data)
+
+            if current.status == "expired":
```
astrozzc, line 161: This would create a new cross-account request but the old one is still there. So when user update it, it still shows up. And it might be better to keep the request_id the same, in the future, we would use that to check the history of the request. Or use super.update? Maybe we could still call the update_cross_account_reuqest, and update the status of it?

rbac/api/cross_access/serializer.py
```diff
@@ -95,3 +95,16 @@ def create(self, validated_data):
         for role in roles:
             request.roles.add(role)
         return request
+
+    def update(self, instance, validated_data):
```
astrozzc, line 99: This seems never get called, do you mean using super().update but used super().create in the view.py?

rbac/api/cross_access/view.py
```diff
@@ -107,6 +116,54 @@ def list(self, request, *args, **kwargs):
             return self.replace_user_id_with_info(result)
         return result
 
+    def partial_update(self, request, *args, **kwargs):
+        """Patch a cross-account request."""
```
astrozzc, line 120: Is this the method for the admin in target account to approve/deny the request?

rbac/api/cross_access/serializer.py
wcmitchell: Oh, that's slick. Lemme test with that.

rbac/api/cross_access/view.py
wcmitchell: Yep, that's exactly what I was thinking. 100%

rbac/api/cross_access/serializer.py
```diff
@@ -95,3 +95,16 @@ def create(self, validated_data):
...
+        """Override the update method to associate the roles to cross account request after it is updated."""
+        role_data = validated_data.pop("roles")
+        display_names = [role["display_name"] for role in role_data]
+        for field in validated_data:
+            setattr(instance, field, validated_data.get(field))
+
+        roles = Role.objects.filter(display_name__in=display_names)
```
coderbydesign, line 106: Based on dev sync, we'll likely want to add a filter on `system` here similar to: https://github.com/RedHatInsights/insights-rbac/pull/424/files#diff-9b7f444eec2c52f6eb5578161c2077e0aa9037fbaa6c8ed89eefdcc46dfee381R192

tests/api/cross_access/test_view.py
```diff
@@ -370,3 +383,121 @@ def test_create_requests_fail_for_over_60_day_start_date(self):
...
+            url,
+            update_data,
+            format="json",
+            **self.associate_admin_request.META
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data.get("start_date"), update_data.get("start_date"))
+
+    def test_partial_update_fail_invalid_field(self):
```
catastrophe-brandon, line 467: Pretty reasonable unit test coverage here. I think I might like to see a little coverage with bad request body or invalid data for both PUT and PATCH. For example:
* Body with extra fields
* Body with fields whose types do not match the spec
* In cases where we take URL params, invalid or mangled parameter values

Historically we've done this type of testing in our integration test suite, but I think they make more sense as unit tests. Also, the generated clients don't easily lend themselves to testing with non-spec inputs, so we may have more flexibility in the unit test coding than we would with the limitations of the IQE framework. If it makes sense we could do these unit tests in a follow-up PR. Alternatively, I we could create a PR with just unit tests that targets this branch.

rbac/api/cross_access/serializer.py
wcmitchell: This should get automatically called any time the user submits a PUT request to the /cross-access-requests/<uuid>/ endpoint using the UpdateModelMixin.

wcmitchell: Oh, wait, disregard last. Wrong update method. let me re-check how this gets called.

rbac/api/cross_access/serializer.py
```diff
@@ -95,3 +95,18 @@ def create(self, validated_data):
...
             request.roles.add(role)
         return request
+
+    def update(self, instance, validated_data):
+        """Override the update method to associate the roles to cross account request after it is updated."""
+        if "roles" in validated_data:
+            role_data = validated_data.pop("roles")
+            display_names = [role["display_name"] for role in role_data]
+            roles = Role.objects.filter(display_name__in=display_names)
+            instance.roles.clear()
```
coderbydesign, line 105: Were we still thinking of wrapping this all in a transaction? Maybe the `update` method already does that?

rbac/api/cross_access/view.py
```diff
@@ -107,6 +112,52 @@ def list(self, request, *args, **kwargs):
...
+        validate_uuid(kwargs.get("pk"), "cross-account request uuid validation")
+
+        with tenant_context(Tenant.objects.get(schema_name="public")):
+            current = self.get_object()
+            if current.status != "pending":
+                self.throw_validation_error("cross-account update", "Only pending requests may be updated.")
+            if "target_account" in request.data and str(request.data.get("target_account")) != current.target_account:
+                self.throw_validation_error("cross-account-update", "Target account may not be updated.")
+
+            self.validate_and_format_input(request.data)
```
coderbydesign, line 152: I think I still have a question here around being able to update the `target_account`. If you supply it in a `PUT`, you can override it currently, correct? I feel like we may want to restrict that similar to the `PATCH`?

coderbydesign: Ah, line 149 may handle this now?

rbac/api/cross_access/model.py
```diff
@@ -56,9 +56,15 @@ def validate_input_value(self):
         if (
             isinstance(self.end_date, datetime.datetime)
             and isinstance(self.start_date, datetime.datetime)
-            and self.start_date.date() > self.end_date.date()
         ):
-            raise ValidationError("Start date must be earlier than end date.")
+            if self.start_date.date() > (datetime.datetime.now() + datetime.timedelta(60)).date():
```
coderbydesign, line 60: Good call on moving all the validation into the model.

wcmitchell: Thanks! Cleaned up a lot and made life a bit easier. May be worth checking which of the other validations we can move in as well.

rbac/api/cross_access/serializer.py
wcmitchell: Did a bit of digging, it's hard to determine if a serializer update() is automatically wrapped in a transaction. I see here [1] that some query attached updates do, but that doesn't seem to extend to an update of this style. I think I'll go ahead and wrap it and we'll see how it behaves.

[1]: https://docs.djangoproject.com/en/3.1/topics/db/transactions/

wcmitchell: A second bit of digging further down the page shows that we could enable ATOMIC_REQUESTS and have all of our HTTP requests take place in a transaction, but as that's a larger change, I'm still thinking I'll just wrap this for now.

rbac/api/cross_access/view.py
wcmitchell: I think this would be the best/easiest way, yep! Might be worth ensuring that only admins can set 'approved' in the status?

rbac/api/cross_access/view.py
wcmitchell: I *think* the serializer update() method can handle this now, right? If nothing comes in, that's fine, nothing happens and we return the same object?

rbac/api/cross_access/view.py
wcmitchell: That's my understanding too, yep!

astrozzc (approved):
Nice! LGTM, up to others to approve

rbac/api/cross_access/view.py
wcmitchell: Yep! You can supply `target_account` since it's a required field, but it has to be the same as originally created with. I could alternatively just grab the one from the current object and shim that in if we'd rather.

tests/api/cross_access/test_view.py
wcmitchell: Can do! I'll slip those in tomorrow

coderbydesign (approved):

PR #435, opened 2021-03-10
Add celery job for expiring cross-account requests

wcmitchell:

## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-11698

If an account's start date has passed and it's not approved, or if it's
end date has passed and it's not cancelled or expired, we'll set it to
expired. Currently, we'll check every day at 00:00:00 UTC for expired
requests.

## Local Testing
This will take a few steps, but here's how I tested it:
1. Pull down changes, reinitdb
2. Create some cross-account requests
    a. For me, this required altering /rbac/rbac/dev_middleware.py to reference a second account (12345)
    b. May also require adding some expected data to the user ref in dev_middleware, first name, last name, etc.
    c. Then hitting /api/rbac/v1/cross-account-requests/ to create multiple requests
3. Directly modify the DB (I used pgadmin) to ensure that some pending requests have passed start or end dates, and some accepted requests have passed end dates
4. Start a celery worker (requires redis running locally):
     celery worker -A rbac.celery --broker=redis://localhost:6379/0 --loglevel=INFO
5. Start a manage.py shell: cd rbac; python manage.py shell
6. Force run the celery process from the manage.py shell: 
    from api.tasks import cross_account_cleanup
    result = cross_account_cleanup.apply()
7. Celery will run through the existing requests and set status to expired where appropriate

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


rbac/api/cross_access/util.py
```diff
@@ -0,0 +1,53 @@
...
+@PROCESSING_TIME.time()
+def check_cross_request_expiry():
+    """Check if a cross-account requests have expired, tag them if so."""
+    with tenant_context(Tenant.objects.get(schema_name="public")):
+        expired_cars = []
+        cars = CrossAccountRequest.objects.filter(Q(status="pending") | Q(status="approved"))
+        logger.info("Running expiry check on %d cross-account requests.", len(cars))
+        for car in cars:
+            logger.debug("Checking for expiration of cross-account request %s.", car.pk)
+            if car.end_date < timezone.now() or (car.status == "pending" and car.start_date < timezone.now()):
```
astrozzc, line 45: Raise a discussion here: 
Associate might create a request starting from today or next day, asking for at least 30 day entry to target account. The account admin took some time (might be two or three days) to approve it but found it expired. Have to ask the associate to update it. Back and forth.
So maybe it is tolerable to to just simply expire only the things "car.end_date < timezone.now()"?

rbac/api/cross_access/util.py
```diff
@@ -0,0 +1,53 @@
...
+                logger.info("Expiring cross-account request with uuid: %s", car.pk)
+                car.status = "expired"
+                expired_cars.append(car.pk)
+                car.save()
```
astrozzc, line 49: A discussion here:
In the plan, the created principal have to be removed for approved cars. But that's when we planned principals have groups, policies roles attached to them. Should we keep them? seems no harm though.

rbac/api/cross_access/util.py
wcmitchell: That's actually what I had originally, @astrozzc, but then I noticed that the AC mentioned checking the start_date too. Personally I'm ok with just checking the end_date if you and @coderbydesign are.

rbac/api/cross_access/util.py
wcmitchell: Ah! I thought I'd responded to this earlier, but I think you're right, we should pull it out. I can get that added in real quick.

rbac/rbac/celery.py
```diff
@@ -38,7 +38,12 @@
         "task": "management.tasks.principal_cleanup",
         "schedule": crontab(0, 0, day_of_month="7-28/7"),
         "args": [],
-    }
+    },
+    "car-wash-daily": {
```
catastrophe-brandon, line 42: car-wash-daily? :)

catastrophe-brandon: Ah, Cross-Account Request (CAR). Ok, the name stays.

rbac/api/cross_access/util.py
```diff
@@ -0,0 +1,68 @@
...
+@PROCESSING_TIME.time()
+def check_cross_request_expiry():
+    """Check if a cross-account requests have expired, tag them if so."""
+    with tenant_context(Tenant.objects.get(schema_name="public")):
+        expired_cars = []
+        cars = CrossAccountRequest.objects.filter(Q(status="pending") | Q(status="approved"))
+        logger.info("Running expiry check on %d cross-account requests.", len(cars))
+        for car in cars:
+            logger.debug("Checking for expiration of cross-account request %s.", car.pk)
+            if car.end_date < timezone.now():
```
catastrophe-brandon, line 47: I haven't dug into the configuration, but I'm assuming we're using tz-aware objects for all the dates, right?

wcmitchell: We are, or at least I was previously getting complaints about comparing "timezone aware" and "timezone naive" objects, and those are now cleared up, so I think we're save

rbac/api/cross_access/util.py
astrozzc: 👍Thanks! And this actually reminds me missing sth😅 https://github.com/RedHatInsights/insights-rbac/pull/439

rbac/api/cross_access/util.py
```diff
@@ -0,0 +1,68 @@
...
+    "cross_account_expiry_processing_second", "Time spent checking and expiring cross-account requests"
+)
+
+
+@PROCESSING_TIME.time()
+def check_cross_request_expiry():
+    """Check if a cross-account requests have expired, tag them if so."""
+    with tenant_context(Tenant.objects.get(schema_name="public")):
+        expired_cars = []
+        cars = CrossAccountRequest.objects.filter(Q(status="pending") | Q(status="approved"))
```
coderbydesign, line 43: Just to clarify, this will/should exclude running this against cars with status set to "cancelled", "denied" or "expired", correct?

rbac/api/cross_access/util.py
coderbydesign: Also, would it be a concern if a principal has multiple cars, and only one has expired?

coderbydesign: Does that mean `remove_cross_principal` goes away based on this conversation?

rbac/api/cross_access/util.py
```diff
@@ -0,0 +1,68 @@
...
+
+        logger.info(
+            "Completed clean up of %d cross-account requests, %d expired.", len(cars), len(expired_cars),
+        )
+
+
+def remove_cross_principal(target_account, user_id):
+    """Remove a cross account principal in the target account."""
+    # Principal has the pattern acctxxx-123456.
+    principal_name = f"{target_account}-{user_id}"
```
coderbydesign, line 62: If we _do_ keep this method, maybe it's worth breaking this out into a util method that could also be used in `/rbac/api/cross_access/view.py:234`?

rbac/api/cross_access/util.py
wcmitchell: Correct, the Q calls build a filter for `status = pending OR approved` that django can understand

rbac/api/cross_access/util.py
astrozzc: Ah, good point! I thought there would be only one approved request in one account for one associate, but seems we do support multiple requests. Maybe we could check if there's any approved left before removing them?

coderbydesign: Or maybe, check that all are "expired"? If not, leave it? Also, historically would we want to keep referential integrity for audit purposes, similar to how we aren't going to actually delete the cars?

rbac/api/tasks.py
```diff
@@ -0,0 +1,28 @@
...
+
+from celery import shared_task
+
+from api.cross_access.util import check_cross_request_expiry
+
+
+@shared_task
+def cross_account_cleanup():
+    """Celery task to mark expired cross-account requests."""
+    check_cross_request_expiry()
```
coderbydesign, line 28: Came across a situation locally testing, where things appeared to not be working when I ran this. It turned out to be that it was failing silently because I didn't setup the principal (I didn't update the car to "approved" via the API, I did it directly in postgres, so it never provisioned it). 

Since it runs in a background process, the exceptions never get raised/logged, it just silently stops. Maybe adding a try/except here at least to log any exceptions would be helpful? And/or a metric (here or in the actual method) so we don't get a false positive that it's running, but something is actually failing to delete them? I was confused because my cars were not updating to expired.

rbac/api/cross_access/util.py
coderbydesign, line 43: Actually, thinking about my [comment here](https://github.com/RedHatInsights/insights-rbac/pull/435/files#r596800512), I believe anything that's `pending` will actually fail, because a principal will have never been created for them, correct? So when `remove_cross_principal` is run, it will currently fail silently on anything that's `pending` with `Principal matching query does not exist.`

coderbydesign (commented):
Hey, @wcmitchell! Left a few comments, but also wanted to note that I did see the metrics against: `http://localhost:8000/metrics`, but I'm not sure it's what's expected?

```
# HELP cross_account_expiry_processing_second Time spent checking and expiring cross-account requests
# TYPE cross_account_expiry_processing_second summary
cross_account_expiry_processing_second_count 0.0
cross_account_expiry_processing_second_sum 0.0
```

I added a sleep in the method as well, but didn't see the metrics bump. Based on the docs, I would expect to see `cross_account_expiry_processing_second_sum` record the number of times the method was called, and `cross_account_expiry_processing_second_sum` give me the total time spent? I wonder if it has to do with it running outside the app process.

rbac/api/cross_access/util.py
wcmitchell: I think this is a good idea, I'd started to refactor it out into a util method, but wasn't sure it'd get used elsewhere. I'll set that up.

rbac/api/tasks.py
wcmitchell: That's a great idea, I'll work on getting that added in!

rbac/api/cross_access/util.py
coderbydesign: @wcmitchell - @astrozzc and I were discussing more in the dev sync, and it seems like we may just want to leave the principals for now, for historical reference and to ensure we don't remove a principal who still has CARs tied to it.

wcmitchell: That seems reasonable. We'd still need to remove the roles from the principal, I guess? Or...hrm. We'll need to do something to actually shut down the access, right?

rbac/api/cross_access/model.py
```diff
@@ -51,7 +51,8 @@ def validate_input_value(self):
         if self.status not in STATUS_LIST:
             raise ValidationError(f'Unknown status "{self.status}" specified, {STATUS_LIST} are valid inputs.')
 
-        [self.validate_date(date) for date in [self.start_date, self.end_date]]
+        if self.status != "expired":
```
astrozzc, line 54: Overall LGTM. Maybe I missed sth, but if the CAR status is "pending" and gets updated, it would skip this date validation here, which seems not desired

wcmitchell: It's been a minute, but if I remember right I had to put this in because otherwise, when we set a requests's status to `expired`, we'd get complaints that the dates are past. Those dates being past is expected, though. since the request is expiring. With this check, any request that's not already in the `expired` state should still get the "in the past" check.

-

rbac/api/cross_access/util.py
```diff
@@ -0,0 +1,87 @@
...
+    # Principal would have the pattern acctxxx-123456.
+    principal_name = get_cross_principal_name(target_account, user_id)
+    tenant_schema = create_schema_name(target_account)
+    with tenant_context(Tenant.objects.get(schema_name=tenant_schema)):
+        cross_account_principal = Principal.objects.get_or_create(username=principal_name, cross_account=True)
+
+    return cross_account_principal
+
+
+def remove_cross_principal(target_account, user_id):
```
coderbydesign, line 73: Can we pull this method out now?

rbac/api/cross_access/util.py
```diff
@@ -0,0 +1,87 @@
...
+from prometheus_client import Counter, Summary
+from tenant_schemas.utils import tenant_context
+
+from api.models import CrossAccountRequest, Tenant
+from api.serializers import create_schema_name
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+# Create processing time metric
+PROCESSING_TIME = Summary(
```
coderbydesign, line 33: Given the conversation around not addressing metrics in this PR but creating another ticket for it, should we just remove any metric related code from this PR?

rbac/api/cross_access/model.py
astrozzc: Ah, I was out of my mind, I thought it was == "expired"

astrozzc (approved):
LGTM

rbac/api/cross_access/util.py
wcmitchell: It's never called, but I had left it in in case we wanted it some time in the future. I can pull it, though, probably safer.

coderbydesign: Yeah, was just thinking it might be confusing if we come across that but forget the context and can't recall why/where we're calling it. :)

coderbydesign (approved):

PR #439, opened 2021-03-17
Skip principal cleanup for cross account principals

astrozzc:
For cross account principals, they are only stored in RBAC,
Therefore, the cleaner would clean those principals since they could not
be found against IT.

## Link(s) to Jira
- 

For cross account principals, they are only stored in RBAC,
Therefore, the cleaner would clean those principals since they could not
be found against IT.

## Local Testing
Unit tests covers the case.

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/management/utils.py
+++ b/rbac/management/utils.py
@@ -202,8 +202,10 @@ def roles_for_cross_account_principal(principal):
     """Return roles for cross account principals."""
     target_account, user_id = principal.username.split("-")
     with tenant_context(Tenant.objects.get(schema_name="public")):
-        cross_account_request = CrossAccountRequest.objects.filter(
+        cross_account_requests = CrossAccountRequest.objects.filter(
             target_account=target_account, user_id=user_id, status="approved"
-        ).first()
-        role_names = list(cross_account_request.roles.all().values_list("name", flat=True))
+        )
+        role_names = []
+        for car in cross_account_requests:
+            role_names.extend(list(car.roles.all().values_list("name", flat=True)))
     return Role.objects.filter(name__in=role_names)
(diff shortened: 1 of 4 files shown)


rbac/management/utils.py
```diff
@@ -202,8 +202,10 @@ def roles_for_cross_account_principal(principal):
...
     with tenant_context(Tenant.objects.get(schema_name="public")):
-        cross_account_request = CrossAccountRequest.objects.filter(
+        cross_account_requests = CrossAccountRequest.objects.filter(
             target_account=target_account, user_id=user_id, status="approved"
-        ).first()
-        role_names = list(cross_account_request.roles.all().values_list("name", flat=True))
+        )
+        role_names = []
+        for car in cross_account_requests:
+            role_names.extend(list(car.roles.all().values_list("name", flat=True)))
```
coderbydesign, line 210: This will return duplicate names if the same role exists in multiple CARs, currently. Probably not a big deal since the filter below will be unique, but wanted to mention it.

I believe we could actually do something like this, and replace:
```
        cross_account_requests = CrossAccountRequest.objects.filter(
            target_account=target_account, user_id=user_id, status="approved"
        )
        role_names = []
        for car in cross_account_requests:
            role_names.extend(list(car.roles.all().values_list("name", flat=True)))
    return Role.objects.filter(name__in=role_names)
```
with this:
```
        role_names = CrossAccountRequest.objects.filter(
            target_account=target_account,
            user_id=user_id,
            status="approved"
        ).values_list("roles__name", flat=True).distinct()
    return Role.objects.filter(name__in=list(role_names))
```
Would need to be tested, though. Just an idea. :)

astrozzc: Nice! Updated

wcmitchell (approved):
👍

coderbydesign (approved):

PR #447, opened 2021-03-23
Fix access permission matching

coderbydesign:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-12725

There is currently a bug where permissions
will be returned against the `/access/` endpoint if a given permission's `application`
value matches a substring of the `?application=` param.

For instance, given the following permission:
```
{application: "foo"}
```

and the query string:
```
/rbac/v1/access/?application=foobar
```

we should *not* return any permissions, as they should be an exact match between
the param and the application name on the permission.

With the change to support multiple application values (81eb66f)
we introduced an edge-case where the match could be a substring of the paramter.
This did not affect supplying a substring as a param, like:
```
/rbac/v1/access/?application=fo
```
which would still return no results.

The proposed fix will match on each split param value in `application=` to ensure
exact matches.

## Local Testing
You can test this on `master` by seeding your database and querying:
```
http://localhost:8000/api/rbac/v1/access/?application=inventoryisgreat
```

Note that you'll get back inventory permissions when you shouldn't. After checking
out this feature branch, you should get no permissions.

All other current behavior should remain unchanged.

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/management/utils.py
+++ b/rbac/management/utils.py
@@ -21,7 +21,7 @@ from uuid import UUID
 
 from django.core.exceptions import PermissionDenied
 from django.utils.translation import gettext as _
-from management.models import Group, Principal, Role
+from management.models import Access, Group, Principal, Role
 from management.principal.proxy import PrincipalProxy
 from rest_framework import serializers, status
 from tenant_schemas.utils import tenant_context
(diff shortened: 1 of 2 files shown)


astrozzc (approved):
LGTM

rbac/management/utils.py
```diff
@@ -104,11 +104,11 @@ def access_for_roles(roles, param_applications):
     """Gathers all access for the given roles and application(s)."""
     access = []
     for role in set(roles):
-        role_access = set(role.access.all())
-        for access_item in role_access:
-            if (not param_applications) or (access_item.permission_application() in param_applications):
-                access.append(access_item)
-    return access
+        if param_applications:
+            access += Access.objects.filter(role=role, permission__application__in=param_applications.split(","))
```
coderbydesign, line 108: Split `param_applications.split(",")` once in the beginning and reuse each iteration.

coderbydesign: This is resolved now!

wcmitchell (approved):
👍

PR #449, opened 2021-03-24
Ensure `DJANGO_DEBUG` is set to "False" by default

coderbydesign:
We have an explicit check on `DJANGO_DEBUG` being set to "False" [1]. I noticed
the way the Clowder config is setup has enabled debug in ci/qa/stage. We need
to ensure this goes in before we release the Clowder version or RBAC into prod.

[1] https://github.com/RedHatInsights/insights-rbac/blob/49909263a7aecf86313c87d3fc035acae1d953d7/rbac/rbac/settings.py#L76


wcmitchell (approved):
👍

PR #450, opened 2021-03-25
Temporarily disable haberdasher/platform logging

coderbydesign:
We need more verbosity in our production logs to help investigate a transient issue
that only surfaces in production.


wcmitchell (approved):
👍

PR #465, opened 2021-04-14
Filter sort

astrozzc:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-13378

Based on the mock, the status should all be able to be filtered AND fields such as request id, created/start/end date sorted: https://marvelapp.com/prototype/257je526/screen/74609498

## Local Testing
Unit tests

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -93,6 +93,38 @@
                 "true"
               ]
             }
+          },
+          {
+            "in": "query",
+            "name": "status",
+            "required": false,
+            "description": "Parameter for filtering resource based on status.",
+            "schema": {
+              "type": "string",
+              "enum": [
+                "pending",
+                "approved",
+                "denied",
+                "cancelled",
+                "expired"
+              ]
+            }
+          },
+          {
+            "in": "query",
+            "name": "order_by",
+            "required": false,
+            "description": "Parameter for ordering by field. For inverse ordering, use '-', e.g. ?order_by=-start_date. Default is -created",
+            "schema": {
+              "type": "string",
+              "enum": [
+                "request_id",
+                "start_date",
+                "end_date",
+                "created",
+                "modified"
+              ]
+            }
           }
         ],
         "responses": {
@@ -3254,7 +3286,7 @@
               "cancelled",
               "denied"
             ]
-          }  
+          }
         }
       },
       "CrossAccountRequestPatch": {
@@ -3283,7 +3315,7 @@
               "cancelled",
               "denied"
             ]
-          }  
+          }
         }
       },
       "CrossAccountRequestByAccount": {
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -114,7 +114,7 @@
             "in": "query",
             "name": "order_by",
             "required": false,
-            "description": "Parameter for ordering by field. For inverse ordering, use '-', e.g. ?order_by=-start_date. Default is -created",
+            "description": "Parameter for ordering by field. For inverse ordering, use '-', e.g. ?order_by=-start_date.",
             "schema": {
               "type": "string",
               "enum": [
--- a/rbac/api/cross_access/view.py
+++ b/rbac/api/cross_access/view.py
@@ -24,6 +24,7 @@ from management.principal.proxy import PrincipalProxy
 from management.utils import validate_and_get_key, validate_limit_and_offset, validate_uuid
 from rest_framework import mixins, viewsets
 from rest_framework import status as http_status
+from rest_framework.filters import OrderingFilter
 from rest_framework.response import Response
 from rest_framework.serializers import ValidationError
 from tenant_schemas.utils import tenant_context
@@ -82,8 +83,10 @@ class CrossAccountRequestViewSet(
     """
 
     permission_classes = (CrossAccountRequestAccessPermission,)
-    filter_backends = (filters.DjangoFilterBackend,)
+    filter_backends = (filters.DjangoFilterBackend, OrderingFilter)
     filterset_class = CrossAccountRequestFilter
+    ordering_fields = ("request_id", "start_date", "end_date", "created", "modified")
+    ordering = "-created"
 
     def get_queryset(self):
         """Get query set based on the queryBy key word."""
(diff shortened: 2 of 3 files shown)


coderbydesign:
Looks like there are a few test failures?

astrozzc:
> Looks like there are a few test failures?

Ah, thanks. Yea, due to the change of default order of return, some test check order is changed.

rbac/api/cross_access/view.py
```diff
@@ -82,8 +83,10 @@ class CrossAccountRequestViewSet(
     """
 
     permission_classes = (CrossAccountRequestAccessPermission,)
-    filter_backends = (filters.DjangoFilterBackend,)
+    filter_backends = (filters.DjangoFilterBackend, OrderingFilter)
     filterset_class = CrossAccountRequestFilter
+    ordering_fields = ("request_id", "start_date", "end_date", "created", "modified")
```
redallen, line 88: Can you also allow ordering by first_name and last_name for the approver view?

redallen (changes requested):

docs/source/specs/openapi.json
```diff
@@ -93,6 +93,38 @@
...
               ]
             }
+          },
+          {
+            "in": "query",
+            "name": "status",
+            "required": false,
+            "description": "Parameter for filtering resource based on status.",
+            "schema": {
+              "type": "string",
```
redallen, line 103: Can you allow filtering by multiple statuses? This is the way the UI is currently setup:
image

this calls: `/api/rbac/v1/cross-account-requests/?query_by=user_id&offset=0&limit=10&status=approved%2Cdenied%2Ccancelled&sort_by=created&sort_direction=desc` which I'd expect to return approved, denied, and cancelled requests.

docs/source/specs/openapi.json
```diff
@@ -93,6 +93,38 @@
...
+          },
+          {
+            "in": "query",
+            "name": "order_by",
+            "required": false,
+            "description": "Parameter for ordering by field. For inverse ordering, use '-', e.g. ?order_by=-start_date. Default is -created",
+            "schema": {
+              "type": "string",
+              "enum": [
+                "request_id",
```
redallen, line 121: Can you allow ordering by first name and last name as well?

rbac/api/cross_access/view.py
```diff
@@ -82,8 +83,10 @@ class CrossAccountRequestViewSet(
...
+    ordering = "-created"
```
coderbydesign, line 89: While I do see the argument of having this default to DESC ordering on `created` for implementation in the UI, I wonder if this would set a precedence that we don't yet have in the other API endpoints, which may lead to some confusion? As Zack mentioned in his initial Slack comment, as long as we allow `order_by=-created` (which would be the case here) that would be enough? We currently have default ordering on most APIs set alphabetically. I know that doesn't make sense here, but I feel like it would be less confusing to have more consistent defaults, allowing for custom sorting, vs adhoc defaults per endpoint?

rbac/api/cross_access/view.py
astrozzc: Unfortunately, that's nasty implementation because we only store the user_id on our side, we fetch the first_name and last_name, then we have to sort the result dict by first_name or last_name.

We still could do that but just not a clean way (sorting a dict doesn't feel like a right thing once the number of request gets large). If that's really desired feature, I'd defer that to the future lease.

docs/source/specs/openapi.json
astrozzc: Sure, will add this

rbac/api/cross_access/view.py
astrozzc: Yep, agree. Will remove the default

rbac/api/cross_access/view.py
redallen: After taking a [closer look at the designs](https://marvelapp.com/prototype/257je526/screen/74764732) it looks like you're right and I'll remove being able to sort those 2 columns.

astrozzc: Cool! BTW, I just look at the design and it requires ability to filter by first/last name. This would encounter the same situation: we search for all the users (db query) and replace the user_id with user's info (call BOP and do it in rbac). So if we want to filter by first/last name, we have to do that for the return result (not like other filter that performs in the db query). Therefore, it would be no better than just get all the request result and search for it on the UI side. I would defer that to future release too if that's really desired. We could discuss with Margot about it

redallen: Okay, I can easily remove those filters. Is filtering by first/last name in the approver view essential @mmenestr?

mmenestr: It is pretty essential in that it's the only way to search for _who_ sent you a request! So I would say, please implement it if you can 👍

coderbydesign: If there's a way to handle this in the UI even as an interim solution, that may be the best. Otherwise as Jay mentions, we'll quickly run into issues with pagination (with potential for timeout from IT). If so, we can/should log something to address this in a future iteration. It's a non-trivial amount of work, and something we have identified as a limitation for some of the other /principals/ endpoints as well.

redallen: All I can do on the frontend is filter the 10-100 rows that are on the page (and from my understanding, your scaling problem still applies to large page sizes since you'll have to make 100 requests to the BOP for a page size of 100). This will be a confusing filter since it does not behave like the other filters. Can you think of any other ways in the UI to handle not being able to filter by first/last name on the backend @coderbydesign or @mmenestr ?

coderbydesign: Based on slack conversation, we'll not be implementing this on this iteration.

coderbydesign (approved):

wcmitchell (approved):
LGTM now!

PR #464, opened 2021-04-12
Associate models with tenant

coderbydesign:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-11196

## Description of Intent of Change(s)
This adds a `TenantAwareModel` class, which inherits from `model.Model`. This allows
us to reuse this class on all models which will have a relation to the `Tenant`
model, meaning we will just need to have those models inherit from `TenantAwareModel`
to pick up any behavior related to the tenant relation.

The following models will now inherit from `TenantAwareModel`:

- access
- group
- permission
- policy
- principal
- resourcedefinition
- role

## Local Testing
Functionally, there should be no change. This is strictly the model code for #430

## Checklist
- [ ] if API spec changes are required, is the spec updated?
- [ ] are there any pre/post merge actions required? if so, document here.
- [ ] are theses changes covered by unit tests?
- [ ] if warranted, are documentation changes accounted for?
- [ ] does this require migration changes?
  - [ ] if yes, are they backwards compatible?
- [ ] is there known, direct impact to dependent teams/components?
  -  if yes, how will this be handled?

## Secure Coding Practices Checklist Link
- https://github.com/RedHatInsights/secure-coding-checklist

## Secure Coding Practices Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [ ] General Coding Practices

PR #467, opened 2021-04-14
Associate with tenant on create

coderbydesign:
_**NOTE:** Relies on #464 being merged before review! Drafting until then._

## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-11196

This is a follow-up to #464

**Override `create` method in `ModelSerializer`** - f02f35a
This overrides the `create` method in the `ModelSerializer` to ensure that we are
able to set the `tenant_id` for any object created in the API.

We'll still need to handle seeds by passing in the tenant, and any other object
creation that happens outside of views. In the future, we'll want to enforce this
via RLS.

We cannot add the `tenant_id` in a signal, because we don't have access to the
request, where the tenant exists. We also cannot do this directly in the views,
since it would require we expose the `tenant_id` field in the serializers which
we don't want to do.

Instead, this provides us the opportunity to add the `tenant_id` after serialization
has succeeded, but before the object is saved. This is abstracted out so we can
include it as a mixin to the serializer classes which inherit from `ModelSerializer`.

The existing behavior is copied directly from our current version 3.12.2 https://github.com/encode/django-rest-framework/blob/3.12.2/rest_framework/serializers.py#L904
and injects the `tenant_id` in the `try` block.

**Ensure `tenant_id` is saved on all creates** - e36b34a
This makes use of the `SerializerCreateOverrideMixin` in order to add the behavior
of saving the `tenant_id` from the request object on creates.

There are also some instances where serializers implement their own create, or
are not model serializers. In these cases, and other cases where we're doing an
explicit `create()` or `save()` we also need to ensure the `tenant_id` is set.

For `get_or_create()` calls, currently we don't have `tenant_id` set on all objects.
If we add `tenant=tenant` directly into these calls, we'd create duplicate records
since the `get` would not match on the tenant identifier. To get around this, we'll
need to update the object after `get_or_create()` to add the `tenant_id` value if
it does not yet exist.

We should be able to pull these noted references out after we're strictly enforcing
the FK constraints and have confirmed all records have `tenant_id` values. At that
point we should be able to add `tenant=tenant` to the `get_or_create()` directly.

One other consideration is that this commit may be something we'll want to revert
if/when we have RLS (row-level security) in place. If we do it correctly, we would
ideally have RLS policies setup for all operations (SELECT, INSERT, etc), which would
mean we'd no longer need to update all queries directly with the tenant, as the
policy would set the `tenant_id` in the middleware.

**Post-deploy**
After this is merged/deployed, we'll need to run the following script to populate `tenant_id` 
on historical data. After a a period of time we should verify that all records in all
schemas (including public) have `tenant_id`s populated.

## Local Testing
1) To first test the post-deploy script, run it against an existing dataset. If you have data
in your local db already, you can do it on this branch. Otherwise, reinit the db on `master`
and run the following:

Usage:
```
-- to run reconciliation as read-only:
./rbac/manage.py reconcile_tenant_relations --readonly

-- to run reconciliation and apply updates for missing `tenant_id` relations:
./rbac/manage.py reconcile_tenant_relations
```

You should see output about setting the `tenant_id`. If the script runs, you can flip
`READ_ONLY = False` and run it again. This will actually make the update. Running
it again should confirm that you have no records without a `tenant_id` (you can also check
the db directly).

2) To confirm that creating new and nested resources associates `tenant` objects,
you can exercise `POST`/create requests  on the API:

```
- curl http://localhost:8000/api/rbac/v1/groups/ -H 'Content-Type: application/json' -d '{"name": "test group"}'
- curl http://localhost:8000/api/rbac/v1/groups/<UUID>/principals/ -H 'Content-Type: application/json' -d '{"principals": [{"username": "test principal in group"}]}'
- curl http://localhost:8000/api/rbac/v1/groups/<UUID>/roles/ -H 'Content-Type: application/json' -d '{"roles": ["<UUID>"]}'
- curl http://localhost:8000/api/rbac/v1/roles/ -H 'Content-Type: application/json' -d '{"name": "test role", "access": [{"permission": "sources:*:*", "resourceDefinitions": [{"attributeFilter": {"key": "foo-key", "operation": "equal", "value": "1234"}}]}]}'
- curl http://localhost:8000/api/rbac/v1/policies/ -H 'Content-Type: application/json' -d '{"name": "test policy", "group": "<UUID>", "roles": ["<UUID>"]}'

```

## Checklist
    - [ ] the documented script above needs to be run

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/api/cross_access/util.py
+++ b/rbac/api/cross_access/util.py
@@ -43,9 +43,7 @@ def check_cross_request_expiry():
                 expired_cars.append(car.pk)
                 car.save()
 
-        logger.info(
-            "Completed clean up of %d cross-account requests, %d expired.", len(cars), len(expired_cars),
-        )
+        logger.info("Completed clean up of %d cross-account requests, %d expired.", len(cars), len(expired_cars))
 
 
 def create_cross_principal(target_account, user_id):
@@ -53,8 +51,17 @@ def create_cross_principal(target_account, user_id):
     # Principal would have the pattern acctxxx-123456.
     principal_name = get_cross_principal_name(target_account, user_id)
     tenant_schema = create_schema_name(target_account)
-    with tenant_context(Tenant.objects.get(schema_name=tenant_schema)):
-        cross_account_principal = Principal.objects.get_or_create(username=principal_name, cross_account=True)
+    tenant = Tenant.objects.get(schema_name=tenant_schema)
+    with tenant_context(tenant):
+        cross_account_principal, _ = Principal.objects.get_or_create(username=principal_name, cross_account=True)
+
+        # NOTE: after we ensure/enforce all object have a tenant_id FK, we can add tenant=tenant
+        # to the get_or_create. We cannot currently, because records without would fail the GET
+        # and would create duplicate records. This ensures we temporarily do an update if
+        # obj.tenant_id is NULL
+        if not cross_account_principal.tenant:
+            cross_account_principal.tenant = tenant
+            cross_account_principal.save()
 
     return cross_account_principal
 
--- a/rbac/management/group/definer.py
+++ b/rbac/management/group/definer.py
@@ -42,9 +42,17 @@ def seed_group(tenant):
                 platform_default=True, defaults={"description": group_description, "name": name, "system": True}
             )
 
+            # NOTE: after we ensure/enforce all object have a tenant_id FK, we can add tenant=tenant
+            # to the get_or_create. We cannot currently, because records without would fail the GET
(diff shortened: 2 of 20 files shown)


wcmitchell (approved):
Seems to all work as detailed, good start to ensuring that everything new has a tenant ID attached. 👍

PR #469, opened 2021-04-16
Ensure Role objects have `modified` updated when seeds change

coderbydesign:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-13355

When a `Role` object is updated via seeds and the defaults on the actual object
change, the `modified` field will update via Django. The same is true if `Access`
objects are added/removed to/from a `Role` via the API.

However, when seeding a change to `Access` (permissions) only on a role, and the
version number is bumped, we update the `Role` with defaults and in this case it
would have no diff, so Django would not update the object, thus not updating the
`modified` field. We then remove and add the `Access` objects outside the context
of an API resource and update the model directly, which does not update the `Role`
either.

This explicitly sets the `modified` field to `timezone.now()` which is the default
on the model, whenever the version changes in the seed files. This will ensure it
updates regardless of whether or not the `Role` data itself is updated.

## Local Testing
- On `master`, update a version and permission (not the role data itself) in a local `/rbac/role/definitions/*.json` file
- Run seeds and note that the role you've updated doesn't have its `modified` value updated
- Checkout this feature branch
- Again update a version and permission (not the role data itself) in a local `/rbac/role/definitions/*.json` file
- Run seeds and note that the role's `modified` value should now be updated

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/management/role/definer.py
+++ b/rbac/management/role/definer.py
@@ -22,6 +22,7 @@ import os
 
 from django.conf import settings
 from django.db import transaction
+from django.utils import timezone
 from management.permission.model import Permission
 from management.role.model import Access, ResourceDefinition, Role
 from tenant_schemas.utils import tenant_context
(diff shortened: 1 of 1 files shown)


astrozzc (approved):

wcmitchell (approved):
👍

PR #471, opened 2021-04-20
Add in uniqueness constraints

wcmitchell:
Adding unique-per-tenant constraints to existing models to prep for
moving to public schema

## Link(s) to Jira
- 

When we copy the currently modified roles, groups, etc to the public schema, we risk data collisions as the uniqueness constraints don't take into account the new tenant field. This migration adds on a dual constraint between unique fields and tenant ids. 

## Local Testing
As this doesn't directly remove the existing unique constraint on given fields, it won't be inherently testable until we migrate data to the public schema. I think locally it could be testable by removing the unique=True on any of the affected fields, ensuring that objects in those fields have tenant_ids attached (this may require direct DB manipulation), and then attempting to create another object with the same name and tenant id, which should fail

## Checklist
- [ x ] does this require migration changes?

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


astrozzc (approved):
Tested locally, works as expected.

rbac/management/role/serializer.py
```diff
@@ -169,7 +164,7 @@ def update(self, instance, validated_data):
         for access_item in access_list:
             resource_def_list = access_item.pop("resourceDefinitions")
             access_permission = access_item.pop("permission")
-            permission, created = Permission.objects.get_or_create(**access_permission)
+            permission, created = Permission.objects.get_or_create(**access_permission, tenant=tenant)
```
coderbydesign, line 167: Do we need to be concerned with creating duplicates here prior to the tenant_id sync job being run without something like this? https://github.com/RedHatInsights/insights-rbac/pull/467/files#diff-ac43ba775b89d5e57f70fdea347fe345ab148950cf8650dd14f1af761edd07e2R139-R149

rbac/management/migrations/0034_remove_field_uniqueness.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 2.2.4 on 2021-04-23 13:08
```
coderbydesign, line 1: I think we probably do still want to remove these existing constraints, but I'm just curious if the serializer validation fix you made in here solves the issue we were hitting with this? Just wondered if it was a red herring initially?

wcmitchell: I went through a few iterations on how best to handle this, but overall I do think this solves the issue we were initially hitting. It also moves the burden from enforcing uniqueness to the DB rather than the serializer, which seems like it should focus more on data validation than enforcing uniqueness rules.

rbac/management/role/serializer.py
wcmitchell: That's a good point, and there likely is a risk here. I could try to write up a version where we do a get, then check if we got anything, and if not then we create with the tenant. It'd be easiest if we could push up the sync command first, but it won't work without these in place, I fear.

rbac/management/migrations/0034_remove_field_uniqueness.py
coderbydesign: Sounds good!

PR #473, opened 2021-04-22
Set celery probe timers using parameters

bsquizz:
So that we can test lowering them in different environments


astrozzc (approved):
LGTM

coderbydesign (approved):

PR #483, opened 2021-05-04
Sync objects

astrozzc:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-11197

Once the tenant_id relation is setup, we should sync the public and tenant schemas moving forward by ensure any non-get operation will affect both the public schema and the tenant schema.

This will keep feature and data parity moving forward, and should not affect the current process of creating tenants and storing/serving data from those tenant schemas. This will add an additional step of performing the same operation on the public schema.

## Local Testing
Unit tests

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/management/role/serializer.py
+++ b/rbac/management/role/serializer.py
@@ -19,9 +19,11 @@
 from django.utils.translation import gettext as _
 from management.group.model import Group
 from management.serializer_override_mixin import SerializerCreateOverrideMixin
-from management.utils import get_principal_from_request
+from management.utils import create_object_in_tenant, get_principal_from_request
 from rest_framework import serializers
+from tenant_schemas.utils import tenant_context
 
+from api.models import Tenant
 from .model import Access, Permission, ResourceDefinition, Role
 
 ALLOWED_OPERATIONS = ["in", "equal"]
@@ -127,15 +129,11 @@ class RoleSerializer(serializers.ModelSerializer):
         access_list = validated_data.pop("access")
         tenant = self.context["request"].tenant
         role = Role.objects.create(name=name, description=description, display_name=display_name, tenant=tenant)
-        role.save()
-        for access_item in access_list:
-            resource_def_list = access_item.pop("resourceDefinitions")
-            access_permission = access_item.pop("permission")
-            permission, created = Permission.objects.get_or_create(**access_permission)
+        role_public, _ = create_object_in_tenant(
+            "public", tenant, Role, **{"name": name, "description": description, "display_name": display_name}
+        )
+        create_access_for_role(role, role_public, access_list, tenant)
 
-            access_obj = Access.objects.create(permission=permission, role=role, tenant=tenant)
-            for resource_def_item in resource_def_list:
-                ResourceDefinition.objects.create(**resource_def_item, access=access_obj, tenant=tenant)
         return role
 
     def update(self, instance, validated_data):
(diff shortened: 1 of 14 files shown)


rbac/management/utils.py
```diff
@@ -213,3 +216,19 @@ def roles_for_cross_account_principal(principal):
         )
         role_names_list = list(role_names)
     return Role.objects.filter(name__in=role_names_list)
+
+
+def create_object_in_tenant(schema_name, associate_tenant, model, **kwargs):
+    """Create object based on model in tenant schema."""
+    tenant = Tenant.objects.get(schema_name=schema_name)
+    with tenant_context(tenant):
+        model_object, created = model.objects.get_or_create(**kwargs)
```
coderbydesign, line 225: After the reconciliation script is run, we'll want this to include the tenant in the get_or_create.

rbac/management/role/serializer.py
```diff
@@ -337,3 +327,29 @@ def obtain_groups_in(obj, request):
...
 
     return Group.objects.filter(policies__in=policy_ids).distinct()
+
+
+def create_access_for_role(role, role_public, access_list, tenant):
+    """Create access objects and relate it to role."""
+    for access_item in access_list:
+        resource_def_list = access_item.pop("resourceDefinitions")
+        access_permission = access_item.pop("permission")
+        permission, _ = create_object_in_tenant(tenant.schema_name, tenant, Permission, **access_permission)
```
coderbydesign, line 306: Per @astrozzc, we'll want to make sure this is a get only.

coderbydesign:
@astrozzc I don't think this works out of the box, but it would be awesome if we were able to use a similar pattern to this:
```python
import contextlib
from management.models import Principal
from api.models import Tenant

# this would provide as a handler to yield/run the same command
# on both the public schema, and the schema that's passed in,
# without changing the implementation logic in the original method
@contextlib.contextmanager
def schema_handler(tenant_schema):
    """Handle events in both public and tenant schemas."""
    public_schema = Tenant.objects.get(schema_name=schema_name)
    for schema in [public_schema, tenant_schema]:
        yield

# as we need to sync in both schemas, this could facilitate that
# by allowing us reuse the same code for storing/updating/deleting
# by calling a handler with the tenant schema
def some_other_method():
...
...
with schema_handler(Tenant.objects.get(schema_name="acct10001")):
    Principal.objects.get_or_create(username="username", cross_account=True)

# once we serve from public, we could just remove the `with schema_handler()`
# call in all places, after we set the schema to public, and not have to change
# any of the actual logic or calls
def some_other_method():
...
...
Principal.objects.get_or_create(username="username", cross_account=True)
```
Again, not sure how easy it would be, but something where we could basically just use a wrapper.

astrozzc:
> @astrozzc I don't think this works out of the box, but it would be awesome if we were able to use a similar pattern to this:
> 
> ```python
> import contextlib
> from management.models import Principal
> from api.models import Tenant
> 
> # this would provide as a handler to yield/run the same command
> # on both the public schema, and the schema that's passed in,
> # without changing the implementation logic in the original method
> @contextlib.contextmanager
> def schema_handler(tenant_schema):
>     """Handle events in both public and tenant schemas."""
>     public_schema = Tenant.objects.get(schema_name=schema_name)
>     for schema in [public_schema, tenant_schema]:
>         yield
> 
> # as we need to sync in both schemas, this could facilitate that
> # by allowing us reuse the same code for storing/updating/deleting
> # by calling a handler with the tenant schema
> def some_other_method():
> ...
> ...
> with schema_handler(Tenant.objects.get(schema_name="acct10001")):
>     Principal.objects.get_or_create(username="username", cross_account=True)
> 
> # once we serve from public, we could just remove the `with schema_handler()`
> # call in all places, after we set the schema to public, and not have to change
> # any of the actual logic or calls
> def some_other_method():
> ...
> ...
> Principal.objects.get_or_create(username="username", cross_account=True)
> ```
> 
> Again, not sure how easy it would be, but something where we could basically just use a wrapper.

Yea, let me check. I think it would be cleaner to have such a method

coderbydesign (approved):
Nice work! This will make things much nicer to manage once we flip to the public schema. :) Functionally tested against:

roles:
- create
- update
- delete

groups:
- create
- update
- delete

- add/remove roles to/from groups
- add/remove principals to/from groups

PR #484, opened 2021-05-11
Remove tenant_id associations on Permissions

coderbydesign:
Permissions are not able to currently be created via the API. This means that
permissions can be shared from the public schema for all tenants.

Currently, seeding requires we seed data in all tenant schemas, so we'll keep
the current setting of tenant_id in the seeds, but remove them from the serializer.

Once we move to serving from the public schema, this will ensure we do not have
superfluous permission data, and we'll just serve permissions out of the public
schema.


wcmitchell (approved):
👍

astrozzc (approved):

catastrophe-brandon:
This looks fine to me; I don't see a need for any QE verification work here.

PR #500, opened 2021-06-04
Reverse `tox` command ordering

coderbydesign:
Instead of running specs first only to find out linting has failed in CI, we should
fail fast if linting doesn't pass, first.


astrozzc (approved):
I like it 😁

PR #503, opened 2021-06-08
Update schema sync command

wcmitchell:
## Link(s) to Jira
- N/A

Found a bug when running the sync in CI, this should resolve errors when copying policies referencing existing Foreign Keys

## Local Testing
This is a bit more difficult than before, as we now have code that ensure new objects are created in the public schema as well as the tenant schema. Either removing those copied objects from the public schema or checking out a branch that does not yet include the creation sync up seems the easiest way to generate objects that need to be synced, then running the command.

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/management/management/commands/sync_schemas.py
+++ b/rbac/management/management/commands/sync_schemas.py
@@ -127,6 +127,8 @@ class Command(BaseCommand):
             roles = list(policy.roles.all())
             new_roles = []
             with tenant_context(public_schema):
+                policy.roles.clear()
+                policy.group = None
                 self.clear_pk(policy)
                 try:
                     policy.save()
--- a/rbac/management/management/commands/sync_schemas.py
+++ b/rbac/management/management/commands/sync_schemas.py
@@ -156,4 +156,4 @@ class Command(BaseCommand):
                     self.copy_custom_groups_to_public(tenant)
                     self.copy_custom_policies_to_public(tenant)
         except Exception as e:
-            self.stderr(f"Failure: {str(e)}")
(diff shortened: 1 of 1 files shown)


rbac/management/management/commands/sync_schemas.py
```diff
@@ -73,6 +73,7 @@ def copy_custom_roles_to_public(self, tenant):
                     if not access.tenant:
                         access.tenant = tenant
                     access.role = role
+                    access.permission = Permission.objects.get(permission=access.permission.permission)
```
astrozzc, line 76: This makes me wonder the line 82 might get into trouble since it is querying within public schema. Where actually we would like to query within tenant schema.

And do we have to do similar thing as line 73-74 for other resources like group, policy, etc.?

wcmitchell: The Permissions should be seeded into the public schema though, right? So we'll want to use those rather than the ones in the given tenant schemas, correct?

And yes, each other object/model has a check on its tenant before we move to the public schema.

astrozzc (approved):
Looks promising

coderbydesign:
@wcmitchell 728877a0871327fe6a6098e7bd4e471cdd78d4c8 was added to address a few issues (notes in the commit message)

PR #506, opened 2021-06-09
[WIP] Testing smoke tests

BlakeHolifield:
Will close after validating


smoke_test.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# create a custom svc acct for the iqe pod to run with that has elevated permissions
+# SA=$(oc get -n $NAMESPACE sa iqe --ignore-not-found -o jsonpath='{.metadata.name}')
+# if [ -z "$SA" ]; then
+#     oc create -n $NAMESPACE sa iqe
+# fi
+# oc policy -n $NAMESPACE add-role-to-user edit system:serviceaccount:$NAMESPACE:iqe
+# oc secrets -n $NAMESPACE link iqe quay-cloudservices-pull --for=pull,mount
+oc apply -f $APP_ROOT/deploy/rbac-cji-smoketest.yml
+sleep 30
```
bsquizz, line 19: :raised_eyebrow: tell me more about this :P 

Are we waiting on the job to appear here or something, is there a more efficient way to wait here?

smoke_test.sh
```diff
@@ -0,0 +1,27 @@
...
+oc logs -n $NAMESPACE job/rbac-smoke-tests-iqe -f &
+oc wait --timeout=3m --for=condition=Complete job/rbac-smoke-tests-iqe || oc wait --timeout=3m --for=condition=Failed job/rbac-smoke-tests-iqe
```
bsquizz, line 21: Timeout is likely going to need to be a variable, I guess rbac's tests just happen to run in under 3 min.

unit_test.sh
```diff
@@ -1,15 +1,22 @@
...
 
+# check if NAMESPACE is set
+[ -z "$NAMESPACE" ] && exit 1
+oc get secret rbac -o json -n $NAMESPACE
+
+
 python3 -m venv app-venv
 . app-venv/bin/activate
 pip install --upgrade pip setuptools wheel pipenv tox psycopg2-binary
+pip install crc-bonfire==v1.4.0b0
```
bsquizz, line 16: You can back this change out now -- the changes in the beta are released

smoke_test.sh
BlakeHolifield: Yeah I'm working on figuring that out now. We need the job to exist, but we have to wait for clowder to form and create it first. The solution in the kuttl tests is `sleep 60` so I'm trying to find a dynamic way to get notified that the job is ready and running.

smoke_test.sh
BlakeHolifield: This is some random test value to see what happens. Right now these fail so 3m is just easier. We also have a Clowder PR to make sure we don't restart the iqe job 6 times and only run once like the v3 pipelines.

smoke_test.sh
bsquizz: Would be far more efficient to do in python I guess... but I think something along these lines could work in bash (need to make sure the script is using `#!/bin/bash` to get `$SECONDS` -- see https://stackoverflow.com/a/11198713)

```bash
#!/bin/bash
job_name=foo
found=0
end=$((SECONDS+30))

echo "Waiting for Job $job_name to appear"

while [ $SECONDS -lt $end ]; do
    if `oc get job $job_name >/dev/null 2>&1`; then
        found=1
        break
    fi
done

if ! found; then
    echo "Job $job_name failed to appear"
    exit 1
fi
```

bsquizz: Oops... forgot you'll want a `sleep 1` at the bottom of that while loop too ;)

catastrophe-brandon:
Based on the Jenkins output it looks like there's a syntactical error in the bash script atm:
```
if [$running == "false"] ; then
13:40:15     echo "Job $job_name failed to start"
13:40:15     exit 1
13:40:15 fi
13:40:15 ++ '[true' == 'false]'
13:40:15 /srv/volume/jenkins/workspace/RedHatInsights-insights-rbac-pr-check/smoke_test.sh: line 52: [true: command not found
13:40:15 

```

smoke_test.sh
```diff
@@ -0,0 +1,63 @@
...
+while [ $SECONDS -lt $end ]; do
+    pod=$(oc get pods -n $NAMESPACE -o json | jq -r '.items[] | select(.status.phase=="Running") | select(.metadata.name|test("rbac-smoke.")) .metadata.name')
+    if [ -z $pod ]; then
+        running=true
+        break
+    fi
+    sleep 1
+done
+
+if [$running == "false"] ; then
```
catastrophe-brandon, line 52: In this case I think you'd need to insert some space before `$running` and maybe wrap it in quotes for good measure. lmk if you want me to keep working in this branch and i'll tack on some commits.

BlakeHolifield: Thanks! My bash is so bad :man_facepalming:

catastrophe-brandon: Nah, it's fine. I've accidentally done this plenty of times :)

catastrophe-brandon:
It looks like the recent failure was an actual test failure due to missing config. Tests appeared to fail because Dynaconf chokes when it can't find users and throws an ugly traceback (gotta love our error handling or lack thereof in IQE).
```
==================================== ERRORS ====================================
_____________________ ERROR at setup of test_status_sanity _____________________

self = <Box: {'metadata': {'default_fields': {'component': 'rbac', 'automation_status': 'automated', 'negative': False, 'impo....redhat.com/wapps/ugc/protected/usermgt/userList.html', 'insights-qa': {'first_name': 'Insights', 'last_name': 'QA'}}}>
item = 'USERS', _ignore_default = True

    def __getitem__(self, item, _ignore_default=False):
        try:
>           value = super(Box, self).__getitem__(item)
E           KeyError: 'USERS'

lib64/python3.6/site-packages/box.py:424: KeyError

During handling of the above exception, another exception occurred:

self = <Box: {'metadata': {'default_fields': {'component': 'rbac', 'automation_status': 'automated', 'negative': False, 'impo....redhat.com/wapps/ugc/protected/usermgt/userList.html', 'insights-qa': {'first_name': 'Insights', 'last_name': 'QA'}}}>
item = 'USERS', args = (), kwargs = {'_ignore_default': True}, n_item = 'users'

    def __getitem__(self, item, *args, **kwargs):
        try:
>           return super(DynaBox, self).__getitem__(item, *args, **kwargs)

lib64/python3.6/site-packages/dynaconf/utils/boxing.py:21: 
```

Based on the traceback I'm guessing this is because we don't have `clowder_smoke` environment in our `settings.default.yaml`?

BlakeHolifield:
Yep! Once the `clowder_smoke` setup is done in the `setting.default.yaml`, rbac will get the correct configs.

pr_check.sh
```diff
@@ -1,18 +1,26 @@
...
 
-IQE_PLUGINS="rbac"  # name of the IQE plugin for this APP
-IQE_MARKER_EXPRESSION="rbac_smoke"  # This is the value passed to pytest -m
-IQE_FILTER_EXPRESSION=""  # This is the value passed to pytest -k
+export IQE_PLUGINS="rbac"  # name of the IQE plugin for this APP
+export IQE_MARKER_EXPRESSION="rbac_smoke"  # This is the value passed to pytest -m
+export IQE_FILTER_EXPRESSION=""  # This is the value passed to pytest -k
+export CJI_PATH="deploy/rbac-cji-smoketest.yml"
+
+find . -name cdappconfig.json
```
bsquizz, line 12: What's this used for? :)

smoke_test.sh
```diff
@@ -0,0 +1,90 @@
...
+    echo "Job $job_name failed to start"
+    exit 1
+fi
+
+# Pipe logs to background to keep them rolling in jenkins
+oc logs -n $NAMESPACE $pod -f &
+
+# Wait for the job to Complete or Fail before we try to grab artifacts
+# condition=complete does trigger when the job fails
+oc wait --timeout=3m --for=condition=Complete -n $NAMESPACE job/$job_name
```
bsquizz, line 61: Some apps' tests will definitely take longer than 3 min... should we make timeout configurable via some env var defined at the top of pr_check.sh and default to something kind of large like 30min to start with?

smoke_test.sh
```diff
@@ -0,0 +1,90 @@
...
+
+# Pipe logs to background to keep them rolling in jenkins
+oc logs -n $NAMESPACE $pod -f &
+
+# Wait for the job to Complete or Fail before we try to grab artifacts
+# condition=complete does trigger when the job fails
+oc wait --timeout=3m --for=condition=Complete -n $NAMESPACE job/$job_name 
+
+# Get the minio client
+curl https://dl.min.io/client/mc/release/linux-amd64/mc -o mc
```
bsquizz, line 64: Should we just go ahead and bake this part into the jenkins agent's ansible playbook?

And then here we could run
```bash
if ! command -v mc &> /dev/null
then
    echo "'mc' command not found on PATH"
    exit 1
fi
```

pr_check.sh
```diff
@@ -1,18 +1,26 @@
...
+export COMPONENT_NAME="rbac"  # name of app-sre "resourceTemplate" in deploy.yaml for this component
+export IMAGE="quay.io/cloudservices/rbac"  # the image location on quay
 
-IQE_PLUGINS="rbac"  # name of the IQE plugin for this APP
-IQE_MARKER_EXPRESSION="rbac_smoke"  # This is the value passed to pytest -m
-IQE_FILTER_EXPRESSION=""  # This is the value passed to pytest -k
+export IQE_PLUGINS="rbac"  # name of the IQE plugin for this APP
+export IQE_MARKER_EXPRESSION="rbac_smoke"  # This is the value passed to pytest -m
+export IQE_FILTER_EXPRESSION=""  # This is the value passed to pytest -k
+export CJI_PATH="deploy/rbac-cji-smoketest.yml"
```
bsquizz, line 10: Being pedantic here-- should we call this IQE_CJI_PATH ?

smoke_test.sh
```diff
@@ -0,0 +1,90 @@
...
+oc get secret env-$NAMESPACE-minio -o json -n $NAMESPACE | jq -r '.data' > minio-creds.json
+
+# Grab the needed creds from the secret
+export MINIO_ACCESS=$(jq -r .accessKey < minio-creds.json | base64 -d)
+export MINIO_SECRET_KEY=$(jq -r .secretKey < minio-creds.json | base64 -d)
+export MINIO_HOST=localhost
+export MINIO_PORT=$LOCAL_SVC_PORT
+
+# Setup the minio client to auth to the local eph minio in the ns
+./mc alias set minio http://$MINIO_HOST:$MINIO_PORT $MINIO_ACCESS $MINIO_SECRET_KEY
```
bsquizz, line 84: If get get mc pre-installed on PATH, remove the `./`

smoke_test.sh
```diff
@@ -0,0 +1,90 @@
...
+
+# "mirror" copies the entire artifacts dir from the pod and writes it to the jenkins node
+./mc mirror --overwrite minio/$pod-artifacts artifacts/
```
bsquizz, line 87: Same here

smoke_test.sh
```diff
@@ -0,0 +1,90 @@
...
+#NAMESPACE="mynamespace" -- namespace to deploy iqe pod into, can be set by 'deploy_ephemeral_env.sh'
+
+
+function kill_port_fwd {
+    echo "Caught signal, kill port forward"
+    if [ ! -z "$PORT_FORWARD_PID" ]; then kill $PORT_FORWARD_PID; fi
+}
+
+# The CJI var name will need to be exported in the main pr_check.sh
+oc apply -n $NAMESPACE -f $APP_ROOT/$CJI_PATH
```
bsquizz, line 15: Perhaps a bonfire command should be written that applies a CJI template for you. Actually, it could probably take care of all the 'wait for job to appear', 'wait for it to run', etc. etc. logic you have below as well (less bash looping :P). And then also if we make small tweaks to the spec of a CJI we won't have to tell 50 repos to update their CJI -- we could just update the template in bonfire

pr_check.sh
BlakeHolifield: Not sure. Seems like a debugging statement I inherited. Will remove

BlakeHolifield:
My work here is done. Closing this WIP in favor of  #516

PR #510, opened 2021-06-17
Set the default `MAX_SEED_THREADS` in Clowder template

coderbydesign:
Without this set, when we trigger seeds via Turnpike, it uses the default value for
`max_workers in `ThreadPoolExecutor` [1] which ends up killing our stage environment
because of resource issues.

[1] https://docs.python.org/3/library/concurrent.futures.html#threadpoolexecutor


wcmitchell (approved):
LGTM

astrozzc (approved):

PR #511, opened 2021-06-17
Clone custom default group into the public schema

coderbydesign:
https://issues.redhat.com/browse/RHCLOUD-14680

Currently when an org modifies their `Default access` group, we change the `system` flag to `False` so that the platform no longer manages it via seeds, and we update the name to `Custom default access`. 

Since the `Default access` and `Custom default access` groups in the public schema should make use of the system roles and not duplicate them (sharing the `Default access` group across all tenants in the public schema), we need to account for orgs modifying the creating a new custom default group by creating a new group, vs renaming it in the public schema.

Before we begin serving data out of the public schema, we'll need to have a condition to serve permissions out of the `Custom default access` group if one exists for the tenant, vs the `Default access` group shared by all tenants.

**Test Locally:**
- make reinitdb
- ./rbac/manage.py seeds
- make serve
- curl http://localhost:8000/api/rbac/v1/groups/
- find your default group and list the roles
- curl http://localhost:8000/api/rbac/v1/groups/<uuid>/roles/
- find a role and remove it from the default group
- curl http://localhost:8000/api/rbac/v1/groups/<uuid>/roles/?roles=<uuid> -DELETE
- confirm the role was removed from both schemas
- find a role and add it to the default group
- curl http://localhost:8000/api/rbac/v1/groups/<uuid>/roles/ -POST -H 'Content-Type: application/json' -d '{"roles": [<uuid>]}'
- confirm the role exists in both schemas

### DIFF
--- a/rbac/management/group/definer.py
+++ b/rbac/management/group/definer.py
@@ -19,6 +19,7 @@
 import logging
 
 from django.db import transaction
+from django.db.models import Q
 from django.db.models.query import QuerySet
 from django.utils.translation import gettext as _
 from management.group.model import Group
@@ -101,9 +102,6 @@ def add_roles(group, roles_or_role_ids, tenant, user=None, replace=False, duplic
 
     for tenant_schema in schema_handler(tenant, include_public=duplicate_in_public):
         group, created = Group.objects.get_or_create(name=group_name, tenant=tenant)
-        # TODO
-        # we can't rely on this (currently roles won't be added in the public schema)
-        # because the UUID is different
         system_policy_name = "System Policy for Group {}".format(group.uuid)
         system_policy, system_policy_created = Policy.objects.update_or_create(
             system=True, group=group, name=system_policy_name, defaults={"tenant": tenant}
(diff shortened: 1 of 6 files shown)


rbac/management/group/definer.py
```diff
@@ -63,12 +65,33 @@ def seed_group(tenant):
...
+    with tenant_context(public_tenant):
+        if Group.objects.filter(name=group.name, platform_default=group.platform_default, tenant=tenant):
+            return
+
+        public_default_roles = (
+            Group.objects.get(platform_default=True, tenant=public_tenant).policies.get(system=True).roles.all()
+        )
+
+        group.save()
+        tenant_default_policy.group = group
```
astrozzc, line 92: Overall LGTM, probably not a big deal but should we concern about the name of the policy. The name is sth like "Default policy for group xxxxxx", where the xxxxxx is the group uuid within the tenant, and changed after copying into public schema.

astrozzc (approved):

rbac/management/group/definer.py
astrozzc: Chatted with Keith, this won't be a problem since the UUID is not changed after clear PK

wcmitchell (approved):
Came at this from every angle I could and couldn't break it. LGTM!

PR #454, opened 2021-03-31
Add DB lock during tenant creation

wcmitchell:
Add explicit DB locks for tenant creation

Signed-off-by: Chris Mitchell <cmitchel@redhat.com>

## Link(s) to Jira
- 

## Description of Intent of Change(s)
Two requests occurring close enough together in this codebase can cause DB deadlocks due to how our migrations are set up. This PR will set up a table lock during creation to prevent that occurrence

## Local Testing
1. Pull down branch
2. make reinitdb
3. make serve
4. In another shell run `curl localhost:8000/api/rbac/v1/roles/ & curl localhost:8000/api/rbac/v1/roles/` to cause a double access
5. Ensure that no 500s are returned and no deadlocks reported

## Checklist
- [ ] if API spec changes are required, is the spec updated?
- [ ] are there any pre/post merge actions required? if so, document here.
- [ ] are theses changes covered by unit tests?
- [ ] if warranted, are documentation changes accounted for?
- [ ] does this require migration changes?
  - [ ] if yes, are they backwards compatible?
- [ ] is there known, direct impact to dependent teams/components?
  -  if yes, how will this be handled?

## Secure Coding Practices Checklist Link
- https://github.com/RedHatInsights/secure-coding-checklist

## Secure Coding Practices Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [ ] General Coding Practices

PR #518, opened 2021-07-21
Revert "Add DB lock during tenant creation"

coderbydesign:
Reverts RedHatInsights/insights-rbac#454

In an effort to resolve the DB lock/latency issues we've been seeing in RBAC prod, we may need to roll this back to see whether or not this is contributing to the issue.

We understand this will add an edge-case bug back in, where during schema creation if multiple requests for the tenant come in in parallel, the last request will win, and those before will fail due to a deadlock. One potential resolution may be to add a status column to the Tenant table which we can flag as pending while the schema is being created and migrations/seeds are run. During that time, subsequent requests could either return a 400 to retry, or we could wait/poll the requests until the status is clear.

### DIFF
--- a/rbac/rbac/middleware.py
+++ b/rbac/rbac/middleware.py
@@ -78,16 +78,11 @@ class IdentityHeaderMiddleware(BaseTenantMiddleware):
                     raise Http404()
             else:
                 with transaction.atomic():
-                    try:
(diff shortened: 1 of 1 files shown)


astrozzc (approved):
LGTM, concurrent requests of new accounts are supposed to get some errors during tenant seeding but that's better than bringing service down

PR #519, opened 2021-07-21
Wait for schema to be ready before serving requests

coderbydesign:
This does serveral things:
- defer the creation of the schema, so it doesn't happen as a create hook
- add a `ready` boolean field to `api.tenant`
- if the tenant is created, run migrations explicitly and then seeds
- then set the tenant to `ready=True`
- concurrent requests will get the tenant object, but will poll the db until the schema
  is ready before continuing

## Local Testing
Send concurrent requests to a new account, and all should succeed.


astrozzc (approved):
LGTM

wcmitchell (approved):
👍

PR #577, opened 2021-11-10
Account for schemas which haven't been seeded in "unmodified" logic

coderbydesign:
In the case where a schema has not yet been seeded, the check around:

```
Group.objects.filter(system=True).count() != 1
```

will cause the method to return `False`, since there will be no groups at all.
Instead, we should just do the same check we do with roles, and compare the number
of system groups to total groups.


wcmitchell (approved):
👍

astrozzc (approved):
LGTM！

PR #579, opened 2021-11-17
Account for tenants with no schema during pruning

coderbydesign:
In the event an RBAC service pod restarts during the creation of a schema, there is
an edge case exposed where a `Tenant` object will be created, but not a schema.

We cannot have both the `Tenant` and schema creation in a transaction currently,
because of a deadlock issue which this PR helped to resolve: https://github.com/RedHatInsights/insights-rbac/pull/519

A more robust fix would be to know when a schema has not been created for a given
tenant, and remove the `Tenant` object. This is a stop-gap until we move off
of schema-based tenancy in the near future, to allow us to manually remove these
via Turnpike.

It ensures that in the check to see if a tenant has been modified, we confirm
the schema exists. We need to do this because if we don't, and it doesn't exist,
the `search_path` on the query will fall back to using the public schema, in
which case there will be custom groups/roles, and we won't be able to propertly
prune the tenant which has been created without a valid schema.

### DIFF
--- a/rbac/internal/views.py
+++ b/rbac/internal/views.py
@@ -34,7 +34,7 @@ from management.tasks import (
     run_seeds_in_worker,
     run_sync_schemas_in_worker,
 )
-from tenant_schemas.utils import tenant_context
+from tenant_schemas.utils import schema_exists, tenant_context
 
 from api.models import Tenant
 from api.tasks import cross_account_cleanup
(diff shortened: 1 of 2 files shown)


rbac/internal/views.py
```diff
@@ -50,16 +50,22 @@ def destructive_ok():
     return now < settings.INTERNAL_DESTRUCTIVE_API_OK_UNTIL
 
 
-def tenant_is_modified():
+def tenant_is_modified(schema_name):
     """Determine whether or not the tenant is modified."""
-    return (Role.objects.filter(system=True).count() != Role.objects.count()) or (
+    has_custom_data = (Role.objects.filter(system=True).count() != Role.objects.count()) or (
```
astrozzc, line 55: LGTM overall, wonder if we could put schema_exists(schema_name) before this line, so if it is false, we don't have to do these queries.

astrozzc (approved):

rbac/internal/views.py
coderbydesign: We can't put it before this line in this method, because the presence of a schema alone isn't enough to tell us if the tenant has been modified. I'd initially considered placing it in `tenant_is_unmodified` as:
```
def tenant_is_unmodified(schema_name):
    """Determine whether or not the tenant is unmodified."""
    if not schema_exists(schema_name):
        return True
    return not tenant_is_modified(schema_name)
```
But that would have only fixed the `tenant_is_unmodified` method, and seemed a bit more confusing. What are your thoughts?

wcmitchell (approved):
👍

rbac/internal/views.py
coderbydesign: Pushed an update in 7aed3a4

PR #589, opened 2022-01-13
Serving from public schema

coderbydesign:

## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-11199

Migration to serve from the public schema


rbac/rbac/middleware.py
```diff
@@ -230,6 +230,9 @@ def process_request(self, request):  # pylint: disable=R1710
 
             super().process_request(request)
             # We are now in the database context of the tenant
+            if settings.SERVE_FROM_PUBLIC_SCHEMA:
```
coderbydesign, line 233: Initially I was thinking we could just return the public schema/tenant from `get_tenant` after we create (when needed) the tenant schema https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/rbac/middleware.py#L106 since that would change the search path for us, but it looks like that wouldn't set the `request.tenant` correctly, it would be the public tenant.

Your approach keeps `get_tenant` the actual tenant schema, and then explicitly sets _just_ the search path, so we still are able to filter on the data for the tenant? I think maybe that's what you were referring to yesterday with the other approach returning everything in the public schema?

rbac/rbac/middleware.py
```diff
@@ -230,6 +230,9 @@ def process_request(self, request):  # pylint: disable=R1710
...
+                connection.set_schema_to_public()
```
coderbydesign, line 234: Can/should we consolidate these two calls so they're the same? https://github.com/RedHatInsights/insights-rbac/pull/589/files#diff-642029e74ba2ca49cb6da38bfbfae49d94851a96e507abf8a9c5cef1dd000ec3R78

PR #592, opened 2022-01-18
Check rbac on rbac for role and principle query access

coderbydesign:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-17081

Currently there are a few checks where we only look for `request.admin`, not whether
or not the request user has the correct RBAC-on-RBAC permissions.

The views use the permissions classes by default, but anywhere else in the code
we check for `request.admin`, we should probably be checking for the correct permission
for the model.

In this case, it affects `/api/rbac/v1/roles/?username=foo` where the username
is not the same as the request user, which currently requires you to be an admin.
This changes that to delegate the check to our permissions classes, which will
check for an org admin _or_ the correct permissions.

Note that we need to update the util method for principals as well, because we're looking
up the principal from the query param in this case, which requires principal permissions.

## Local Testing
- set your local request user to a non-admin (user_1)
- ensure you have another user with roles which we can query for (user_2)
- with user_1 a non-admin, without any RBAC-on-RBAC permissions, the following should fail: `/api/rbac/v1/roles/?username=user_2`
- grant user_1 `rbac:*:*` permissions
- now with user_1 a non-admin, with the RBAC admin permissions, the following should succeed: `/api/rbac/v1/roles/?username=user_2` and you should see user_2's roles

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


rbac/management/utils.py
```diff
@@ -50,11 +51,11 @@ def get_principal_from_request(request):
     """Obtain principal from the request object."""
     current_user = request.user.username
     qs_user = request.query_params.get(USERNAME_KEY)
+    principal_permission = PrincipalAccessPermission()
```
astrozzc, line 54: Should we make the class (RoleAccessPermission, PrincipalAccessPermission) into singleton or make that method static so we don't have to initiate it each time?

coderbydesign: @astrozzc it doesn't look like we can use DRF to do that out of the box, since it's a view-level method which the view takes an instance of: https://www.django-rest-framework.org/api-guide/permissions/#custom-permissions

But I do agree that it's not ideal to have this instantiated on each request. We could consider adding an attribute to the module, to allow for something like:

```python
PRICIPAL_PERMISSION_INSTANCE = PrincipalAccessPermission()
...
...
def get_principal_from_request(request):
    ...
    ...
    if qs_user and not PRICIPAL_PERMISSION_INSTANCE.has_permission(request=request, view=None):
```

astrozzc (approved):
LGTM

PR #598, opened 2022-01-25
Add internal endpoint for initializing tenants

coderbydesign:
This will add `POST /_private/api/tenant/<schema-name>/init/` which will be an
idempotent way to do the following for a tenant:

- create a schema and run migrations if one does not exists for the tenant
- run seeds after the schema is configured
- flip `ready=True`

This will allow us to manage `Tenant` objects which are in a "hung" state, by
running the schema creation and migrations/seeds if the schema does not exist.
If it does exist, it will skip the schema creation and migration, and just run
seeds, which will not be an issue.

Tenants are able to get into this state when the service restarts during the tenant
creation process. We're not able to run the tenant object creation, and schema
setup in a transaction because of an edge-case with concurrent requests during
creation. This means we have the potential to orphan Tenant records, leaving them
in a state where `ready=False`, and subsequent requests will hang.

We'll also check to see that the schema has been created, and if it has, run the seeds. 
Otherwise this means the schema is setup, and we just need to flip the flag.

Context: https://github.com/bernardopires/django-tenant-schemas/blob/7f72d7af039e6cf2a15520d525636d6e11e72efc/tenant_schemas/models.py#L108-L109

`_private/api/tenant/?ready=true|false`
We're also adding an endpoint to collect the `ready=false` tenants. This will allow 
us to collect the schema names for any tenants in a "bad", not ready state, 
to run against the "init" endpoint.

## Local Testing
- create a tenant that's `ready=false`, without schema (just manually in the DB to simulate an orphaned row)
- update dev middleware identity to use that account
- confirm it hangs when you try to make a request (this is the edge-case where it gets stuck in the loop in middleware)
- update dev middleware to be an `associate` so you can run internal requests
- hit `/_private/api/tenant/` and confirm your newly created tenant is not ready
- POST to `/_private/api/tenant/<schema-name>/init/` with your schema name and confirm it runs migrations/seeds and resolves the issue
- POST to `/_private/api/tenant/<schema-name>/init/` with an existing schema name and confirm it does nothing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/internal/views.py
+++ b/rbac/internal/views.py
@@ -27,14 +27,13 @@ from django.db.migrations.recorder import MigrationRecorder
 from django.http import HttpResponse
 from django.shortcuts import get_object_or_404
 from management.cache import TenantCache
-from management.group.definer import seed_group
 from management.models import Group, Role
-from management.role.definer import seed_permissions, seed_roles
 from management.tasks import (
     run_migrations_in_worker,
     run_reconcile_tenant_relations_in_worker,
     run_seeds_in_worker,
     run_sync_schemas_in_worker,
+    run_init_tenant_in_worker,
 )
 from tenant_schemas.utils import schema_exists, tenant_context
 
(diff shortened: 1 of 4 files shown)


rbac/internal/views.py
```diff
@@ -95,6 +97,37 @@ def list_unmodified_tenants(request):
     return HttpResponse(json.dumps(payload), content_type="application/json")
 
 
+def list_tenants(request):
+    """List tenant details.
+
+    GET /_private/api/tenant/?ready=<true|false>&limit=<limit>&offset=<offset>
+    """
+    limit = int(request.GET.get("limit", 0))
```
astrozzc, line 105: wonder if we should use a positive number as default

astrozzc (approved):
Nice, LGTM!

rbac/internal/views.py
coderbydesign: If we use `0` it will default to not being set, which is what we want, since `if limit` won't pass.

astrozzc: oh, got it 👍

wcmitchell (approved):

PR #611, opened 2022-02-07
Fix the access cache key when serving from public

coderbydesign:
Since we always set the access policy cache key based on the `request.tenant.schema_name` [1]
which is what we want so that it's specific to each account, we need to ensure that
when we invalidate the cache, we're using the correct key.

These changes will ensure we key off the schema name from the object in the signal,
vs the case where the `public` schema is being used and the current code will
cause a miss on the key lookup.

[1] https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/access/view.py#L108

### DIFF
--- a/rbac/management/group/model.py
+++ b/rbac/management/group/model.py
@@ -20,7 +20,7 @@ import logging
 from uuid import uuid4
 
 from django.conf import settings
-from django.db import connections, models
+from django.db import models
 from django.db.models import signals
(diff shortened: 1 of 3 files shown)


wcmitchell (approved):
Local testing seems good, LGTM

astrozzc (approved):
Tested locally, works great!

PR #613, opened 2022-02-16
Return `id` along with `schema_name` for internal Tenant endpoint

coderbydesign:
In the case where we need to compare tenants which are not "ready" with RDS
queries, the queries only have the PK id. This will help us correlate those
repeated queries in RDS with the schema names.


astrozzc (approved):
LGTM

PR #656, opened 2022-02-23
Add ability to run seeds by schema

coderbydesign:
This allows seeds to be run for a specific schema/set of schemas. This will help
us to run seeds for the public schema only, as we transition away from tenant
schemas, similar to the same ability for migrations.

This will allow for:
```
curl https://internal.console.redhat.com/api/rbac/seeds/run/ -H 'Content-Type: application/json' -d '{"schemas": ["public"]}'
```


wcmitchell (approved):
LGTM

astrozzc (approved):
LGTM

PR #660, opened 2022-03-04
Changes pulling out django_tenant_schemas plugin

wcmitchell:
Unit tests passing locally, manual testing exercises yet to be done, but wanted to get a working push up and under the other devs' eyes. It's a big mess, touches nearly everything, but hopefully working!

Signed-off-by: Chris Mitchell <cmitchel@redhat.com>

## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-18022
- https://issues.redhat.com/browse/RHCLOUD-18026

## Description of Intent of Change(s)
Now that we've collapsed everything into the public schema, we should be able to clean up the code and remove references to django_tenant_schemas

## Local Testing
How can the feature be exercised? 
How can the bug be exploited and fix confirmed?
Is any special local setup required?

## Checklist
- [ ] if API spec changes are required, is the spec updated?
- [ x ] are there any pre/post merge actions required? if so, document here.
  - Two new migrations to rename the 'schema_name' field to 'tenant_name'
- [ x ] are theses changes covered by unit tests?
- [ ] if warranted, are documentation changes accounted for?
- [ x ] does this require migration changes?
  - [ ] if yes, are they backwards compatible?
- [ ] is there known, direct impact to dependent teams/components?
  -  if yes, how will this be handled?

## Secure Coding Practices Checklist Link
- https://github.com/RedHatInsights/secure-coding-checklist

## Secure Coding Practices Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [ ] General Coding Practices

PR #661, opened 2022-03-10
Always run init container

coderbydesign:
This will ensure that now we always run the init container when deploying with
Clowder, regardless of environment. The init container will run seeds and migrations.
Currently we still need to explicitly target the public schema, but once DTS
is removed and the updates from #660 are in, we will be able to update our commands
to just run Django migrations with the default command, and seeds will not run
against all schemas.

- updates the Makefile to only run migrations again public (local only)
- updates README for consistency and to avoid confusion
- in the init container script, no longer check for ephemeral env, and run seeds/migrationss
  in the context of the public schema
- remove `EPH_ENV` from the template
- update the OpenShift run script (even though we're not currently using this)
- update the command args to accept a list of `schema_list`

We'll still want to:
- remove the `EPH_ENV` params from app-interface deployments

## Local Testing
You can test by running the updated commands locally:
```
./rbac/manage.py seeds --schema_list=public
./rbac/manage.py migrate_schemas --schema=public
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,7 +99,7 @@ make-migrations:
 	DJANGO_READ_DOT_ENV_FILE=True $(PYTHON) $(PYDIR)/manage.py makemigrations api management
 
 run-migrations:
-	DJANGO_READ_DOT_ENV_FILE=True $(PYTHON) $(PYDIR)/manage.py migrate_schemas --executor=parallel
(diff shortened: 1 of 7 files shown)


wcmitchell (approved):
Looks good to me! Should unblock other work and move us another step towards cleaning up DTS!

astrozzc (approved):
LGTM

PR #666, opened 2022-03-18
Admin default access

mjholder:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-16784

The purpose of these changes is to have admin default access be returned on the access endpoint, exclusively for org admins. Non-org admins and users given RBAC admin roles will not be able to see admin default access.

## Local Testing
This passed local tox tests and I created new unit tests for the new feature.

## Checklist
- [X] are theses changes covered by unit tests?

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [X] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [X] General Coding Practices

### DIFF
--- a/rbac/management/querysets.py
+++ b/rbac/management/querysets.py
@@ -15,13 +15,11 @@
 #    along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 """Queryset helpers for management module."""
-import binascii
+
 import logging
-from json.decoder import JSONDecodeError
 
 from django.conf import settings
 from django.db.models.aggregates import Count
-from django.http import HttpResponseForbidden
 from django.urls import reverse
 from django.utils.translation import gettext as _
 from management.group.model import Group
@@ -41,9 +39,7 @@ from management.utils import (
 )
 from rest_framework import permissions, serializers
 
-from api.common import RH_IDENTITY_HEADER
 from api.models import Tenant
-from api.serializers import extract_header
 from rbac.env import ENVIRONMENT
 
 
@@ -58,17 +54,6 @@ PRINCIPAL_QUERYSET_MAP = {
     Role.__name__: roles_for_principal,
 }
 logger = logging.getLogger(__name__)
-
-
-def _get_org_admin(request):
-    try:
-        _obj, json_rh_auth = extract_header(request, RH_IDENTITY_HEADER)
-    except (JSONDecodeError, binascii.Error):
-        logger.exception("Invalid X-RH-Identity header.")
-        return HttpResponseForbidden()
-    is_org_admin = json_rh_auth["identity"]["user"]["is_org_admin"]
-
-    return is_org_admin
 
 
 def get_annotated_groups():
@@ -218,7 +203,8 @@ def get_access_queryset(request):
         raise serializers.ValidationError({key: _(message)})
 
     app = request.query_params.get(APPLICATION_KEY)
-    is_org_admin = _get_org_admin(request)
+    is_org_admin = request.user.admin
+
     return get_object_principal_queryset(
         request,
         PRINCIPAL_SCOPE,
--- a/rbac/management/querysets.py
+++ b/rbac/management/querysets.py
@@ -15,8 +15,13 @@
 #    along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 """Queryset helpers for management module."""
+import binascii
+import logging
+from json.decoder import JSONDecodeError
+
 from django.conf import settings
 from django.db.models.aggregates import Count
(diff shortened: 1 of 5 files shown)


rbac/management/querysets.py
```diff
@@ -206,6 +227,7 @@ def get_access_queryset(request):
             APPLICATION_KEY: app,
             "prefetch_lookups_for_ids": "resourceDefinitions",
             "prefetch_lookups_for_groups": "policies__roles__access",
+            "is_org_admin": is_org_admin,
```
coderbydesign, line 230: I _think_ we should be able to simplify this and just use `request.user.admin` without the need for any of the other changes in this file, since we add the user (and the admin flag from the identity header) onto the request object in the middleware: 

https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/rbac/middleware.py#L176
https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/rbac/middleware.py#L215

rbac/management/utils.py
```diff
@@ -126,6 +126,10 @@ def groups_for_principal(principal, tenant, **kwargs):
         assigned_group_set = assigned_group_set.prefetch_related(prefetch_lookups)
         platform_default_group_set = platform_default_group_set.prefetch_related(prefetch_lookups)
 
+    if kwargs.get("is_org_admin"):
+        admin_default_group_set = Group.objects.filter(admin_default=True)
```
coderbydesign, line 130: It's probably not totally necessary, but for consistency maybe we could add a `Group` model helper similar to `Group.platform_default_set()` here, in the event we need to reuse that filter? Also, just for explicitness, we'll probably want to add a filter to query by the public tenant only. Right now we should only ever have one default admin group, but if we change that in the future to behave like the (custom) default access group(s), it would be helpful to have this as a baseline.

tests/management/test_querysets.py
```diff
@@ -351,6 +361,47 @@ def test_get_policy_queryset_bad_scope(self):
...
+        self._setup_group_for_org_admin_tests()
+
+        user = Mock(spec=User, account="00001", username="test_user")
+        req = Mock(user=user, method="GET", tenant=self.tenant, query_params={APPLICATION_KEY: "app"})
+        req.META = encoded_req.META
+
+        queryset = get_access_queryset(req)
+        self.assertEquals(queryset.count(), 1)
+
+    def test_get_access_queryset_non_org_admin(self):
```
coderbydesign, line 380: Can we also add an explicit test case for a non admin who has RBAC admin access via `rbac:*:*` for this AC?

> As a **non** org-admin who has RBAC admin access via the "User Access administrator" role, when I list my permissions on /api/rbac/v1/access/?application= I should **not** see the access from any "admin_default" roles listed.

coderbydesign: I tested manually locally and it works as expected, but would be good to have unit test coverage on that case specifically. 😄

coderbydesign (commented):
Looks good, @mjholder! Had a couple of small comments, but overall functionally this is solid. 🎉

tests/management/test_querysets.py
mjholder: I am testing it now, just to be clear if I set the permission to `rbac:*:*` that is enough to consider it an RBAC admin? It's working as expected but just want to make sure I have the test setup correctly.

rbac/management/utils.py
mjholder: Would we want the filtering of the schema to be in the helper method?

tests/management/test_querysets.py
coderbydesign: Yes, that's correct. The funky thing is that the `rbac:*:*` permission is in the admin default group locally, so you'll still see it in your permissions regardless. What I did was changed the admin default role to have a different permission and tested with that.

rbac/management/utils.py
coderbydesign: I think where you've got it makes sense, and follows the pattern we have.

coderbydesign (approved):
🌴

rbac/management/querysets.py
```diff
@@ -49,6 +53,7 @@
     Policy.__name__: policies_for_principal,
     Role.__name__: roles_for_principal,
 }
+logger = logging.getLogger(__name__)
```
astrozzc, line 56: Looks this is not used anywhere?

astrozzc (approved):
LGTM!

PR #671, opened 2022-03-25
Clean principals based on account

coderbydesign:
Currently in the cleaner we don't filter the BoP call by account, which means if
a principal changes accounts for instance, we'll find the record(s) and not clean
them. This ensures that the calls to the BoP are filtering by account.

Note that this will need to change with https://issues.redhat.com/browse/RHCLOUD-18474


astrozzc (approved):
LGTM!

PR #672, opened 2022-03-25
Admin default edit

mjholder:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-16786

This is to prevent changes to admin_default groups in roles via api commands.

## Local Testing
Passed unit testing, with new unit tests for this feature.

## Checklist
- [X] are theses changes covered by unit tests?

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [X] General Coding Practices

### DIFF
--- a/rbac/management/group/view.py
+++ b/rbac/management/group/view.py
@@ -147,11 +147,16 @@ class GroupViewSet(
         return GroupSerializer
 
     def protect_default_groups(self, action):
-        """Deny modifications on platform_default groups."""
+        """Deny modifications on platform_default and admin_default groups."""
         group = self.get_object()
         if group.platform_default:
             key = "group"
             message = "{} cannot be performed on platform default groups.".format(action.upper())
+            error = {key: [_(message)]}
+            raise serializers.ValidationError(error)
+        if group.admin_default:
+            key = "group"
+            message = "{} cannot be performed on admin default groups.".format(action.upper())
             error = {key: [_(message)]}
             raise serializers.ValidationError(error)
 
--- a/rbac/management/group/view.py
+++ b/rbac/management/group/view.py
@@ -149,14 +149,9 @@ class GroupViewSet(
     def protect_default_groups(self, action):
         """Deny modifications on platform_default and admin_default groups."""
         group = self.get_object()
-        if group.platform_default:
+        if group.platform_default or group.admin_default:
             key = "group"
-            message = "{} cannot be performed on platform default groups.".format(action.upper())
-            error = {key: [_(message)]}
-            raise serializers.ValidationError(error)
-        if group.admin_default:
-            key = "group"
-            message = "{} cannot be performed on admin default groups.".format(action.upper())
+            message = "{} cannot be performed on default groups.".format(action.upper())
             error = {key: [_(message)]}
             raise serializers.ValidationError(error)
 
@@ -433,6 +428,7 @@ class GroupViewSet(
         group = self.get_object()
         account = self.request.user.account
         if request.method == "POST":
+            self.protect_default_groups("add principals")
             if not request.user.admin:
                 for role in group.roles_with_access():
                     for access in role.access.all():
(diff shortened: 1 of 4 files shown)


rbac/management/group/view.py
```diff
@@ -147,13 +147,18 @@ def get_serializer_class(self):
...
     def protect_default_groups(self, action):
-        """Deny modifications on platform_default groups."""
+        """Deny modifications on platform_default and admin_default groups."""
         group = self.get_object()
         if group.platform_default:
             key = "group"
             message = "{} cannot be performed on platform default groups.".format(action.upper())
             error = {key: [_(message)]}
             raise serializers.ValidationError(error)
+        if group.admin_default:
```
coderbydesign, line 157: Sort of a nit, but I wonder if we could reuse the check above on 152, and just have it look for platform _or_ admin default, and dynamically set the failure message (or generalize it), to keep it one condition?

rbac/management/role/serializer.py
```diff
@@ -332,6 +332,11 @@ def validate_role_update(instance, validated_data):
         message = "System roles may not be updated."
         error = {key: [_(message)]}
         raise serializers.ValidationError(error)
+    if instance.admin_default:
```
coderbydesign, line 335: All `admin_default` roles should be `system` roles by default [via seeding](https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/role/definer.py#L39), correct? We may not even need this check.

rbac/management/group/view.py
```diff
@@ -147,13 +147,18 @@ def get_serializer_class(self):
         return GroupSerializer
 
     def protect_default_groups(self, action):
```
coderbydesign, line 149: Will this protect against principals being added to the group as well?

wcmitchell (approved):
Looks good here. Keith's notes are reasonable, and worth doing I think, but functionality looks good!

rbac/management/group/view.py
mjholder: No it doesn't. Should it?

coderbydesign: Yeah, that's called out in the acceptance criteria:

> As any API user, I should not be able to modify any "admin_default" groups, and should receive a validation error (not updating the group or adding/removing principals).

Of note: we _should_ have this in place for the `platform_default` group as well with regard to disallowing principal association, since these two groups have implicit access only, but we don't currently do that. Might be worth logging something to also have this check on `platform_default`.

rbac/management/role/serializer.py
mjholder: I see where you are coming from. I guess I tend to lean towards explicitly checking for conditions, but if you think it would be redundant I can remove that check. Is there any likely hood of there being an `admin_default=true` but `system=false` role being created? Since these admin_default groups would be coming from the RBAC config repo, we would expect it to be `system=true`, is that the idea of why we don't need the check?

rbac/management/group/view.py
coderbydesign: Sounds perfect!

rbac/management/group/view.py
mjholder: Ok I generalized the message to say the action 'cannot be performed on default groups'.

rbac/management/role/serializer.py
coderbydesign: Right, you can't supply the `admin_default` value (default false) or the `system` value (default false) via the API, so custom roles will never be `system=false` with `admin_default=true`. The only case this would happen is in the case where the business wants us to change to allow for "Custom Default Admin Access", similar to `platform_default`, in which case we'd flip the `system` flag to false when we update the group. But in that case, we'd still want the check as is, since you _should_ be able to update the group if it's no longer a `system` group. You make a good point about being explicit, though. In this case though it's really any `system` object which we want to protect.

mjholder: Ok that's a lot of good information and really clears it up for me. Thanks!

coderbydesign (approved):
Works great locally!

astrozzc (approved):

PR #683, opened 2022-04-11
EBS to Org ID feature flag in RBAC

mjholder:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-18471
- https://issues.redhat.com/browse/RHCLOUD-18474
- https://issues.redhat.com/browse/RHCLOUD-18473
- https://issues.redhat.com/browse/RHCLOUD-18661

This handles the implementation of the feature flag AUTHENTICATE_WITH_ORG_ID, which is used to flip between keying off EBS account id and org id. The two main feature affected are tenancy and CARs. This does not support anemic tenants, tenants **without** an ebs account. That support will come in a later PR.

Note: The kafka notifications change, from PR [678](https://github.com/RedHatInsights/insights-rbac/pull/678), isn't fully transitioned to flip the key as well. Another PR will be created to fully integrate notifications and org id.

## Local Testing
Unit tests have been updated to support the new org_id's and the flipping between which field we use as a key. It has been tested to work with the feature flag both on and off.

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


rbac/api/cross_access/view.py
```diff
@@ -215,17 +256,30 @@ def validate_and_format_input(self, request_data):
...
-        target_account = request_data.get("target_account")
-        if target_account == self.request.user.account:
-            self.throw_validation_error(
-                "cross-account-request", "Creating a cross access request for your own account is not allowed."
-            )
+        if settings.AUTHENTICATE_WITH_ORG_ID:
+            target_org = request_data.get("target_org")
+            if target_org == self.request.user.org_id:
+                self.throw_validation_error(
+                    "cross-account-request", "Creating a cross access request for your own account is not allowed."
```
coderbydesign, line 263: We should update this to reflect org.
```suggestion
                    "cross-account-request", "Creating a cross access request for your own org is not allowed."
```

rbac/api/cross_access/access_control.py
```diff
@@ -22,6 +22,7 @@
 QUERY_BY_KEY = "query_by"
 USER_ID = "user_id"
 ACCOUNT = "target_account"
+ORG_ID = "target_org"
```
coderbydesign, line 25: We'll also need to update the check below conditionally based on the flag being set, to require admin access when `query_by == ORG_ID`

rbac/api/cross_access/view.py
```diff
@@ -47,6 +53,11 @@
 class CrossAccountRequestFilter(filters.FilterSet):
     """Filter for cross account request."""
 
+    def org_id_filter(self, queryset, field, values):
```
coderbydesign, line 56: I know we're not using it on the current `account_filter` method, but I think we could make use of this shared filter here: https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/filters.py#L38

mjholder: Do you mean how the group objects UUID filter is setup? https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/group/view.py#L72

coderbydesign: Yeah, similar to that.

deploy/rbac-clowdapp.yml
```diff
@@ -2019,3 +2019,6 @@ parameters:
 - description: Enable sending out notification events of Red Hat changes
   name: NOTIFICATIONS_RH_ENABLED
   value: 'False'
+- description: Enable RBAC tenancy using Org ID
+  name: AUTHENTICATE_WITH_ORG_ID
```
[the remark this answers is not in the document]
mjholder: Ok, I added the variable to the `env` section.

rbac/rbac/settings.py
```diff
@@ -374,6 +374,10 @@
 except ValueError as e:
     INTERNAL_DESTRUCTIVE_API_OK_UNTIL = datetime.datetime(1970, 1, 1, tzinfo=pytz.UTC)
 
+# Prep for public schema changeover
+SERVE_FROM_PUBLIC_SCHEMA = ENVIRONMENT.bool("SERVE_FROM_PUBLIC_SCHEMA", default=False)
```
coderbydesign, line 378: This looks like something we don't need?

coderbydesign (approved):
Had one last minor comment, but otherwise this looks good! Tested locally, both with unit tests with the flag enabled/disabled, as well as manual poking with the flag in both positions. 🎉

rbac/api/cross_access/util.py
```diff
@@ -44,12 +45,16 @@ def check_cross_request_expiry():
     logger.info("Completed clean up of %d cross-account requests, %d expired.", len(cars), len(expired_cars))
 
 
-def create_cross_principal(target_account, user_id):
+def create_cross_principal(target_account, user_id, target_org=None):
```
astrozzc, line 48: Should we update here target_account=None? So when the flag flipped the account is not required

tox.ini
```diff
@@ -18,14 +18,15 @@ application-import-names = rbac, api
 [testenv]
 ;passenv = CI TRAVIS TRAVIS_*
 setenv =
-  DATABASE_NAME={env:DATABASE_NAME:rbac_test}
```
astrozzc, line 21: I think it might be better to keep the original one.
We have some data setup locally for testing in the database postgres.
The rbac_test would be destroyed. If we change it to postgres, I am afraid it would add some data we don't want to our current db postgres and destroy the db postgres after test finished.

astrozzc (commented):
Overall LGTM! Just some questions to clarify

rbac/rbac/middleware.py
```diff
@@ -76,20 +76,36 @@ class IdentityHeaderMiddleware(MiddlewareMixin):
...
+            tenant_name = create_tenant_name(request.user.account)
+            tenant = TENANTS.get_tenant(request.user.org_id)
+            if tenant is None:
+                if request.user.system:
+                    try:
+                        tenant = Tenant.objects.get(org_id=request.user.org_id)
+                    except Tenant.DoesNotExist:
+                        raise Http404()
+                else:
+                    tenant, created = Tenant.objects.get_or_create(
```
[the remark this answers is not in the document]
mjholder: I moved the org_id out of the defaults section. I still have tenant_name being populated so the information is there if we flip the feature flag.

tox.ini
mjholder: Good to know! I reverted the database name back to rbac_test.

rbac/api/cross_access/util.py
mjholder: Good catch. I updated the header to make target_account default to None.

astrozzc (approved):
LGTM!

wcmitchell (approved):

PR #684, opened 2022-04-11
Install `uvicorn[standard]`

coderbydesign:
Based on https://github.com/encode/uvicorn/issues/797 we may need this to behave
correctly. Locally things work fine, but in stage we're seeing 400s.


wcmitchell (approved):
:+1: Worth a shot!

PR #688, opened 2022-04-26
Upgrade to python 3.8

coderbydesign:
This works to upgrade RBAC to Python 3.8.x since 3.6 has reached EOL [1] and we've
started experiencing issues, specifically with `pipenv` and `pyenv` around breaking
changes with this unsupported version.

There's an issue (at least on macs) with `psycopg` being installed on 3.8, which
is why it's explicitly installed in the `tox.ini`. Before running `pipenv install`
in your virtual environment, explicitly running:
```
pipenv install psycopg==2.8.6
pipenv install psycopg-binary==2.8.6
pipenv install
```
seems to resolve any issues locally with installing the dependency.

If we were to upgrade `psycopg` to 2.9.x, there is an issue with Django where
UTC support [2] would need to be disalbed, though the impact to the service is
not fully understood (especially on Django 2), which is why we should leave the
version at 2.8.6 if possible for now.

[1] https://endoflife.date/python
[2] https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/rbac/settings.py#L188


astrozzc (approved):
LGTM

astrozzc:
Waiting for PR build but looks solid to me

PR #711, opened 2022-06-06
Fix default value on `build_internal_user`

coderbydesign:
Since `org_id` doesn't exist on a Turnpike user's identity, we can't default
the value if we aren't able to find it off the request path.

In this case, it means we're accessing an internal endpoint that's not specific
to an org/tenant anyway, so we do not need it on the user.


wcmitchell (approved):
:+1: :crossed_fingers:

PR #723, opened 2022-06-29
Fix internal openapi.json spec route

coderbydesign:
This was working locally because of the server being run not being gunicorn. Once
deployed, this started throwing a 500. Locally it was resolving, but also silently
recording a 404 against the `_private/` path.

Testing with `make gunicorn-serve` locally exposed the issue. The prefix change
resolves the problem and renders the json without warnings or errors.


wcmitchell (approved):
👍

PR #733, opened 2022-07-18
Bump gunicorn workers

coderbydesign:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-20254

We're running into a handful of sporadic timeout events (and pod restarts due to failed probes) in RBAC production, which we think may be caused by gunicorn under-resourcing, since our memory, cpu and db metrics do not appear to be spiking during these times.

All RBAC environments set a `CPU_LIMT` of `1` by default - https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/insights/rbac/deploy-clowder.yml#L57

`POD_CPU_LIMIT` is populated by `limits.cpu` on a pod - https://github.com/RedHatInsights/insights-rbac/blob/master/deploy/rbac-clowdapp.yml#L363

We set workers based on `POD_CPU_LIMIT * 2` - https://github.com/RedHatInsights/insights-rbac/blob/fb86e205b390d0ac7925fe58a17a2497a88d196f/rbac/gunicorn.py#L7 which means we have 2 workers, with 10 threads

[The docs](https://docs.gunicorn.org/en/stable/design.html#how-many-workers) suggest `(2 * cores) + 1` which would give us 3, but a guideline also suggests reasonable ranges up to `4 * cores` which would give us 4.

This change proposed we bump the workers to 4, since we're currently not at risk of hitting memory quota, which we can still track after bumping workers to determine any impact. This would allow us to not have to obtain more workers by increasing CPU out of the gate.

## Local Testing
Run `make gunicorn-serve` to see the new (4 workers * 10 threads) worker/thread count locally.

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/gunicorn.py
+++ b/rbac/gunicorn.py
@@ -4,6 +4,6 @@ import os
 
 bind = "unix:/var/run/rbac/gunicorn.sock"
 cpu_resources = int(os.environ.get("POD_CPU_LIMIT", multiprocessing.cpu_count()))
-workers = cpu_resources * 2
+workers = cpu_resources * 4
 threads = 10
 limit_request_field_size = 16380


mjholder (approved):
LGTM!

PR #734, opened 2022-07-20
Parameterize gunicorn worker count multiplier

coderbydesign:
A follow-up to #733 where we bumped the multiplier from 2 to 4. This defaults
to 2 (as it was) and will allow us to override the param in production to bump
as needed.


wcmitchell (approved):
LGTM!

PR #737, opened 2022-07-22
Restrict `/group/<uuid>/` DELETEs to `system` groups only

coderbydesign:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-19159

By default, the shared "Default access" group is `system=True` and `platform_default=True`,
tied to the `public` tenant.

When a `platform_default=True` group does not exist for a specific account/org tenant,
the `public` tenant "Default access" group is used.

Once a tenant modifies their default access, we create a new "Custom default acces"
group, which is `system=False` and `platform_default=True`. The `system` flag being
set to `False` means the platform no longer manages this group.

Given that we now need to allow for "reverting" back to the platform-managed "Default
access" group, the way we're going to go about this is to allow deletion of custom
default groups, where `system=False` and `platform_default=True`.

Admin default access is also `system=True`, so keepin the protection of groups based
on `system=True` will continue to protect the admin and default access groups on
the public tenant.

This will require integration with the UI, so that when someone wants to "revert"
their custom default access to be in line with the platform default group, the
UI can issue a DELETE against this custom default access group for the tenant.
This will delete the group (which was cloned at the time of "creation"), and based
on this query [1] will then grab the public default access group.

[1] https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/utils.py#L138-L140

## Local Testing
- modify the "Default access" group by adding/removing a role
- ensure that `/groups/` shows the "Custom default access" group and not "Default access"
- DELETE your "Custom default access group"
- ensure that `/groups/` shows the "Default access" group and not "Custom default access"

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


mjholder (approved):
Tested locally and it works as expected. LGTM!

wcmitchell (approved):

PR #741, opened 2022-07-28
Creating a "Custom default access" group should only use public tenant roles by default

coderbydesign:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-20603

Currently the query to grab roles for a custom default group [1] has an implicit
expectation that any `platform_default` roles will be coming from the public tenant
as that's the only way they can be created via seeds.

There's an edge-case where in stage, as part of our public schema data migration,
one role from a non-public tenant (an old `platform_default` role from the tenant's
schema) was copied into the public schema, still tied to that tenant.

This in turn means that when a new custom default group is cut in stage, it will
include that duplicate group since the filter doesn't explicitly require the
tenant be the public tenant.

These changes close that edge-case.

[1] https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/group/definer.py#L100

## Local Testing
- on master, manually create a duplicate `platform_default` role, which already exists
  for the public tenant, and update the `tenant_id` to be that of another tenant
- update the "Default access" group, which will create a "Custom default access"
  group, and note that if you query `/rbac/api/v1/groups/<uuid>/roles/` you'll
  see the duplicate role from the non-public tenant
- check these changes out and run the same test as above, and note the duplicate
  role does not exist on the new "Custom default access" group

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/management/group/definer.py
+++ b/rbac/management/group/definer.py
@@ -85,6 +85,7 @@ def set_system_flag_before_update(group, tenant, user):
 
 def clone_default_group_in_public_schema(group, tenant):
     """Clone the default group for a tenant into the public schema."""
+    public_tenant = Tenant.objects.get(tenant_name="public")
     tenant_default_policy = group.policies.get(system=True)
     group.name = "Custom default access"
     group.system = False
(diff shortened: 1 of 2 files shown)


wcmitchell (approved):
Good change! And good find by QE!

PR #748, opened 2022-09-01
Explicitly set Kafka API version to 1.0.2

coderbydesign:
Context is in #743, but it's been reported that 1.0.2 may resolve some missing
header issues as well. It looks like this is in fact still needed, as we're now
seeing `kafka.errors.UnrecognizedBrokerVersion: UnrecognizedBrokerVersion` with
managed Kafka.

If this does help the issues we're seeing, we should parameterize the version number.


casey-williams-rh (approved):
👍

PR #757, opened 2022-10-06
Ensure the notifications topic is set in the worker

coderbydesign:
When seeds run, we need to ensure we set `NOTIFICATIONS_TOPIC` on the worker pods, otherwise the topic will not be set for the producer, and will cause the client to connect and disconnect indefinitely.


lpichler:
https://issues.redhat.com/browse/RHCLOUD-23368

PR #758, opened 2022-10-12
Fix `username_only` spec type

coderbydesign:
The `username_only` param should be updated in both internal and external API specs to a boolean, to be consistent with other params.


mjholder (approved):
LGTM

PR #778, opened 2023-01-03
tox config update for linters

petracihalova:
Because tests and linters check via tox is not running on my MacBook with Apple chip (I think it is because incompatibility with module pipenv - the virtual env is created via pipenv but I am not able to install dependencies). I believe this solution is more simple and faster (because for running linters we dont install not needed dependencies).

Tox config update for running linters.
1. added description of all ignored types of flake8 warnings
2. D212 warning removed because it is not present in the codebase so it is not needed to skip it
3. removed not needed commands - for running linters we dont need install project dependencies 

Tox config update for running tests
1. dependencies are installed from requirements.txt

## Local Testing
run linters check `tox -elint` 
run tests `tox -epy38`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -178,7 +178,7 @@ This will rebuild the tox virtual env and then run all tests.
 
 To run unit tests specifically::
 
-    tox -e py36
+    tox -e py38
 
 To lint the code base ::
 
--- a/tox.ini
+++ b/tox.ini
@@ -34,13 +34,10 @@ setenv =
   MOCK_KAFKA=True
   KAFKA_ENABLED=True
 deps =
-  pipenv
   codecov
+  faker
 commands =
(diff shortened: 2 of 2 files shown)


lpichler:
@petracihalova do those changes requires any change in readme ? Like set export LDFLAGS for Apple chip, .. ?

petracihalova:
> @petracihalova do those changes requires any change in readme ? Like set export LDFLAGS for Apple chip, .. ?

yes and no
no, because this change is related with running linter and tests via tox and commands stays same as is in readme file
the export of LDFLAGS etc. is related with running the local virtual environment 

yes, because one tox command should be updated
now we have in readme
`tox -e py36`
and we are using
`tox -e py38`

lpichler (approved):

PR #782, opened 2023-01-10
[RHCLOUD-23571] Transition to ubi8-minimal image & remove s2i scripts

abaiken:
## Link(s) to Jira
- RHCLOUD-23571

RBAC is not based on the ubi8-minimal image. We need to tranisiton to Ubi8-minimal for easier maintenance and upkeep. 
* Move to UBI8-minimal 
* Remove s2i scripts 
* Remove apidoc 
* Changes the way the git commit is put into the image 
* Run migrations/seeding inside init container on the rbac-service 

## Local Testing
1. Bring up docker-compose like normal and ensure everything is working (I changed the docker-compose to use our actual image) 
2. Deploy to ephemeral manually & ensure it works 

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


syncrou (commented):
Great to see this!

jhutar:
Hello. When we tried to deploy this PR images to Perf cluster, rbac-service fails to start with this in the log: `Error: 'gunicorn.py' doesn't exist`

abaiken:
> Hello. When we tried to deploy this PR images to Perf cluster, rbac-service fails to start with this in the log: `Error: 'gunicorn.py' doesn't exist`

Hey Jan! I know. I am trying to resolve this issue currently.

lpichler:
> Run migrations/seeding inside init container on the rbac-service

What are reason(s) to have this in rbac-service ? When the RBAC service server starts, does it also run seed process? If so, do we want to have here also ?

abaiken:
> > Run migrations/seeding inside init container on the rbac-service
> 
> What are reason(s) to have this in rbac-service ? When the RBAC service server starts, does it also run seed process? If so, do we want to have here also ?

@lpichler My reasoning was that the service depends on the migrations in order to become ready so I thought it made more sense to have them on an init-container on the service pod versus the worker pod. When written this way, it does run the seed process on startup - I'm still new so I will divert to someone who knows more about RBAC on if we want this or not - I just used the existing script which runs both migrations & seeding.

lpichler:
> > > Run migrations/seeding inside init container on the rbac-service
> > 
> > 
> > What are reason(s) to have this in rbac-service ? When the RBAC service server starts, does it also run seed process? If so, do we want to have here also ?
> 
> @lpichler My reasoning was that the service depends on the migrations in order to become ready so I thought it made more sense to have them on an init-container on the service pod versus the worker pod. When written this way, it does run the seed process on startup - I'm still new so I will divert to someone who knows more about RBAC on if we want this or not - I just used the existing script which runs both migrations & seeding.

@abaiken yes it does make sense to me to have migration in init container. Seed process is ran during rbac service startup - so without seed process it would be better. (or we can remove it from startup)
If there is no any relation with changes in this PR - what about to do this change in other PR/task ?

abaiken:
> > > > Run migrations/seeding inside init container on the rbac-service
> > > 
> > > 
> > > What are reason(s) to have this in rbac-service ? When the RBAC service server starts, does it also run seed process? If so, do we want to have here also ?
> > 
> > 
> > @lpichler My reasoning was that the service depends on the migrations in order to become ready so I thought it made more sense to have them on an init-container on the service pod versus the worker pod. When written this way, it does run the seed process on startup - I'm still new so I will divert to someone who knows more about RBAC on if we want this or not - I just used the existing script which runs both migrations & seeding.
> 
> @abaiken yes it does make sense to me to have migration in init container. Seed process is ran during rbac service startup - so without seed process it would be better. (or we can remove it from startup) If there is no any relation with changes in this PR - what about to do this change in other PR/task ?

@lpichler, Keith, and I talked about this in slack and we came to the conclusion that it is okay to leave this as is since seeding should be a no-opp most of the time 👍

Dockerfile
```diff
@@ -35,41 +32,81 @@ LABEL summary="$SUMMARY" \
...
-#   writable as OpenShift default security model is to run the container
-#   under random UID.
-RUN pip install virtualenv && \
-    virtualenv ${APP_ROOT} && \
-    chown -R 1001:0 ${APP_ROOT} && \
-    fix-permissions ${APP_ROOT} -P && \
-    rpm-file-permissions && \
-    $STI_SCRIPTS_PATH/assemble || true
 
-RUN curl -L -o /usr/bin/haberdasher \
```
lpichler, line 65: great to see this - there is also lot of environment variables with name `HABERDASHER` which needs to be removed also - there are not needed now(+ app interface in `data/services/insights/rbac/deploy-clowder.yml`) - but you can do it in follow up PR.

deploy/rbac-clowdapp.yml
```diff
@@ -181,6 +181,8 @@ objects:
             value: ${NOTIFICATIONS_TOPIC}
           - name: EXTERNAL_SYNC_TOPIC
             value: ${EXTERNAL_SYNC_TOPIC}
+          - name: SHOULD_MIGRATE
```
lpichler, line 184: As this looks that it will be used only for init container - maybe it would be nice that name would contain that fact, something like:
`INIT_MIGRATE`

Other fact is there is also seed process and not only migrate but with this I am not sure how to name it :)
something like `MIGRATE_AND_SEED_ON_INIT` 

maybe you will have better idea :) Let me know what do you think, thanks!

lpichler (approved):

Dockerfile
```diff
@@ -1,24 +1,21 @@
-FROM registry.access.redhat.com/ubi8/python-38
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest AS base
```
vbelchio, line 1: @abaiken, is it not better to force the number of the version to avoid future unexpected changes?

abaiken: It likely is better for consistency to specify the version - I pushed up a change 😁

vbelchio:
lgtm

PR #789, opened 2023-01-17
Add the `system` flag to the internal integration spec

coderbydesign:
The internal integrations endpoint for `/_private/api/v1/integrations/tenant/11111/groups/` currently supports `?system=true|false` to filter in or out system groups, but it's not reflected in the spec.

This will add it to the spec which should allow clients to use the param.


abaiken (approved):
LGTM

Ellen-Yi-Dong (approved):
LGTM!

PR #810, opened 2023-02-15
[RHCLOUD-24172] RBAC service should return error for non existing user

petracihalova:
## Link(s) to Jira
- [RHCLOUD-24172](https://issues.redhat.com/browse/RHCLOUD-24172)

if x-rh-identity header contains non existing username, we should return 400 bad request
we already had this check for non existing usernames that were given in the query
e.g. `GET /api/rbac/v1/access/?application=&username=blabla`

this PR adds the same check for usernames from the x-rh-identity header
e.g. `GET /api/rbac/v1/access/?application=`
with x-rh-identity header
```
{
  "identity": {
    "type": "User",
    "user": {
      "username": "non_existing_username_lksjfhskdfhl"
    },
    "internal": {
      "org_id": "12345"
    }
  }
}
```
the response will be 400 Bad Request
```
{
    "errors": [
        {
            "detail": "No data found for principal with username 'non_existing_username_lksjfhskdfhl'.",
            "source": "detail",
            "status": "400"
        }
    ]
}
```

### DIFF
--- a/rbac/management/utils.py
+++ b/rbac/management/utils.py
@@ -61,7 +61,7 @@ def get_principal_from_request(request):
         username = qs_user
         from_query = True
 
-    return get_principal(username, request, verify_principal=bool(qs_user), from_query=from_query)
+    return get_principal(username, request, verify_principal=True, from_query=from_query)
 
 
 def get_principal(username, request, verify_principal=True, from_query=False):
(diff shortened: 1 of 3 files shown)


abaiken (approved):
LGTM - thanks for the pointers on BOP 🎉

PR #798, opened 2023-02-03
[RHCLOUD-24242] Security vulnerability fixes

abaiken:
## Link(s) to Jira
- [RHCLOUD-24242](https://issues.redhat.com/browse/RHCLOUD-24242)

## Description of Intent of Change(s)
After the update to ubi8-minimal & updating Django the following packages were still left with vulnerabilities: 
```
setuptools
kombu 
daphne 
celery 
pip 
py (NO FIX)
```
* Celery, kombu, setuptools, & pip packages were updated
* py is a no fix 
* daphne is a package introduced by channels-redis which is being removed in this PR 
* This PR removes websocket support which was implemented in this [PR](https://github.com/RedHatInsights/insights-rbac/pull/657) but never used 
## Local Testing
* View the image in quay and check the vulnerabilities. Setuptools & Pip are flagging but they are false positives after speaking with Casey 

## Checklist
- [ ] if API spec changes are required, is the spec updated?
- [ ] are there any pre/post merge actions required? if so, document here.
- [ ] are theses changes covered by unit tests?
- [ ] if warranted, are documentation changes accounted for?
- [ ] does this require migration changes?
  - [ ] if yes, are they backwards compatible?
- [ ] is there known, direct impact to dependent teams/components?
  - [ ] if yes, how will this be handled?

## Secure Coding Practices Checklist Link
- https://github.com/RedHatInsights/secure-coding-checklist

## Secure Coding Practices Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [ ] General Coding Practices

PR #813, opened 2023-02-16
requirements.txt aligned with pipfile

petracihalova:
I think we need to clean the mess we have in pipfiles and requirements files, first step is to align the `pipfile` and `requirements.txt`

this change is result of command
`pipenv requirements > requirements.txt `

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,11 +2,11 @@
 aioredis==1.3.1
 amqp==2.6.1 ; python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3, 3.4'
 anyio==3.6.2 ; python_full_version >= '3.6.2'
-app-common-python==0.2.3
+app-common-python==0.2.4
 asgiref==3.6.0 ; python_version >= '3.7'
 async-timeout==4.0.2 ; python_version >= '3.6'
 attrs==22.2.0 ; python_version >= '3.6'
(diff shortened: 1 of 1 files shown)


abaiken:
I think this is fine to do for now - a lot of this will be cleaned up in my PR to remove the rest of the vulnerabilities from the rbac image - https://github.com/RedHatInsights/insights-rbac/pull/798/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552. We could even edit the `make requirements` command in the makefile to also update the requirements.txt and not just the rtd_requirements.txt so that if someone changes the pipfile - they just run `make requirements` and it automatically updates the other requirements files to try to keep everything in sync better 😁

petracihalova:
> I think this is fine to do for now - a lot of this will be cleaned up in my PR to remove the rest of the vulnerabilities from the rbac image - https://github.com/RedHatInsights/insights-rbac/pull/798/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552. We could even edit the `make requirements` command in the makefile to also update the requirements.txt and not just the rtd_requirements.txt so that if someone changes the pipfile - they just run `make requirements` and it automatically updates the other requirements files to try to keep everything in sync better 😁

100% agree...we can close this one, I didn't notice it was part of your PR 😁😁😁

PR #829, opened 2023-03-13
[RHCLOUD-5481] Optional trailing slashes 

petracihalova:
## Link(s) to Jira
- [RHCLOUD-5481](https://issues.redhat.com/browse/RHCLOUD-5481)

**CURRENT STATE:** 
endpoints  contain trailing slash 
without trailing slash we get response with error in the html format (status 404)

example of request without trailing slash:
```
POST api/rbac/v1/groups
{
  "name": "Group_name_2",
  "description": "A description of Group"
}
```

response:
```
Status 404 Not Found

<!doctype html>
<html lang="en">
<head>
    <title>Not Found</title>
</head>
<body>
    <h1>Not Found</h1>
    <p>The requested resource was not found on this server.</p>
</body>
</html>
```

this current state is affected by:
- Django settings `APPEND_SLASH = False` ([docs here](https://docs.djangoproject.com/en/4.1/ref/settings/#append-slash))
- `rest_framework.routers.DefaultRouter()`  as our router ([docs here](https://www.django-rest-framework.org/api-guide/routers/#defaultrouter)) with default value `trailing_slash=True`
- using trailing slash in `urlpatterns` for endpoints that are not registered in `DefaultRouter()` - e.g. api/rbac/v1/principals/

Django settings can be found [here](https://github.com/RedHatInsights/insights-rbac/blob/4c72d551c81ca69f7bfc6d8a82bb90dbc46ff0c6/rbac/rbac/settings.py#L305). `DefaultRouter()` and `urlpatterns` for example [here](https://github.com/RedHatInsights/insights-rbac/blob/4c72d551c81ca69f7bfc6d8a82bb90dbc46ff0c6/rbac/management/urls.py#L23-L33).

**ANALYSIS:**
We cannot change Django settings `APPEND_SLASH` to `True` because then if the request does not match the pattern, the request is redirected to the same url with trailing slash, but this redirecting causes the data submitted in a POST request to be lost (and redirected request is send as GET).

We can change the parameter `trailing_slash to False`
```
ROUTER = DefaultRouter(trailing_slash=False)
```
then the endpoints registered in `DefaultRouter()` are working without trailing slash and we can have our api that works without trailing slashes 
(in this case we need to do appropriate change for endpoints that are not registered in `DefaultRouter()`)

We can have api that accepts both options - so it won't matter if the endpoint contains trailing slash or not and both options will return the same result.
But I believe it is fine to have an api with all its endpoints end with a slash or not end with a slash -> so there is no need to support both options.

So my recommendation is to not change anything. All our endpoints now contain trailing slash (and it is also stated this way in our openapi spec).

However, if we want, we can set our API to accept both variants by defining a custom router as captured in this PR. Please note that this is only a draft and the PR does not contain the necessary changes for internal endpoints.

## Local Testing
just test that endpoints are working correctly with and without trailing slash

## Checklist

### DIFF
--- a/rbac/api/cross_access/view.py
+++ b/rbac/api/cross_access/view.py
@@ -16,6 +16,8 @@
 #
 
 """View for cross access request."""
+import re
+
 from django.conf import settings
 from django.db.models import Q
 from django.utils import timezone
@@ -123,7 +125,7 @@ class CrossAccountRequestViewSet(
 
     def get_serializer_class(self):
         """Get serializer based on route."""
-        if self.request.path.endswith("cross-account-requests/") and self.request.method == "GET":
+        if re.match(".*/cross-account-requests/?$", self.request.path) and self.request.method == "GET":
             return CrossAccountRequestSerializer
         return CrossAccountRequestDetailSerializer
 
--- a/rbac/api/urls.py
+++ b/rbac/api/urls.py
@@ -16,16 +16,17 @@
 """Describes the urls and patterns for the API application."""
 from django.conf.urls import include
 from django.urls import re_path
-from rest_framework.routers import DefaultRouter
+from management.urls import OptionalSlashRouter
 
 from api.views import CrossAccountRequestViewSet, openapi, status
 
-ROUTER = DefaultRouter()
+
+ROUTER = OptionalSlashRouter()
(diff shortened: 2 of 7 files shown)


petracihalova:
we decide to not change anything because the issue reported in JIRA ticket is not a bug, it is how our API is buld (with trailing slashes in all our endpoints)

PR #831, opened 2023-03-13
[RHCLOUD-24682] Remove principal from group (= remove relation between principal and group in the management_group_principal table)

petracihalova:
## Link(s) to Jira
- [RHCLOUD-24682](https://issues.redhat.com/browse/RHCLOUD-24682)

when we try to delete not existing principal from a group
we get 500 Internal Server Error because there is [referenced variable before assignment](https://github.com/RedHatInsights/insights-rbac/blob/4c72d551c81ca69f7bfc6d8a82bb90dbc46ff0c6/rbac/management/group/view.py#L388)
+
in this case we should return 404

## Local Testing
`DELETE api/rbac/v1/groups/{{uuid}}/principals/?usernames=not-valid-username`

### DIFF
--- a/rbac/management/group/view.py
+++ b/rbac/management/group/view.py
@@ -382,9 +382,19 @@ class GroupViewSet(
 
             for username in principals:
                 try:
-                    principal = Principal.objects.get(username__iexact=username, tenant=tenant)
+                    principal = Principal.objects.get(username__iexact=username, tenant=tenant, group=group)
                 except Principal.DoesNotExist:
                     logger.info("No principal %s found for org id %s.", username, org_id)
+                    return {
+                        "status_code": status.HTTP_404_NOT_FOUND,
+                        "errors": [
+                            {
+                                "detail": f"User '{username}' not found in the group '{group.name}'.",
+                                "status": "404",
+                                "source": "principals",
+                            }
+                        ],
+                    }
                 if principal:
                     group.principals.remove(principal)
                     group_principal_change_notification_handler(self.request.user, group, username, "removed")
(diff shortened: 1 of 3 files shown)


vbelchio:
LGTM

rbac/management/group/view.py
```diff
@@ -382,9 +382,19 @@ def remove_principals(self, group, principals, account=None, org_id=None):
...
-                    principal = Principal.objects.get(username__iexact=username, tenant=tenant)
+                    principal = Principal.objects.get(username__iexact=username, tenant=tenant, group=group)
                 except Principal.DoesNotExist:
                     logger.info("No principal %s found for org id %s.", username, org_id)
+                    return {
+                        "status_code": status.HTTP_404_NOT_FOUND,
+                        "errors": [
+                            {
+                                "detail": f"User '{username}' not found in the group '{group.name}'.",
+                                "status": "404",
```
dagbay-rh, line 393: nitpick: instead of hard coding 404 can we reference `status.HTTP_404_NOT_FOUND` like above? just looks a little cleaner imo

dagbay-rh (approved):

rbac/management/group/view.py
petracihalova: good catch 👍

abaiken (approved):

rbac/management/group/view.py
```diff
@@ -382,9 +382,19 @@ def remove_principals(self, group, principals, account=None, org_id=None):
 
             for username in principals:
                 try:
-                    principal = Principal.objects.get(username__iexact=username, tenant=tenant)
+                    principal = Principal.objects.get(username__iexact=username, tenant=tenant, group=group)
```
lpichler, line 385: My suggestion is here to check existence of all usernames in DB before we process them:
- when all of them exist
  - then we should process all 
- otherwise return err and don't process any of them.

petracihalova:
updated logic for removing relation between principal and group (removing record from table `management_group_principals`) ... now we check if all given usernames are valid otherwise we return 404 so in case 1 valid + 1 invalid username the 404 is returned (before this change, the first user was deleted and processing the second resulted in a 404)

vbelchio:
LGTM
Tests are passing locally

dagbay-rh (approved):

petracihalova:
rebased

PR #842, opened 2023-03-24
groups_in returns default roles when not in group

dagbay-rh:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-17918

Right now, when hitting `/api/rbac/v1/roles/?add_fields=groups_in` and either `&scope=principal` or `&username=...`, roles will be listed as part of `platform_default` roles even if they are not in that group. These changes fix that by filtering default groups that are not related to the policies of a given role. 

There is also a handful of whitespace changes that were made when running the linter locally. The changes related to the ticket specifically are in `rbac/management/role/serializer.py` and `tests/management/role/test_view.py`. I can remove the excess linter changes if desired.

## Local Testing
- Running locally and confirming the `groups_in` field in each role from `/api/rbac/v1/roles/?add_fields=groups_in&scope=principal` matches whats expected given the default db seeding
- without fix, payload includes this role which is listed as a part of the platform default group even tho it is not:
```
        {
            "uuid": "a5816cca-221d-4e70-bf6a-19de53d9ef0e",
            "name": "RBAC Administrator Local Test",
            "display_name": "RBAC Administrator Local Test",
            ...
            "groups_in": [
                {
                    "name": "Default access",
                    "uuid": "cbf8ee36-5128-49c0-8916-a68cc1b13766",
                    "description": "This group contains the roles that all users inherit by default. Adding or removing roles in this group will affect permissions for all users in your organization."
                }
            ],
            ...
        },
```
- with the fix, `groups_in` is an empty array as expected

~~No, I was able to repro the bug and confirm the fix with the default db seeding from `make run-migrations`.~~
Yes and no:

To test query as an admin: hit `.../roles?add_fields=groups_in` and confirm only the roles actually in the default access group are listed as in that group, and the roles in the default admin group are listed as in that group. With the default data set from `make run-migrations` there should be 7 in the default access group and one in the default admin group.

To test query for non admin: 
Set up some data: create a new role, create a new group, create a new policy linking the new role to the new group, add the principal for the user you are using (default with `dev_middleware` is `user_dev`) to the new group

Then: hit `.../roles?add_fields=groups_in&scope=principal`  and the new role you made should only be listed as in the new group you made, not the default access group

Note that this was not actually reproducible in master because the current query filters the entire result set on the request tenant, which removed groups belonging to the public tenant aka the default groups. So the `groups_in` field was still not working correctly but for slightly different reasons

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


.pre-commit-config.yaml
```diff
@@ -12,7 +12,7 @@ repos:
   rev: 22.3.0
   hooks:
   - id: black
-    args: ["--check", "-l", "119", "-t", "py39", "rbac", "tests"]
+    args: ["--check", "-l", "119", "-t", "py39", "rbac", "tests", "--diff"]
```
dagbay-rh, line 15: added `--diff` flag here to show what changes to make if check fails

README.rst
```diff
@@ -75,7 +75,7 @@ PostgreSQL is used as the database backend for Insights-rbac. A docker-compose f
 
 Assuming the default .env file values are used, to access the database directly using psql run ::
 
-    psql rbac -U rbacadmin -h localhost -p 15432
+    psql postgres -U postgres -h localhost -p 15432
```
dagbay-rh, line 78: default user/pass was `postgres` for me locally, so updated this. afaik they are the default but if not ill revert

abaiken: Thanks for updating this ^ i've been meaning to fix that 😂

tests/management/role/test_view.py
```diff
@@ -649,6 +659,42 @@ def test_list_role_with_additional_fields_success(self):
...
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        for keyname in ["meta", "links", "data"]:
+            self.assertIn(keyname, response.data)
+        self.assertIsInstance(response.data.get("data"), list)
+
+        for iterRole in response.data.get("data"):
+            self.assertEqual(new_display_fields, set(iterRole.keys()))
+
+            # a role thats in 2 groups, one is platform default
+            if iterRole.get("name") == "default_role":
```
dagbay-rh, line 683: - [ ] need to rework this test: this will only execute the asserts if there are roles with specified names, if they are not returned the test is useless
- [ ] add another test to test both a principal and not principal scoped request

petracihalova (approved):
nice job 🥇

PR #866, opened 2023-06-06
Add external source id to principals

coderbydesign:
This returns `external_source_id` as part of a `/principals/` GET request, in order
to tie user records back to the external source of truth.

These IDs should be populated by bop/mbop regardless of the source of truth in
a given environment and is already part of the API spec for bop.

We may want to include a source name mapping at some point as well, but since this
is environment specific, it should be implicit knowledge from whoever is consuming
and requiring this new field.


vbelchio:
@coderbydesign is there some JIRA ticket for this?
Maybe we need to change our integration tests after this change

PR #887, opened 2023-07-27
Add Worker Alerts

Ellen-Yi-Dong:
## Link(s) to Jira
- [RHCLOUD-25588](https://issues.redhat.com/browse/RHCLOUD-25588)

The what, why and how.
Adding worker alerts so when celery or redis down, there will be a 500 status and it times out. 

We can then check the state of the pods as well through the grafana dashboard. 

## Local Testing

Here is a curl command to hit the endpoint: curl -v -X GET -H "Accept: application/json" http://localhost:8000/api/rbac/ht/

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


Ellen-Yi-Dong:
> hi @Ellen-Yi-Dong is there a ticket we can link to this?

Oh yes there is!! I completely forgot to update the summary for some reason! I'll do that right now! Thank you so much for letting me know!

rbac/rbac/urls.py
```diff
@@ -38,7 +39,8 @@
     re_path(r"^{}v1/".format(API_PATH_PREFIX), include("api.urls")),
     re_path(r"^{}v1/".format(API_PATH_PREFIX), include("management.urls")),
     re_path(r"^_private/", include("internal.urls")),
-    path("", include("django_prometheus.urls")),
+    re_path(r"", include("django_prometheus.urls")),
+    path(r"{}ht/".format(API_PATH_PREFIX), include("health_check.urls")),
```
dagbay-rh, line 43: just curious here, where does `/ht/` come from? I just wouldn't guess `ht` means health check endpoint

also is there an api spec to update with the new endpoint? not sure if thats automatically generated or something with rbac

Ellen-Yi-Dong: The "/ht/" endpoint was what was suggested from the django-health-check github readme: [https://github.com/revsys/django-health-check/tree/master](https://github.com/revsys/django-health-check/tree/master) but I can change it to "health" if that's more preferable!

dagbay-rh: ok! no i would go with what the docs recommend for sure :) they know more than me heh

abaiken (approved):
🎉

dagbay-rh (approved):
LGTM! Nice work :)

lpichler:
Hey @Ellen-Yi-Dong, 

with this change - worker is keeping failing 
https://console-openshift-console.apps.crcs02ue1.urby.p1.openshiftapps.com/k8s/ns/rbac-stage/pods/rbac-worker-service-cd86f787d-fv6gc

it is because endpoint `ht` is not reachable 

https://console-openshift-console.apps.crcs02ue1.urby.p1.openshiftapps.com/k8s/ns/rbac-stage/pods/rbac-worker-service-cd86f787d-fv6gc/events


I think you cannot directly call API to another pods  in liveness and probe checks  - I think you can call it only on that pods - so if you want to use that API call on rbac service I think you would need to add api server to the worker and that call rbac service from that server - but It looks it is not necessary.


  
but now I see two  liveness and probe checks for worker ? do we need to add another one ? is it enough to have the original one ?

PR #900, opened 2023-09-06
Audit Logs for Roles and Groups  (Creating, Deleting, and Editing Roles/Groups  within a Tenant) 

Ellen-Yi-Dong:
## Link(s) to Jira
-  [RHCLOUD-19119](https://issues.redhat.com/browse/RHCLOUD-19119)

The what, why and how.
Created the audit logs for when a role or group is created, deleted or edited within a tenant. 
(Will continue working on the next few aspects in different tickets. Just breaking it down in a smaller PR for feedback and to see if it works) 

## Local Testing

Here is the related URL to check to see the whole audit logs: 
`curl localhost:8000/api/rbac/v1//auditlogs/`
## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


rbac/management/group/view.py
```diff
@@ -211,7 +213,13 @@ def create(self, request, *args, **kwargs):
...
         """
         validate_group_name(request.data.get("name"))
-        return super().create(request=request, args=args, kwargs=kwargs)
+        create_group = super().create(request=request, args=args, kwargs=kwargs)
+        if status.is_success(create_group.status_code):
+            auditlog = AuditLogModel()
+            auditlog.create_data(request, AuditLogModel.GROUP)
+            return create_group
+
+        # return super().create(request=request, args=args, kwargs=kwargs)
```
petracihalova, line 222: 🤔 maybe forgotten comment?

rbac/management/auditLogs/model.py
```diff
@@ -0,0 +1,82 @@
...
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+"""Model for audit logging."""
+from django.db import models
+
+
+class AuditLogModel(models.Model):
```
lpichler, line 22: 1. I believe that we have naming convention without model word. I would prefer `AuditLog`.
2. if model is using Tenancy (corresponding  table has tenant_id column) model inherits from `TenantAwareModel` for example like in role model `class Role(TenantAwareModel):`

rbac/management/auditLogs/model.py
```diff
@@ -0,0 +1,82 @@
...
+        (REMOVE, "Remove"),
+    )
+
+    date = models.DateField(auto_now_add=True)
+    requester = models.TextField(max_length=255, null=False)  # probs need a max length allowed here
+    description = models.TextField(max_length=255, null=False)  # probs need a max length allowed here
+    resource = models.CharField(max_length=32, choices=RESOURCE_CHOICES)
+    action = models.CharField(max_length=32, choices=ACTION_CHOICES)
+
+    def create_data(self, request, resource):
```
lpichler, line 55: I think that name could be just `create` it implies to create one record in Audit Log table.

lpichler: OK, I misunderstood it - you are basically creating log records which would not makes in update/delete method. So maybe it would be better to have word as suffix. But I am thinking about names of methods - name should represent fact that:

I am creating log record about create/update/delete action, naming is hard

lpichler: all methods are going to create log record so maybe it could something like 
log_create_action
log_delete_action 

not sure exactly,  I am definitely interested in your ideas! 💡

rbac/management/migrations/0041_auditlogmodel.py
```diff
@@ -0,0 +1,65 @@
+# Generated by Django 4.1.7 on 2023-08-23 18:41
```
lpichler, line 1: it looks that there is one migration and some other migration which is changing the previous ones .  If you add one table, can you create just one final migration with adding one table ? remove/alter are needed only when database changes are in stage/production applied and we need to do some change.

rbac/management/auditLogs/model.py
```diff
@@ -0,0 +1,82 @@
+#
```
lpichler, line 1: I would prefer name of `audit_log` as we use singular form in project for models

lpichler:
@Ellen-Yi-Dong thanks for doing audit logs changes and I left few comments at this moment. Let me know if you need more information regard them and if you have other ideas.

I also I would prefer to create smaller PR for this:
1. PR for just migration and model (without logging actions)
2. PR with creating log records into db
3. API

Ideally I would split point 2. 
also to separate actions like one PR just with method for creating logs with create action, etc. This is just me but we can do better and faster review with this.


let me know what do you think!

thanks!!!

rbac/management/auditLogs/model.py
Ellen-Yi-Dong: Maybe we can have it even shorter like: 
log_create
log_delete? 
I do like having the word log in front to make it clear!

Ellen-Yi-Dong:
> @Ellen-Yi-Dong thanks for doing audit logs changes and I left few comments at this moment. Let me know if you need more information regard them and if you have other ideas.
> 
> I also I would prefer to create smaller PR for this:
> 
> 1. PR for just migration and model (without logging actions)
> 2. PR with creating log records into db
> 3. API
> 
> Ideally I would split point 2. also to separate actions like one PR just with method for creating logs with create action, etc. This is just me but we can do better and faster review with this.
> 
> let me know what do you think!
> 
> thanks!!!

I can do that!! 
Though do you think it would be okay to split into two PRs, where the migration of the new model with the new model methods as one PR, and then the other PR having the audit log code within the API and the unit tests? 

Thank you so much!! 
And thank you once again for reviewing my PR!!

lpichler:
> I can do that!! Though do you think it would be okay to split into two PRs, where the migration of the new model with the new model methods as one PR, and then the other PR having the audit log code within the API and the unit tests?

@Ellen-Yi-Dong yes, awesome, thanks!

PR #906, opened 2023-09-07
Which Workspaces in this Organization have access to this Service?

lpichler:
- migration to remove permission unique key from Permission
- migration to add key permission and workspace - this will allow us to have same permissions in different workspaces

Update pdp endpoints:

**list_workspaces=true** - enabling feature

**service=openshift&resource_type=host&list_workspaces=true** - filter if we ask for service or service with resource_type

**Which Workspaces in this Organization have access to this Service?**
```
GET http://localhost:8080/api/rbac/v1/access/?service=openshift&list_workspaces=true&pdp=true

{
  "access": [
    {
      "permission": "openshift:cluster:read",
      "resourceDefinitions": []
    }
  ]
}
```

**Which Workspaces in this Organization have access to this Service?** - more granular with resource_type restriction

```
GET http://localhost:8080/api/rbac/v1/access/?service=openshift&resource_type=host&list_workspaces=true&pdp=true


{
  "access": [
    {
      "permission": "openshift:cluster:read",
      "resourceDefinitions": []
    }
  ]
}
```


abaiken (approved):
LGTM!

PR #913, opened 2023-09-14
 RHCLOUD-28261 | feature: return stubs for the service accounts

MikelAlejoBR:
## Link(s) to Jira
-  [[RHCLOUD-28261]](https://issues.redhat.com/browse/RHCLOUD-28261)

The UI team is blocked because RBAC still doesn't fully support the service accounts. Even though the ADR is nearly finished, the sooner we can let the UI team consume the endpoints with the service accounts the better.

They have told us that it is not required to return real data, so that is why we are returning static stubs instead. Once the ADR is finished, we will remove them.

The default behavior is to honor old query parameters, that is why they have precedence over the rest.

### How can the feature be exercised?
The affected endpoints are:
- `/principals/`
- `/groups/{uuid}/principals/`

You can append the following query parameters to the endpoints:

- `GET /principals/?type=user`
- `GET /principals/?type=service-account`

- `GET /groups/{uuid}/principals/?type=user`
- `GET /groups/{uuid}/principals/?type=service-account`

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -644,6 +644,19 @@
                 false
               ]
             }
+          },
+          {
+            "in": "query",
+            "name": "type",
+            "required": false,
+            "description": "Parameter for selecting the type of principal to be returned.",
+            "schema": {
+              "type": "string",
+              "enum": [
+                "service-account",
+                "user"
+              ]
+            }
           }
         ],
         "responses": {
@@ -1237,6 +1250,19 @@
               "enum": [
                 true,
                 false
+              ]
+            }
+          },
+          {
+            "in": "query",
+            "name": "principal_type",
+            "required": false,
+            "description": "Parameter for selecting the type of principal to be returned.",
+            "schema": {
+              "type": "string",
+              "enum": [
+                "service-account",
+                "user"
               ]
             }
           }
--- a/rbac/management/group/view.py
+++ b/rbac/management/group/view.py
@@ -57,12 +57,14 @@ USERNAMES_KEY = "usernames"
 ROLES_KEY = "roles"
 EXCLUDE_KEY = "exclude"
 ORDERING_PARAM = "order_by"
+PRINCIPAL_TYPE_KEY = "principal_type"
 VALID_ROLE_ORDER_FIELDS = list(RoleViewSet.ordering_fields)
 ROLE_DISCRIMINATOR_KEY = "role_discriminator"
 VALID_EXCLUDE_VALUES = ["true", "false"]
 VALID_GROUP_ROLE_FILTERS = ["role_name", "role_description", "role_display_name", "role_system"]
 VALID_GROUP_PRINCIPAL_FILTERS = ["principal_username"]
 VALID_PRINCIPAL_ORDER_FIELDS = ["username"]
+VALID_PRINCIPAL_TYPE_VALUE = ["service-account", "user"]
 VALID_ROLE_ROLE_DISCRIMINATOR = ["all", "any"]
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
@@ -428,6 +430,8 @@ class GroupViewSet(
         @apiHeader {String} token User authorization token
 
         @apiParam (Path) {String} id Group unique identifier
+        @apiParam (Query) {String} principal_type Parameter for selecting the type of principal to be returned (either
+                                                  "service-account" or "user").
 
         @apiSuccess {String} uuid Group unique identifier
         @apiSuccess {String} name Group name
@@ -540,6 +544,16 @@ class GroupViewSet(
                     request.query_params, USERNAME_ONLY_KEY, VALID_BOOLEAN_VALUE, "false"
                 ),
             }
+
+            # Attempt validating and obtaining the "principal type" query
+            # parameter. For now, having "service-account" as the "principal
+            # type" returns a stub.
+            principalType = validate_and_get_key(
+                request.query_params, PRINCIPAL_TYPE_KEY, VALID_PRINCIPAL_TYPE_VALUE, required=False
+            )
+
+            if principalType == "service-account":
+                return self.user_service_accounts_response()
 
             admin_only = validate_and_get_key(request.query_params, ADMIN_ONLY_KEY, VALID_BOOLEAN_VALUE, False, False)
             if admin_only == "true":
@@ -745,3 +759,26 @@ class GroupViewSet(
         # Exclude the roles in the group
         roles_for_group = group.roles().values("uuid")
         return roles.exclude(uuid__in=roles_for_group)
+
+    def user_service_accounts_response(self):
+        """Return a stubbed user or service account response.
+
+        Returns a dictionary with a mocked empty array response, in order to use
+        the structure as the returning values when a "user" or "service-account"
+        principal types are specified as query parameters. For more information
+        check https://issues.redhat.com/browse/RHCLOUD-28261.
+        """
+        return Response(
+            data={
+                "meta": {"count": 0, "limit": 0, "offset": 0},
+                "links": {
+                    "first": "",
+                    "next": "",
+                    "previous": "",
+                    "last": "",
+                },
+                "data": [],
+                "status": status.HTTP_200_OK,
+            },
+            status=status.HTTP_200_OK,
+        )
--- a/rbac/management/principal/view.py
+++ b/rbac/management/principal/view.py
@@ -36,6 +36,8 @@ VALID_STATUS_VALUE = ["enabled", "disabled", "all"]
 ADMIN_ONLY_KEY = "admin_only"
 VALID_BOOLEAN_VALUE = ["true", "false"]
 USERNAME_ONLY_KEY = "username_only"
+PRINCIPAL_TYPE_KEY = "type"
+VALID_PRINCIPAL_TYPE_VALUE = ["service-account", "user"]
 
 
 class PrincipalView(APIView):
@@ -52,6 +54,8 @@ class PrincipalView(APIView):
 
     @apiParam (Query) {Number} offset Parameter for selecting the start of data (default is 0).
     @apiParam (Query) {Number} limit Parameter for selecting the amount of data (default is 10).
+    @apiParam (Query) {String} type Parameter for selecting the type of principal to be returned
+                                    (either "service-account" or "user").
 
     @apiSuccess {Object} meta The metadata for pagination.
     @apiSuccess {Object} links  The object containing links of results.
@@ -111,7 +115,16 @@ class PrincipalView(APIView):
         if offset - limit > 0:
             previous_offset = offset - limit
 
-        resp, usernames_filter = self.users_from_proxy(user, query_params, options, limit, offset)
+        # Attempt validating and obtaining the "prinicpal type" query
+        # parameter. For now, specifying the query parameter results
+        # in returning the mocked response.
+        principalType = validate_and_get_key(
+            query_params, PRINCIPAL_TYPE_KEY, VALID_PRINCIPAL_TYPE_VALUE, required=False
+        )
+        if principalType == "service-account":
+            resp = self.user_service_accounts_response()
+        else:
+            resp, usernames_filter = self.users_from_proxy(user, query_params, options, limit, offset)
 
         status_code = resp.get("status_code")
         response_data = {}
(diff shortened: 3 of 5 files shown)


rbac/management/group/view.py
```diff
@@ -541,6 +561,15 @@ def principals(self, request, uuid=None):
...
 
+            # Attempt validating and obtaining the "principal type" query
+            # parameter. For now, having "service-account" as the "principal
+            # type" returns a stub.
+            principalType = validate_principal_type(request.query_params)
+            if principalType == "service-account":
+                resp = service_accounts_response()
+
+                return self.get_paginated_response(resp["data"]["users"])
+
```
MikelAlejoBR, line 572: For some reason this one doesn't return the correct count in the `meta` key.

tests/management/group/test_view.py
```diff
@@ -2005,3 +2005,15 @@ def test_group_role_filter_as_non_admin(self):
...
+        """Test that removing the stub service accounts from a group returns successfully."""
+        test_user = Principal.objects.create(username="test_user", tenant=self.tenant)
+        self.group.principals.add(test_user)
+
+        url = reverse("group-principals", kwargs={"uuid": self.group.uuid})
+        client = APIClient()
+
+        url = "{}?service-accounts={}".format(url, "one,two,three")
+        response = client.delete(url, format="json", **self.headers)
+        self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
```
MikelAlejoBR, line 2019: This one fails because the endpoint returns a `403 Forbidden` status code.

MikelAlejoBR: I have tried adding the `send_kafka_message` and `mock_request` arguments to the test, similar to what the following test had...

```python
    @patch(
        "management.principal.proxy.PrincipalProxy.request_filtered_principals",
        return_value={"status_code": 200, "data": [{"username": "test_user"}]},
    )
    @patch("core.kafka.RBACProducer.send_kafka_message")
    def test_remove_group_principals_success(self, send_kafka_message, mock_request):
```

But that doesn't make it work either :/

rbac/management/principal/view.py
```diff
@@ -111,7 +115,14 @@ def get(self, request):
         if offset - limit > 0:
             previous_offset = offset - limit
 
-        resp, usernames_filter = self.users_from_proxy(user, query_params, options, limit, offset)
+        # Attempt validating and obtaining the "prinicpal type" query
+        # parameter. For now, having "service-account" as the "principal type"
+        # returns a stub.
+        principalType = validate_principal_type(query_params)
```
coderbydesign, line 121: Here we could define the valid values and call `validate_and_get_key` as well?

coderbydesign (commented):
Hey @MikelAlejoBR! This is going to be super helpful. Just left a few comments/suggestions, but feel free to reach out and I'll be happy to clarify/chat! 😄

rbac/management/utils.py
```diff
@@ -291,3 +292,55 @@ def get_admin_from_proxy(username, request):
...
+
+def validate_principal_type(queryParams):
+    """Validate the principal type.
+
+    Makes sure that the given "type" query parameter has a valid value, and returns it.
+
+    :param queryParams: The query parameters of the request.
+    :rtype: str
+    """
+    return validate_and_get_key(queryParams, "type", ["service-account", "user"], "user")
```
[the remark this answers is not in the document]
MikelAlejoBR: Perfect, yes! Initially I did it that way but I was unsure about repeating the query parameters in two places. Since we're changing those query parameters for the group view, this simply goes away. Changed!

rbac/management/group/view.py
```diff
@@ -541,6 +561,15 @@ def principals(self, request, uuid=None):
                 ),
             }
 
+            # Attempt validating and obtaining the "principal type" query
+            # parameter. For now, having "service-account" as the "principal
+            # type" returns a stub.
+            principalType = validate_principal_type(request.query_params)
```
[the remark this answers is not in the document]
MikelAlejoBR: Yes, changed!

rbac/management/group/view.py
```diff
@@ -428,6 +431,8 @@ def principals(self, request, uuid=None):
         @apiHeader {String} token User authorization token
 
         @apiParam (Path) {String} id Group unique identifier
+        @apiParam (Query) {String} type Parameter for specifying the type of principal to be returned
```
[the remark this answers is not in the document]
MikelAlejoBR: Yes, changed!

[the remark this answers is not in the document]
MikelAlejoBR: I updated the spec with the new query parameters.

rbac/management/group/view.py
```diff
@@ -510,13 +521,22 @@ def principals(self, request, uuid=None):
             serializer = GroupPrincipalInputSerializer(data=request.data)
             if serializer.is_valid(raise_exception=True):
                 principals = serializer.data.pop("principals")
```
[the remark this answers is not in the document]
MikelAlejoBR: I have changed the stubbed responses to be simply an empty array. However, I am unsure if I should keep the part of the code which simply echoes the `service-account`s when `POST`ing them to a group?

rbac/management/group/view.py
```diff
@@ -553,12 +582,30 @@ def principals(self, request, uuid=None):
...
                 message = "Query parameter {} is required.".format(USERNAMES_KEY)
                 raise serializers.ValidationError({key: _(message)})
-            username = request.query_params.get(USERNAMES_KEY, "")
+
+            # If the old query parameter is set, then we prioritize that one.
+            if usernames is not None:
+                username = usernames
+            else:
+                username = users
+
```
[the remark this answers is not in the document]
MikelAlejoBR: All right, simplified!

rbac/management/principal/view.py
MikelAlejoBR: Yes, changed!

rbac/management/group/view.py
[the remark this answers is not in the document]
coderbydesign: I believe the goal of this is purely to support the `GET`, and I'd suggest leaving anything for the `POST` and `DELETE` to when that's actually implemented.

rbac/management/principal/view.py
coderbydesign: Awesome!!

rbac/management/group/view.py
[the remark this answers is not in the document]
coderbydesign: Perfect! 😄

rbac/management/utils.py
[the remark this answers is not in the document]
coderbydesign: That makes sense, thanks! 😄

rbac/management/group/view.py
```diff
@@ -541,6 +562,18 @@ def principals(self, request, uuid=None):
...
             }
 
+            # Attempt validating and obtaining the "principal type" query
+            # parameter. For now, having "service-account" as the "principal
+            # type" returns a stub.
+            principalType = validate_and_get_key(
+                request.query_params, PRINCIPAL_TYPE_KEY, VALID_PRINCIPAL_TYPE_VALUE, required=False
+            )
+
+            if principalType:
```
coderbydesign, line 572: For these checks, do we want:
```suggestion
            if principalType == "service-account": 
```
so that we still return user-based principals either when it's not supplied _or_ `?principal_type=user`, since we do actually have the user data? Otherwise it might be unexpected if the UI starts getting empty arrays back when pulling existing user data (assuming they use the new query params for everything).

rbac/management/principal/view.py
```diff
@@ -111,7 +115,16 @@ def get(self, request):
...
             previous_offset = offset - limit
 
-        resp, usernames_filter = self.users_from_proxy(user, query_params, options, limit, offset)
+        # Attempt validating and obtaining the "prinicpal type" query
+        # parameter. For now, specifying the query parameter results
+        # in returning the mocked response.
+        principalType = validate_and_get_key(
+            query_params, PRINCIPAL_TYPE_KEY, VALID_PRINCIPAL_TYPE_VALUE, required=False
+        )
+        if principalType:
```
coderbydesign, line 124: Same with this one?
```suggestion
        if principalType == "service-account":
```

docs/source/specs/openapi.json
```diff
@@ -1305,6 +1331,24 @@
             "schema": {
               "type": "string"
             }
+          },
```
coderbydesign, line 1334: Do we want these to be added now? That would generate clients with this capability, which doesn't actually exist (yet).

rbac/management/group/view.py
```diff
@@ -510,13 +522,22 @@ def principals(self, request, uuid=None):
...
                 resp = self.add_principals(group, principals, org_id=org_id)
             else:
                 resp = self.add_principals(group, principals, account=account)
             if isinstance(resp, dict) and "errors" in resp:
                 return Response(status=resp["status_code"], data=resp["errors"])
             output = GroupSerializer(resp)
+
+            # Append the service accounts back to the payload.
+            output.appendServiceAccounts(serviceAccounts)
+
```
coderbydesign, line 540: Is this able to go away from this PR now that we're mocking based on the query param alone on the `GET`?

docs/source/specs/openapi.json
```diff
@@ -1305,6 +1331,24 @@
...
+            }
+          },
+          {
+            "in": "query",
+            "name": "service-accounts",
+            "required": false,
+            "description": "A comma separated list of service accounts to remove from the group.",
+            "schema": {
+              "type": "string"
+            }
```
coderbydesign, line 1351: Do we want these (the `DELETE` spec updates) to be added now? That would generate clients with this capability, which doesn't actually exist (yet).

rbac/management/group/view.py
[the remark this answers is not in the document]
MikelAlejoBR: Ooooh, sorry about that. The linked Jira ticket has in turn a linked GDOC which listed all these endpoints I've modified, and that's why I also implemented the `POST` and `DELETE` ones...

rbac/management/group/view.py
MikelAlejoBR: Sure. Again sorry, I misunderstood your previous comment. I understood that we had to return the stubbed response regardless of what query param we had specified.

rbac/management/principal/view.py
MikelAlejoBR: Changed!

rbac/management/group/view.py
MikelAlejoBR: Removed!

docs/source/specs/openapi.json
MikelAlejoBR: Yes, removed!

rbac/management/group/view.py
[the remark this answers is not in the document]
coderbydesign: Makes sense, no worries! 😄

rbac/management/group/view.py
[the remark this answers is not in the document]
coderbydesign: Awesome!

rbac/management/principal/view.py
coderbydesign: 🎉

MikelAlejoBR:
Hey,

I don't know if there is anything else to do here! The only thing that I would need help with is actually making the "group princpals" tests work, and after that we should be golden. I've tried adding/removing annotations, but that doesn't seem to work :/

If you could help me identifying what is missing, I'd greatly appreciate it!

Thank you in advance!

tests/management/group/test_view.py
coderbydesign: Hey @MikelAlejoBR! It's because you've added these tests under the `GroupViewNonAdminTests`, which has different test setup from `GroupViewsetTests`. If you move the tests up into the `GroupViewsetTests` class, you shouldn't get 403s.

MikelAlejoBR: Perfect, thank you for the suggestions! I was able to make it work :tada:

coderbydesign (approved):
The scope of this makes sense, and it all works as expected locally! Thanks, @MikelAlejoBR! 🚀

MikelAlejoBR:
Perfect, what's the

abaiken (approved):
LGTM! Thank you Mikel!

PR #922, opened 2023-10-03
[RHCLOUD-27277] Change type of value in ResourceDefinitionFilter component of openapi spec

lpichler:
Change is from type `string` to one of:
`string`
`null`
**array**: elements are one of:
`string`
`null`

### Links
https://issues.redhat.com/browse/RHCLOUD-27277

### DIFF
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -3725,8 +3725,32 @@
             ]
           },
           "value": {
-            "type": "string",
-            "example": "123456"
+            "oneOf": [
+              {
+                "type": "string",
+                "example": "123456"
+              },
+              {
+                "type": "null",
+                "example": null
+              },
+              {
+                "type": "array",
+                "items": {
+                  "oneOf": [
+                    {
+                      "type": "string",
+                      "example": "value1"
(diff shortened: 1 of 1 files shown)


romanblanco (approved):

vkrizan:
@lpichler can you please also change your documentation? Thank you!

romanblanco:
@vkrizan, you mean at https://developers.redhat.com/api-catalog/?

vkrizan:
I've meant this one: https://github.com/RedHatInsights/insights-rbac/blob/master/docs/source/management/role.rst#resource-definitions
Libor might know more...

abaiken (commented):
This makes sense to me. I think that the `null` type is not supported in openapi 3.0.0 - when I load this file into swagger, I get errors, we might have to do `nullable=True` instead

romanblanco (approved):
Tested by regenerating a new client in https://github.com/RedHatInsights/insights-rbac-api-client-ruby/pull/15

lpichler:
> @lpichler can you please also change your documentation? Thank you!

I will cover this in follow-up PR.

PR #931, opened 2023-10-12
refactor: removed not used method _create_customer()

petracihalova:
this PR removes method `_create_customer()` from `IdentityRequest` class from file `/insights-rbac/tests/identity_request.py`

I noticed that another method `_create_request_context()` in the same class uses this method like this:
```
        if create_customer:
            cls.customer = cls._create_customer(account, create_tenant=create_tenant, org_id=org_id)
```
but `cls.customer` is not used anymore

and the method `_create_customer()` itself is doing something only when param `create_tenant=True`, but default value is `False` + there is not a single occurrence of this function being used with parameter `create_tenant=True`

## Local Testing
just run tests via tox
`$ tox -r`


Ellen-Yi-Dong (approved):
LGTM!!

abaiken (approved):
LGTM! Nice catch Petra

PR #940, opened 2023-10-16
Remove unnecessary testdata

petracihalova:
this PR can me merged after https://github.com/RedHatInsights/insights-rbac/pull/938

## Description of Intent of Change(s):
in the class `RoleViewsetTests` I found we create set of test data around tenant `self.test_tenant`
for this tenant we create principal `self.test_principal`, header `self.test_header`, groups `self.test_group` and `self.test_group_two` and roles, policies and so on (everything with pattern `self.test_XXX`)

test data is created in the `setUp()` method and is used ONLY for 1 test (from 65 in this class) => `test_list_role_with_additional_fields_username_success()` 

so I changed this test to use the already existing test data (used by rest of tests) and removed the not needed tests data

bonus: faster tests (on my machine 5 - 6 seconds)

## Local testing:
- without the change remove all test data and run tests via `tox` and check that only 1 test is affected
- with the change just run tests via `tox`

### DIFF
--- a/tests/management/role/test_view.py
+++ b/tests/management/role/test_view.py
@@ -831,23 +831,7 @@ class RoleViewsetTests(IdentityRequest):
 
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
 
-    @patch(
-        "management.principal.proxy.PrincipalProxy.request_filtered_principals",
-        return_value={
-            "status_code": 200,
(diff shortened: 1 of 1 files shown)


ddonahue007 (approved):
lgtm

Ellen-Yi-Dong (approved):
LGTM!

PR #941, opened 2023-10-16
Fix for test_list_role_with_groups_in_fields_for_principal_scope_success()

petracihalova:
this PR should be merged after https://github.com/RedHatInsights/insights-rbac/pull/940

parentheses were missing in `test_list_role_with_groups_in_fields_for_principal_scope_success()` to construct the url variable - so the url didn't contain username
```
/api/v1/roles/?add_fields=groups_in_count,groups_in&username=
```
in this case the API processed the request as if there was no username param like this
```
/api/v1/roles/?add_fields=groups_in_count,groups_in
```


ddonahue007 (approved):
lgtm

Ellen-Yi-Dong (approved):
LGTM!

PR #942, opened 2023-10-16
[RHCLOUD-18213] groups_in logic

petracihalova:
this PR can me merged after https://github.com/RedHatInsights/insights-rbac/pull/941

## Link(s) to Jira
- [RHCLOUD-18213](https://issues.redhat.com/browse/RHCLOUD-18213)

BEFORE:
when you call the `/roles` endpoint with `groups_in` and/or `groups_in_count` params and with `username` param then for each role you get a list/count of the groups the role is in and these groups are including `Default access group` (or `Custom default access` group) and `Default admin access` group

but the `Default admin access` group should be shown only when principal is org admin

NOW:
with this change the `Default admin access` group is shown only when principal is org admin

## Local Testing
you need to add/create a non org admin principal under your tenant and 
you need a role that is in both system groups (`Default access group` and `Default admin access`)
then call the endpoint 
```
/api/v1/roles/?add_fields=groups_in_count,groups_in&username=<your_username>
```
and check that under the role you can see only `Default access group`

### DIFF
--- a/rbac/management/querysets.py
+++ b/rbac/management/querysets.py
@@ -37,7 +37,7 @@ from management.utils import (
 )
 from rest_framework import permissions, serializers
 
-from api.models import Tenant
+from api.models import Tenant, User
 from rbac.env import ENVIRONMENT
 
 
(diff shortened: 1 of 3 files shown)


Ellen-Yi-Dong (approved):
LGTM!

Ellen-Yi-Dong (approved):
LGTM!

dagbay-rh (approved):

PR #959, opened 2023-11-08
RHCLOUD-27878 | fix: unintended "principal_username" filter applied to query

MikelAlejoBR:
## Link(s) to Jira
- [[RHCLOUD]](https://issues.redhat.com/browse/RHCLOUD-27878)

Sometimes the UI seems to send "false" when the principal username filter doesn't want to be applied, which makes the query to actually try to filter by the "false" username.

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/management/group/view.py
+++ b/rbac/management/group/view.py
@@ -761,8 +761,12 @@ class GroupViewSet(
                         },
                     )
 
-                # Get the principal username option parameter and the limit and offset parameters too.
-                options[PRINCIPAL_USERNAME_KEY] = request.query_params.get(PRINCIPAL_USERNAME_KEY)
+                # Get the principal username option parameter and the limit and offset parameters too. Sometimes the UI
+                # sends "false" as the "principal username" query parameter. In that case we assume no filtering was
+                # intended to be done.
+                principal_username = request.query_params.get(PRINCIPAL_USERNAME_KEY)
+                if principal_username and principal_username != "false":
+                    options[PRINCIPAL_USERNAME_KEY] = principal_username
                 options["limit"] = int(request.query_params.get("limit", StandardResultsSetPagination.default_limit))
                 options["offset"] = int(request.query_params.get("offset", 0))
 


rbac/management/group/view.py
```diff
@@ -761,10 +761,14 @@ def principals(self, request, uuid=None):
...
-                options["limit"] = int(request.query_params.get("limit", StandardResultsSetPagination.default_limit))
-                options["offset"] = int(request.query_params.get("offset", 0))
+                # Get the principal username option parameter and the limit and offset parameters too. Sometimes the UI
+                # sends "false" as the "principal username" query parameter. In that case we assume no filtering was
+                # intended to be done.
+                principal_username = request.query_params.get(PRINCIPAL_USERNAME_KEY)
+                if principal_username and principal_username != "false":
+                    options[PRINCIPAL_USERNAME_KEY] = principal_username
+                options["limit"] = request.query_params.get("limit", StandardResultsSetPagination.default_limit)
+                options["offset"] = request.query_params.get("offset", 0)
```
petracihalova, line 771: you change back what was changed in the previous PR? https://github.com/RedHatInsights/insights-rbac/pull/958

MikelAlejoBR: Great catch, Petra. Many thanks!

Ellen-Yi-Dong (approved):
LGTM!

rbac/management/group/view.py
```diff
@@ -761,8 +761,12 @@ def principals(self, request, uuid=None):
                         },
                     )
 
-                # Get the principal username option parameter and the limit and offset parameters too.
-                options[PRINCIPAL_USERNAME_KEY] = request.query_params.get(PRINCIPAL_USERNAME_KEY)
+                # Get the principal username option parameter and the limit and offset parameters too. Sometimes the UI
+                # sends "false" as the "principal username" query parameter. In that case we assume no filtering was
```
petracihalova, line 765: well with this I am not sure and 😇 I'll wait for the arguments of someone more experienced because the
`principal_username` has this definition in open api spec "Parameter for filtering group principals by principal username using string contains search." ... so it seems to me that the fix should be done on the UI side = not to send request with this param

coderbydesign: Yeah without the full context on why this is happening, I'd agree with that. Given the spec, it should be a consumer of the API's responsibility to send a meaningful query param, otherwise I feel like we'd start coupling edge-cases to the logic?

MikelAlejoBR: Yeah, I think you're both right... I thought about the fact that this might not be our responsibility to fix, but like I wanted to badly to fix this and try the stuff and make it work that I clearly made the wrong decision. I am going to talk to the UI team and let her now.

Thanks for the feedback!

PR #964, opened 2023-11-10
Revert "Create new Grafana Dashboard for RBAC Worker Alerts "

Ellen-Yi-Dong:
Reverts RedHatInsights/insights-rbac#961


Ellen-Yi-Dong:
I need to change the datasource for when the cache is disabled so reverting my previous PR in order to get the correct one in

petracihalova (approved):

PR #974, opened 2023-11-23
[RHCLOUD-29513] open api spec - added base urls for prod/stage/local

petracihalova:
The Open API spec for RBAC in the API Catalog and Documentation Red Hat Developer pages ([link](https://developers.redhat.com/api-catalog/api/rbac)) doesn't contain full url for production, stage and local environment. This causes the url in the black panel is not correct.

image

This PR fixes it.

### DIFF
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -2848,9 +2848,36 @@
   },
   "servers": [
     {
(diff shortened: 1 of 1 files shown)


lpichler (approved):

mfindra:
QE pass

PR #992, opened 2024-01-11
some rbac cleanup

dagbay-rh:
## Link(s) to Jira
- n/a

What:
I went thru our various docker files and removed all of the following for each respective reason:
* `.travis.yml` - we no longer use travis for pr checks, we use jenkins
* `Dockerfile-env` - not used anywhere I could find here or in app interface. It appeared really old, building on centos
* `Dockerfile-test` - only used by redundant testing method in our makefile. We can run unit tests locally without it
* `Dockerfile.clowder` - not used anywhere I could find here or in app interface. Our clowder deployment template does not use it 
* `Jenkinsfile-smoke.groovy` - not used anywhere I could find here or in app interface. As far as I can tell our smoke tests do not use this
* `full_tests.sh` - redundant, it does the same thing as `pr_check.sh` but its older and not up to date. Also not used anywhere afaik
* `rbac-test.yml` - a docker compose file used to spin up the server and run unit tests. Seems redundant to me since we can run the tests in a container via `pr_check.sh` 

Why:
I found there to be a lot of noise in the repo, and want to try and help tidy things up a bit. 5 dockerfiles is wild 

Happy to revert things people object to removing


petracihalova (approved):
👍 awesome

abaiken (approved):

PR #993, opened 2024-01-12
RHCLOUD-30375 | feature: filter service accounts by their description and name

MikelAlejoBR:
## Link(s) to Jira
- [[RHCLOUD-30375]](https://issues.redhat.com/browse/RHCLOUD-30375)

In order to provide a better user experience, we need to be able to return filtered results for service accounts by filtering them by their descriptions and/or their names.

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/management/principal/it_service.py
+++ b/rbac/management/principal/it_service.py
@@ -53,6 +53,11 @@ it_request_error = Counter(
     ["error"],
 )
 
+# Keys for the "options" dictionary. The "options" dictionary represents the query parameters passed by the calling
+# client.
+SERVICE_ACCOUNT_DESCRIPTION_KEY = "service_account_description"
+SERVICE_ACCOUNT_NAME_KEY = "service_account_name"
+
 
 def limit_offset_validation(offset, limit):
     """Limit and offset should not be negative number."""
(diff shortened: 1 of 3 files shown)


rbac/management/principal/it_service.py
```diff
@@ -275,7 +280,36 @@ def get_service_accounts_group(self, group: Group, bearer_token: str, options: d
...
+            for sa in service_accounts:
+                # Initialize both matches as "True", because having them at "False" would force us to make them "True"
+                # whenever one of the filters is not specified, since otherwise not specifying a filter would stop the
+                # service account to be included in the resulting list.
+                sa_description_filter_matches: bool = True
+                sa_name_filter_matches: bool = True
+
+                if sa_description_filter:
+                    sa_description_filter_matches = ("description" in sa) and (
+                        sa_description_filter in sa["description"]
```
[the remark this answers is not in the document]
MikelAlejoBR: I have echoed the question in the ticket and mentioned you too, to see what the rest think about it, because it is a interesting point. Thanks for the review!

rbac/management/principal/it_service.py
```diff
@@ -275,7 +280,36 @@ def get_service_accounts_group(self, group: Group, bearer_token: str, options: d
             service_account_principals=sap_dict, it_service_accounts=it_service_accounts, options=options
         )
 
-        return service_accounts
+        # If either the description or name filters were specified, we need to only keep the service accounts that
```
petracihalova, line 283: this filtering is for `GET groups/uuid/principals/` endpoint ... is it expected the same for `GET principals/?`

rbac/management/principal/it_service.py
```diff
@@ -275,7 +280,36 @@ def get_service_accounts_group(self, group: Group, bearer_token: str, options: d
...
+
+        if sa_description_filter or sa_name_filter:
+            filtered_service_accounts: list[dict] = []
+
+            for sa in service_accounts:
+                # Initialize both matches as "True", because having them at "False" would force us to make them "True"
+                # whenever one of the filters is not specified, since otherwise not specifying a filter would stop the
+                # service account to be included in the resulting list.
+                sa_description_filter_matches: bool = True
+                sa_name_filter_matches: bool = True
```
petracihalova, line 296: I can see that you use the type annotations here but the rbac project doesnt use them otherwise, so it is maybe question to whole team ... in any case, the annotations aren't checked at the moment in this repo

rbac/management/principal/it_service.py
petracihalova: yes I can see now in the ticket that the answer is NO, so for the record this filtering is expected only for groups endpoint

rbac/management/principal/it_service.py
MikelAlejoBR: The reason why I use type annotations is because I'll get nice suggestions from VSCode while developing, and I figured that the rest of the team would also like having those suggestions too. However in this case it might be a bit unnecessary since the type can be easily inferred? What do you think?

petracihalova: I dont say that I want to remove it, to be honest I dont know ... because it it not only one occurrence of this so removing doesnt solve it ... these 2 annotations are correct but there are annotations added by previous PRs that are not ... so what I am saying is that in python if we start to use annotations, we have to check them by some tool and we are not doing that yet ... from this PR perspective I believe we can let it as it is, this is probably topic for whole team discussion

MikelAlejoBR: Oops, sorry! I did not want to stir things up :joy: I thought it'd be a helpful thing, but please let me know what's the outcome of that discussion —or if I could be there too to listen to opinions that'd be great!— so that I can stop/keep type hinting things :smile: :+1:

petracihalova: that's fine, it just causes type annotation checker like mypy finds some output, it has no other effect

petracihalova (commented):
I have same question as Libor about case sensitivity and for rest LGTM 👍

rbac/management/principal/it_service.py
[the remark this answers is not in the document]
MikelAlejoBR: I am going to resolve this conversation since it seems that we are going to make it case sensitive, as in the PR.

petracihalova (approved):

abaiken (approved):

PR #994, opened 2024-01-15
RHCLOUD-30035 | feature: tests for the "get_principal" method

MikelAlejoBR:
## Link(s) to Jira
- [[RHCLOUD-30035]](https://issues.redhat.com/browse/RHCLOUD-30035)

Adds tests for the "get_principal" method, to make sure that it creates the user and service account principals when they're not present in the database.

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/management/utils.py
+++ b/rbac/management/utils.py
@@ -84,11 +84,11 @@ def get_principal(username, request, verify_principal=True, from_query=False):
             principal = Principal.objects.get_or_create(
                 username=user.username, tenant=tenant, type=SERVICE_ACCOUNT_KEY, service_account_id=user.client_id
             )
-
-        # Avoid possible race condition if the user was created while checking BOP
-        principal, created = Principal.objects.get_or_create(
-            username=username, tenant=tenant
-        )  # pylint: disable=unused-variable
+        else:
+            # Avoid possible race condition if the user was created while checking BOP
+            principal = Principal.objects.get_or_create(
+                username=username, tenant=tenant
+            )  # pylint: disable=unused-variable
 
     return principal
 
--- a/rbac/management/utils.py
+++ b/rbac/management/utils.py
@@ -81,12 +81,12 @@ def get_principal(username, request, verify_principal=True, from_query=False):
         # In the case that the user that made the request was a service account, store it accordingly in the database.
         user: User = request.user
         if user and user.is_service_account:
-            principal = Principal.objects.get_or_create(
+            principal, _ = Principal.objects.get_or_create(
                 username=user.username, tenant=tenant, type=SERVICE_ACCOUNT_KEY, service_account_id=user.client_id
             )
         else:
             # Avoid possible race condition if the user was created while checking BOP
-            principal = Principal.objects.get_or_create(
+            principal, _ = Principal.objects.get_or_create(
                 username=username, tenant=tenant
             )  # pylint: disable=unused-variable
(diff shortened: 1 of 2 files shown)


tests/management/test_utils.py
```diff
@@ -130,3 +134,48 @@ def test_account_number_from_tenant_name(self):
         """Test that we get the expected account number from a tenant name."""
         tenant = Tenant.objects.create(tenant_name="acct1234")
         self.assertEqual(account_id_for_tenant(tenant), "1234")
+
+    @mock.patch("management.utils.verify_principal_with_proxy")
+    def test_get_principal_created(self, mocked):
+        """Test that when a user principal does not exist in the database, it gets created."""
+        # Build a non existent service account.
```
petracihalova, line 141: `# Build a non existent user-based principal.`

rbac/management/utils.py
```diff
@@ -84,11 +84,11 @@ def get_principal(username, request, verify_principal=True, from_query=False):
             principal = Principal.objects.get_or_create(
                 username=user.username, tenant=tenant, type=SERVICE_ACCOUNT_KEY, service_account_id=user.client_id
             )
-
-        # Avoid possible race condition if the user was created while checking BOP
-        principal, created = Principal.objects.get_or_create(
-            username=username, tenant=tenant
-        )  # pylint: disable=unused-variable
+        else:
```
petracihalova, line 87: tests are failing because this piece of code should be outside the else block + tuple of 2 values is returned from `Principal.objects.get_or_create()`, the principal object and then boolean (`True` if object was created, `False` if object was already in db) so we need 2 variables on the left side, in the python you can use `_` as variable name and then this variable is skipped by linters

```
        user: User = request.user
        if user and user.is_service_account:
            Principal.objects.get_or_create(
                username=user.username, tenant=tenant, type=SERVICE_ACCOUNT_KEY, service_account_id=user.client_id
            )
        # Avoid possible race condition if the user was created while checking BOP
        principal, _ = Principal.objects.get_or_create(username=username, tenant=tenant)
```

petracihalova (changes requested):

tests/management/test_utils.py
MikelAlejoBR: Fixed, thanks!

rbac/management/utils.py
MikelAlejoBR: Thanks for the review!

> tests are failing because this piece of code should be outside the else block

Actually I think the code is properly indented, but I wanted to avoid running two very similar queries for the same outcome and the same purpose. I would vote for keeping it indented inside the `else:` block, what do you think?

> tuple of 2 values is returned from Principal.objects.get_or_create(), the principal object and then boolean (True if object was created, False if object was already in db) so we need 2 variables on the left side, in the python you can use _ as variable name and then this variable is skipped by linters

You are right, thanks so much! I've added those variable placeholders and now tests pass! :smile:

petracihalova: yes you are right with the "else" block 👍 now I can see it

petracihalova (approved):

PR #997, opened 2024-01-17
 RHCLOUD-30043 | fix: the access endpoint does not work for checking principals' permissions

MikelAlejoBR:
## Link(s) to Jira
- [[RHCLOUD-30043]](https://issues.redhat.com/browse/RHCLOUD-30043)

The "access" endpoint should return the permissions for a user for a given application when the user or service account that is requesting them has the proper permissions.

However, since the "access" endpoint receives so much traffic, loading permissions for every user that sent a request to the endpoint was becoming much of a burden. Therefore, loading the users' permissions was restricted to a few particular cases.

These changes should allow to satisfy the use case we are after without sacrificing the restrictions that were set for the endpoint.

### How can the bug be exploited and fix confirmed?
1. Create a new service account in stage.
2. Add it to a new group.
3. Add the "User Administrator" permission to the group, which should allow the service account to query for other users' permissions for particular applications.
4. Send a request to the `/access/` endpoint with the `?application=rbac&username=<another-user>` query parameters.

The endpoint should return the permissions for that user since the service account has the proper permissions, but it will not.

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/rbac/middleware.py
+++ b/rbac/rbac/middleware.py
@@ -22,8 +22,9 @@ import logging
 from json.decoder import JSONDecodeError
 
 from django.conf import settings
+from django.core.handlers.wsgi import WSGIRequest
 from django.db import IntegrityError
-from django.http import Http404, HttpResponse
+from django.http import Http404, HttpResponse, QueryDict
 from django.urls import resolve
 from django.utils.deprecation import MiddlewareMixin
 from management.cache import TenantCache
@@ -259,7 +260,7 @@ class IdentityHeaderMiddleware(MiddlewareMixin):
                 return HttpResponse(json.dumps(payload), content_type="application/json", status=400)
 
             if settings.AUTHENTICATE_WITH_ORG_ID:
-                if not user.admin and not (request.path.endswith("/access/") and request.method == "GET"):
+                if self.is_allowed_load_user_permissions(request, user):
                     try:
                         tenant = Tenant.objects.filter(org_id=user.org_id).get()
                     except Tenant.DoesNotExist:
@@ -268,7 +269,7 @@ class IdentityHeaderMiddleware(MiddlewareMixin):
 
                     user.access = IdentityHeaderMiddleware._get_access_for_user(user.username, tenant)
             else:
-                if not user.admin and not (request.path.endswith("/access/") and request.method == "GET"):
+                if self.is_allowed_load_user_permissions(request, user):
                     try:
                         tenant_name = create_tenant_name(user.account)
                         tenant = Tenant.objects.filter(tenant_name=tenant_name).get()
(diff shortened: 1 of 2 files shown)


astrozzc (approved):
LGTM! The current usage of access endpoint does not have username. So it won't impact current usage.
Maybe the function name could be should_load_user_permissions or something to be more accurate.

MikelAlejoBR:
> LGTM! The current usage of access endpoint does not have username. So it won't impact current usage. Maybe the function name could be should_load_user_permissions or something to be more accurate.

You are right, yes. That name makes more sense. Thanks for the review!

MikelAlejoBR:
Force pushed to fix a typo on a commit message.

rbac/rbac/middleware.py
```diff
@@ -416,6 +417,29 @@ def process_response(self, request, response):  # pylint: disable=no-self-use
...
+        #
+        # - RHCLOUD-15394
+        # - RHCLOUD-29631
+        #
+        # There is one use case where we need to load the user's permissions: whenever they want to query for their
+        # or other users' permissions. In that case, we need to know if they're allowed to do so, and for that, we
+        # need to preload their permissions to check them afterwards in the subsequent permission checkers.
+        if request.path.endswith("/access/") and request.method == "GET":
+            query_params: QueryDict = request.GET
+            return "application" in query_params and "username" in query_params
```
petracihalova, line 439: well it is really tiny comment
"application" is every time in the `query_params` because it is mandatory parameter (it can be empty like `application=` but has to be present in the query) 

so this part of condition is true every time and can be removed
`"application" in query_params`

if you insist that you want this condition there then consider the flip order of conditions into
`"username" in query_params and "application" in query_params`
because python uses lazy evaluation here, when first condition is False then python stop the evaluation (it doesnt matter if the second condition is True or False, the whole expression will be False)

rbac/rbac/middleware.py
```diff
@@ -416,6 +417,29 @@ def process_response(self, request, response):  # pylint: disable=no-self-use
...
+        else:
+            return True
```
petracihalova, line 441: the `else` block is not needed because the `if` block ends with `return`

tests/rbac/test_middleware.py
```diff
@@ -273,6 +274,74 @@ def test_tenant_process_without_org_id(self):
...
+        user.admin = False
+
+        request = Mock()
+        request.path = "/access/"
+        request.method = "GET"
+        request.GET = QueryDict("application=rbac&username=foo")
+        middleware = IdentityHeaderMiddleware(get_response=Mock())
+        self.assertEqual(middleware.should_load_user_permissions(request, user), True)
+
+    def test_should_load_user_permissions_regular_user_access_missing_query_params(self):
```
petracihalova, line 323: the `application` param is mandatory so it is enforced in the` rbac/rbac/midleware.py `

petracihalova:
@MikelAlejoBR let me know what you think, but overall I like it 👍

rbac/rbac/middleware.py
MikelAlejoBR: I'd prefer to leave the `else:` statement because I think having it improves readability:

```python
if request.path.endswith("/access/") and request.method == "GET":
    query_params: QueryDict = request.GET
    return "username" in query_params and "application" in query_params
else:
    return True
```

vs.

```python
if request.path.endswith("/access/") and request.method == "GET":
    query_params: QueryDict = request.GET
    return "username" in query_params and "application" in query_params

return True
```

tests/rbac/test_middleware.py
MikelAlejoBR: I don't follow on this one, what do you mean?

petracihalova: this was related to my first comment about the `application` param, you left the param in the condition so this comment is no longer valid

rbac/rbac/middleware.py
petracihalova: ok, for me it's the opposite, but that's probably subjective 😂😂😂

MikelAlejoBR:
Thank you for your review, @petracihalova! :tada:

PR #999, opened 2024-01-19
RHCLOUD-30299 | feature: return if service accounts are in group

MikelAlejoBR:
## Link(s) to Jira
- [[RHCLOUD-30299]](https://issues.redhat.com/browse/RHCLOUD-30299)

The UI needs a way to discern which service accounts are already present in the group, so that they can show them as either enabled or disabled when presenting them to the user.

### How can the feature be exercised?
1. Create a new group.
2. Add some service accounts to it —or not—.
3. Send a `GET /groups/{uuid}/principals/?service_account_client_ids=<clientID>,<clientID>,<clientID>` request.

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -1277,7 +1277,7 @@
             "in": "query",
             "name": "service_account_client_ids",
             "required": false,
-            "description": "By specifying a list of client IDs with this query parameter, RBAC will return an object with the specified client ID and it's matching boolean value to flag whether the client ID is present in the group or not",
+            "description": "By specifying a list of client IDs with this query parameter, RBAC will return an object with the specified client ID and it's matching boolean value to flag whether the client ID is present in the group or not. This query parameter cannot be used along with any other query parameter.",
             "schema": {
               "type": "array",
               "items": {
@@ -1316,6 +1316,9 @@
                     },
                     {
                       "$ref": "#/components/schemas/ServiceAccountPagination"
+                    },
+                    {
+                      "$ref": "#/components/schemas/ServiceAccountInGroupResponse"
                     }
                   ]
                 }
@@ -3380,6 +3383,30 @@
             }
           }
         ]
+      },
+      "ServiceAccountInGroupResponse": {
+        "properties": {
+          "meta": {
+            "$ref": "#/components/schemas/PaginationMeta"
+          },
+          "links": {
+            "description": "The links object for this particular response will be empty, since there is no pagination available for the query parameter",
+            "type": "object",
+            "example": {}
+          },
+          "data": {
+            "description": "Object which indicates whether the given service account UUIDs in the query parameter are present in the specified group or not",
+            "type": "object",
+            "additionalProperties": {
+              "description": "The response is a map of the form \"UUID\": (true|false)",
+              "type": "boolean"
+            },
+            "example": {
+              "dd946f24-cfda-11ee-acb6-7b2702ff4dc8": true,
+              "3e728bb0-b167-013c-c455-6aa2427b506c": false
+            }
+          }
+        }
       },
       "Group": {
         "required": [
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -1275,6 +1275,18 @@
           },
           {
             "in": "query",
+            "name": "service_account_client_ids",
+            "required": false,
+            "description": "By specifying a list of client IDs with this query parameter, RBAC will return an object with the specified client ID and it's matching boolean value to flag whether the client ID is present in the group or not",
+            "schema": {
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            }
+          },
+          {
+            "in": "query",
             "name": "service_account_description",
             "required": false,
             "description": "Parameter for filtering the service accounts by their description.",
--- a/rbac/management/group/view.py
+++ b/rbac/management/group/view.py
@@ -17,7 +17,8 @@
 
 """View for group management."""
 import logging
-from typing import Iterable
+from typing import Iterable, Optional
+from uuid import UUID
 
 import requests
 from django.conf import settings
@@ -70,6 +71,7 @@ PRINCIPAL_TYPE_KEY = "principal_type"
 PRINCIPAL_USERNAME_KEY = "principal_username"
 VALID_ROLE_ORDER_FIELDS = list(RoleViewSet.ordering_fields)
 ROLE_DISCRIMINATOR_KEY = "role_discriminator"
+SERVICE_ACCOUNT_CLIENT_IDS_KEY = "service_account_client_ids"
 SERVICE_ACCOUNT_DESCRIPTION_KEY = "service_account_description"
 SERVICE_ACCOUNT_NAME_KEY = "service_account_name"
 SERVICE_ACCOUNT_USERNAME_FORMAT = "service-account-{clientID}"
@@ -500,7 +502,7 @@ class GroupViewSet(
         return group
 
     @action(detail=True, methods=["get", "post", "delete"])
-    def principals(self, request: Request, uuid=None):
+    def principals(self, request: Request, uuid: Optional[UUID] = None):
         """Get, add or remove principals from a group."""
         """
         @api {get} /api/v1/groups/:uuid/principals/    Get principals for a group
@@ -713,6 +715,78 @@ class GroupViewSet(
                 serializer = ServiceAccountSerializer(page, many=True)
 
                 return self.get_paginated_response(serializer.data)
+
+            if SERVICE_ACCOUNT_CLIENT_IDS_KEY in request.query_params:
+                # This query parameter is incompatible with any other query parameter, because it changes the
+                # payload that gets returned.
+                if len(request.query_params) > 1:
+                    return Response(
+                        status=status.HTTP_400_BAD_REQUEST,
+                        data={
+                            "errors": [
+                                {
+                                    "detail": f"The '{SERVICE_ACCOUNT_CLIENT_IDS_KEY}' parameter is incompatible with"
+                                    f" any other query parameter. Please, use it alone",
+                                    "source": "groups",
+                                    "status": str(status.HTTP_400_BAD_REQUEST),
+                                }
+                            ]
+                        },
+                    )
+                else:
+                    # Check that the specified query parameter is not empty.
+                    service_account_client_ids_raw = request.query_params.get(SERVICE_ACCOUNT_CLIENT_IDS_KEY).strip()
+                    if not service_account_client_ids_raw:
+                        return Response(
+                            status=status.HTTP_400_BAD_REQUEST,
+                            data={
+                                "errors": [
+                                    {
+                                        "detail": "Not a single client ID was specified for the client IDs filter",
+                                        "source": "groups",
+                                        "status": str(status.HTTP_400_BAD_REQUEST),
+                                    }
+                                ]
+                            },
+                        )
+
+                    # Turn the received and comma separated client IDs into a manageable set.
+                    received_client_ids: set[str] = set(service_account_client_ids_raw.split(","))
+
+                    processed_client_ids: set[UUID] = set()
+                    for rci in received_client_ids:
+                        try:
+                            processed_client_ids.add(UUID(str(rci)))
+                        except ValueError:
+                            return Response(
+                                status=status.HTTP_400_BAD_REQUEST,
+                                data={
+                                    "errors": [
+                                        {
+                                            "detail": f"The specified client ID '{rci}' is not a valid UUID",
+                                            "source": "groups",
+                                            "status": str(status.HTTP_400_BAD_REQUEST),
+                                        }
+                                    ]
+                                },
+                            )
+
+                    # Generate the report of which of the tenant's service accounts are in a group, and which
+                    # ones are available to be added to the given group.
+                    it_service = ITService()
+                    result: dict = it_service.generate_service_accounts_report_in_group(
+                        group=group, client_ids=processed_client_ids
+                    )
+
+                    # Prettify the output payload and return it.
+                    return Response(
+                        status=status.HTTP_200_OK,
+                        data={
+                            "meta": {"count": len(result)},
+                            "links": {},
+                            "data": result,
+                        },
+                    )
 
             principals_from_params = self.filtered_principals(group, request)
             page = self.paginate_queryset(principals_from_params)
--- a/rbac/management/group/view.py
+++ b/rbac/management/group/view.py
@@ -655,6 +655,79 @@ class GroupViewSet(
             # ... and return it.
             response = Response(status=status.HTTP_200_OK, data=output.data)
         elif request.method == "GET":
+            # Check if the request comes with a bunch of service account client IDs that we need to check. Since this
+            # query parameter is incompatible with any other query parameter, we make the checks first. That way if any
+            # other query parameter was specified, we simply return early.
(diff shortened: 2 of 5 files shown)


petracihalova:
I have a group with 1 SA `0bba10f8-7ae0-4148-a29f-592c9992caa0` and I tried to send this query without any clientID (tested localy) 
`http://localhost:8000/api/rbac/v1/groups/b9bd9669-8e3e-485e-b504-ca029ce3fcf4/principals/?service_account_client_ids=`

and I get
```
{
    "meta": {
        "count": 2
    },
    "links": {},
    "data": {
        "": false,
        "0bba10f8-7ae0-4148-a29f-592c9992caa0": false
    }
}
```
please notice that response contains one empty row and one row for `0bba10f8-7ae0-4148-a29f-592c9992caa0` (this SA belongs in my group)
so for
`http://localhost:8000/api/rbac/v1/groups/b9bd9669-8e3e-485e-b504-ca029ce3fcf4/principals/?principal_type=service-account`
I get
```
{
    "meta": {
        "count": 1,
        "limit": 10,
        "offset": 0
    },
    "links": {
        "first": "/api/rbac/v1/groups/b9bd9669-8e3e-485e-b504-ca029ce3fcf4/principals/?limit=10&offset=0&principal_type=service-account",
        "next": null,
        "previous": null,
        "last": "/api/rbac/v1/groups/b9bd9669-8e3e-485e-b504-ca029ce3fcf4/principals/?limit=10&offset=0&principal_type=service-account"
    },
    "data": [
        {
            "clientID": "0bba10f8-7ae0-4148-a29f-592c9992caa0",
            "name": "0bba10f8-7ae0-4148-a29f-592c9992caa0-name",
            "description": "0bba10f8-7ae0-4148-a29f-592c9992caa0-description",
            "owner": "service-account-0bba10f8-7ae0-4148-a29f-592c9992caa0",
            "time_created": 1707500727,
            "type": "service-account",
            "username": "service-account-0bba10f8-7ae0-4148-a29f-592c9992caa0"
        }
    ]
}
```
I believe this lines in `it_service.py` should be changed
```
        for gsap in group_service_account_principals:
            # The "result" dictionary already contains the client IDs specified by the client. If the group's service
            # accounts are present in the "result" dictionary, we need to flag it as such.
            result[gsap] = gsap in result
```
e.g. into something like this
```
        for gsap in group_service_account_principals:
            # The "result" dictionary already contains the client IDs specified by the client. If the group's service
            # accounts are present in the "result" dictionary, we need to flag it as such.
            if gsap in result:
                result[gsap] = True
```
and it should be added the check if the query param contains a value


then I tried to use `principal_type` and `service_account_client_ids` in one request
`http://localhost:8000/api/rbac/v1/groups/b9bd9669-8e3e-485e-b504-ca029ce3fcf4/principals/?principal_type=service-account&service_account_client_ids=0bba10f8-7ae0-4148-a29f-592c9992caa0`
and I get same response as is only with `principal_type` even I can see in the code that for other combinations the error 400 is expected

MikelAlejoBR:
Thank you for your review, @petracihalova! I've pushed some changes and also some tests to make sure that the handler returns proper error messages to the client when invalid input has been given.

rbac/management/group/view.py
```diff
@@ -502,7 +505,7 @@ def remove_principals(self, group, principals, account=None, org_id=None):
         return group
 
     @action(detail=True, methods=["get", "post", "delete"])
-    def principals(self, request, uuid=None):
+    def principals(self, request: Request, uuid: UUID = None):
```
petracihalova, line 508: in this case you should use `Optional` from typing `package`
`def principals(self, request: Request, uuid: Optional[UUID] = None):`

petracihalova:
when I use the `service_account_client_ids` with a SA that not belongs into my group and in the same time there is another SA that belongs in this group, the response contains data about both? that is expected?
```
/groups/{{group-uuid}}/principals/?service_account_client_ids=18aaa3d0-9d8f-013c-1dc8-6aa2427b506a

{
    "meta": {
        "count": 2
    },
    "links": {},
    "data": {
        "0bba10f8-7ae0-4148-a29f-592c9992caa0": true,
        "18aaa3d0-9d8f-013c-1dc8-6aa2427b506a": false
    }
}
```

rbac/management/group/view.py
MikelAlejoBR: Refactored, thanks!

MikelAlejoBR:
> when I use the `service_account_client_ids` with a SA that not belongs into my group and in the same time there is another SA that belongs in this group, the response contains data about both? that is expected?
> 
> ```
> /groups/{{group-uuid}}/principals/?service_account_client_ids=18aaa3d0-9d8f-013c-1dc8-6aa2427b506a
> 
> {
>     "meta": {
>         "count": 2
>     },
>     "links": {},
>     "data": {
>         "0bba10f8-7ae0-4148-a29f-592c9992caa0": true,
>         "18aaa3d0-9d8f-013c-1dc8-6aa2427b506a": false
>     }
> }
> ```

Yeah, for some reason I interpreted it as if we had to return all the service accounts present in the group + the ones specified by the client, but thinking about it it is true that if they ask us about SA "A", we should only respond to that one. So I've refactored to just return the information about the ones that we've been asked about.

Thanks!

petracihalova:
hey Mikel I have one last question, what is expected when you use `principal_type` and `service_account_client_ids param` in the same time? because now the `service_account_client_ids`param is ignored but when I use combination of e.g. `service_account_description` and `service_account_client_ids` the 400 is returned

tests/management/group/test_view.py
```diff
@@ -2129,6 +2131,101 @@ def test_get_group_service_account_invalid_limit_offset(self, mock_request):
             self.assertEqual(int(response.data.get("meta").get("count")), 3)
             self.assertEqual(len(response.data.get("data")), 3)
 
+    def test_get_group_principals_check_service_account_ids_incompatible_query_parameters(self):
```
petracihalova, line 2134: I like all these tests and it would be nice to have some positive test too here

tests/management/principal/test_it_service.py
```diff
@@ -0,0 +1,334 @@
...
+            service_account_id=uuid.uuid4(),
+            type="service-account",
+            tenant=self.tenant,
+        )
+
+        # Create a set with the service accounts that will go in the group. It will make it easier to make assertions
+        # below.
+        group_service_accounts_set = {str(sa_1.service_account_id), str(sa_2.service_account_id)}
+
+        # For this test, we don't associate any service accounts to it.
```
petracihalova, line 157: not sure if I understand this comment? below some SA are added into group

tests/management/principal/test_it_service.py
```diff
@@ -0,0 +1,334 @@
...
+        # below.
+        group_service_accounts_set = {
+            str(sa_1.service_account_id),
+            str(sa_2.service_account_id),
+            str(sa_3.service_account_id),
+            str(sa_4.service_account_id),
+            str(sa_5.service_account_id),
+        }
+
+        # For this test, we don't associate any service accounts to it.
```
petracihalova, line 274: same here

tests/management/principal/test_it_service.py
```diff
@@ -0,0 +1,334 @@
...
+        # Add the service accounts to the group.
+        group.principals.add(sa_1)
+        group.principals.add(sa_2)
+        group.principals.add(sa_3)
+        group.principals.add(sa_4)
+        group.principals.add(sa_5)
+        group.save()
+
+        # Simulate that a few client IDs were specified in the request.
+        request_client_ids = set[str]()
```
petracihalova, line 291: the `request_client_ids` is initialised as `set` of strings and then the `UUID` instances are added

tests/management/principal/test_it_service.py
```diff
@@ -0,0 +1,334 @@
...
+                self.assertEqual(
+                    True,
+                    key in group_service_accounts_set,
+                    "a client ID which was not part of the group was incorrectly flagged as if it was",
+                )
+            else:
+                self.assertEqual(
+                    True,
+                    key in request_client_ids_str,
+                    "a client ID which was part of the group was incorrectly flagged as if it wasn't",
```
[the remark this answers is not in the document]
MikelAlejoBR: Actually the test tests mixed results: the group only has two service accounts, and the other three are not part of it. It returns false for some service accounts, and the code takes the `else:` path in the snippet you reviewed. Nonetheless, I've refactored it to include the `False` explicit assertion and also renamed the key and the value to make it clear what they represent.

tests/management/principal/test_it_service.py
MikelAlejoBR: Refactored!

tests/management/principal/test_it_service.py
MikelAlejoBR: Refactored!

tests/management/principal/test_it_service.py
MikelAlejoBR: Refactored!

tests/management/principal/test_it_service.py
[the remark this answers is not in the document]
petracihalova: I'm sorry I explained it wrong ... what I wanted to say is something different
if I understand it correctly, you use in your tests 5 UUIDs (A, B, C, D, E) and desired output is that 2 of them belongs into group (A, B) and 3 of them not (C, D, E)
`group_service_accounts_set` contains A, B
`request_client_ids_str` contains A, B, C, D, E

so in the end the result contains these keys and values
`result = {A: True, B: True, C: False, D: False, E: False}`

and my question is what happens if the function in the test returns wrong result like this?
`result = {A: False, B: False, C: False, D: False, E: False}`

I think in this case the test will pass but it should fail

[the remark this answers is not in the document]
MikelAlejoBR: Aaaah, yes, you are right. I have created a new set with the specified UUIDs that should not be in the group, and now the assertion in the loop compares against that set, instead of the `request_client_ids_str` set.

Thanks for pointing it out!

tests/management/group/test_view.py
MikelAlejoBR: I have added a couple of tests for the view:

- A test for mixed results.
- A test for when a different group which does not have any service accounts is specified.

docs/source/specs/openapi.json
```diff
@@ -1273,6 +1273,18 @@
...
+            "name": "service_account_client_ids",
+            "required": false,
+            "description": "By specifying a list of client IDs with this query parameter, RBAC will return an object with the specified client ID and it's matching boolean value to flag whether the client ID is present in the group or not",
+            "schema": {
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            }
+          },
```
petracihalova, line 1287: I realised that this new query param is resulting in new response schema, can you please add it? I believe we need it for QE tests cc @vbelchio @mfindra

MikelAlejoBR: I have updated the specification for this one. I have used `additionalProperties` for the response, although I am not very sure if that's correct or not. Let me know if this breaks any clients.

petracihalova: awesome, thank you

rbac/management/principal/it_service.py
```diff
@@ -315,6 +317,24 @@ def get_service_accounts_group(self, group: Group, user: User, options: dict = {
...
+    def generate_service_accounts_report_in_group(self, group: Group, client_ids: set[UUID]) -> dict[str, bool]:
+        """Check if the given service accounts are in the specified group."""
+        # Fetch the service accounts from the group.
+        group_service_account_principals = (
+            group.principals.values_list("service_account_id", flat=True)
+            .filter(type=TYPE_SERVICE_ACCOUNT)
+            .filter(service_account_id__in=client_ids)
+        )
+
+        # Mark the specified client IDs as "present or missing" from the result set.ç
```
petracihalova, line 329: `set.ç` in the comment is probably a typo?

petracihalova (approved):
good job Mikel 👍

tests/management/group/test_view.py
petracihalova: thank you ❤️

rbac/management/group/view.py
```diff
@@ -653,6 +655,79 @@ def principals(self, request: Request, uuid=None):
...
+                        },
+                    )
+
+                # Turn the received and comma separated client IDs into a manageable set.
+                received_client_ids: set[str] = set(service_account_client_ids_raw.split(","))
+
+                processed_client_ids: set[UUID] = set()
+                for rci in received_client_ids:
+                    try:
+                        processed_client_ids.add(UUID(str(rci)))
```
astrozzc, line 699: Seems the str() is not necessary cause it is already a string?

rbac/management/principal/it_service.py
MikelAlejoBR: Yes, you're right. Good catch!

rbac/management/group/view.py
```diff
@@ -653,6 +655,79 @@ def principals(self, request: Request, uuid=None):
...
+                                        "source": "groups",
+                                        "status": str(status.HTTP_400_BAD_REQUEST),
+                                    }
+                                ]
+                            },
+                        )
+
+                # Generate the report of which of the tenant's service accounts are in a group, and which
+                # ones are available to be added to the given group.
+                it_service = ITService()
```
astrozzc, line 716: Do you mind making this ITService a singleton too? Or a JIRA task is fine. So we don't have to init it for each request.

rbac/management/principal/it_service.py
```diff
@@ -315,6 +317,24 @@ def get_service_accounts_group(self, group: Group, user: User, options: dict = {
         else:
             return service_accounts
 
+    def generate_service_accounts_report_in_group(self, group: Group, client_ids: set[UUID]) -> dict[str, bool]:
+        """Check if the given service accounts are in the specified group."""
+        # Fetch the service accounts from the group.
+        group_service_account_principals = (
+            group.principals.values_list("service_account_id", flat=True)
+            .filter(type=TYPE_SERVICE_ACCOUNT)
+            .filter(service_account_id__in=client_ids)
```
astrozzc, line 326: I believe we can supply a list of string directly in this query instead of making them into UUIDs, so it can save us some time converting back and forth. You can confirm by if that still passes your unit test.

And turn the group_service_account_principals into a set would help improving the speed when checking if the id is in it.

astrozzc (approved):
Nice job! Just some small comments

rbac/management/group/view.py
astrozzc: Oh, sorry, ignore my comment. Just remembered we talked about this the other day. So it would be fixed once you merge the other PR.

rbac/management/group/view.py
MikelAlejoBR: Not really, refactored!

rbac/management/principal/it_service.py
MikelAlejoBR: Yeah, I wanted to be like as most type precise as possible. But it is true that it leads to many conversions left and right. I've the code that attempts to build UUIDs from the given client IDs, though, to make sure that we receive UUIDs and not any other type of string.

rbac/management/group/view.py
MikelAlejoBR: Yeah, I was going to tell you that. Thank you, though!

MikelAlejoBR:
All right, rebased to resolve conflicts!

PR #1010, opened 2024-01-31
[WIP][RHCLOUD-30665] moved logic for limit and offset out of it_service class to fix the calculation of 'count' in the response

petracihalova:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-30665

with previous logic the "count" for the response was not correctly calculated and that caused that "links" were not created correctly and this led to the UI issues with pagination

this PR move the limit+offset logic to the `it_service.py` file because first we need to calculate "count" and then filter the service account for the page

## Local Testing
check acceptance criteria in the JIRA ticket but generally you can test this like this
send request to the `GET principals/` or GET `groups/{uuid}/principals/` endpoint and filter only service accounts, try different combination of limit and offset and check:

- values of count, limit, offset are set correctly
- links are created correctly
- returned data contains correct service account

### DIFF
--- a/rbac/management/group/view.py
+++ b/rbac/management/group/view.py
@@ -54,7 +54,6 @@ from rest_framework.decorators import action
 from rest_framework.filters import OrderingFilter
 from rest_framework.response import Response
 
-from api.common.pagination import StandardResultsSetPagination
 from api.models import Tenant, User
 from .insufficient_privileges import InsufficientPrivilegesError
 from .service_account_not_found_error import ServiceAccountNotFoundError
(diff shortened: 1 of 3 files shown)


MikelAlejoBR:
> > I don't know if I agree with moving the limit and offset values outside the IT Service class, just because the way I see it view classes should simply get the results from the service or database layers and polish them to make them look as the contracts.
> > I feel like applying the limit and offset to the service accounts collection in the view will trigger a database query from there, where in my opinion that should be done in the IT Service class. Since we're applying all the query options there, I think it makes sense to have them in the same function, instead of having them scattered.
> > What do you think?
> 
> well the main reason for this was we need to calculate the "count" as number of all service accounts and then sometimes return only some of them according the value of offset and limit and now (before the change) the count is calculated from the filtered data so the count and links are not created correctly
> 
> for example if I have in the group 5 service accounts and I set the limit=1, then in the response the count=1 but it should be count=5 ... if we leave the logic for the limit and offset inside the IT Service class, then we have to return not only list of SA but the value of the "count" too and then override the count value in the `paginator` before `page = self.paginate_queryset(service_accounts)` ... I can create this version for comparison and at the same time I am very open to any different solution because I can't think of any other way to solve it so any help very welcome

I understand, yeah. Okay, let's go with it, then!

MikelAlejoBR (approved):

petracihalova:
I prepared another PR with slightly different logic as was discussed for the comparison, let me know, what you think ... I believe the https://github.com/RedHatInsights/insights-rbac/pull/1011 is maybe better approach

PR #1011, opened 2024-01-31
[RHCLOUD-30665] fix logic for the 'count' calculation - needed for the response

petracihalova:
this is second version of https://github.com/RedHatInsights/insights-rbac/pull/1010 (in the end only one version will be merged)

## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-30665

with previous logic the "count" for the response was not correctly calculated and that caused that "links" were not created correctly and this led to the UI issues with pagination

this PR adds a new return value for the "count" to the IT Service class methods used for Service Accounts

## Local Testing
check acceptance criteria in the JIRA ticket but generally you can test this like this
send request to the `GET principals/` or GET `groups/{uuid}/principals/` endpoint and filter only service accounts, try different combination of limit and offset and check:

- values of count, limit, offset are set correctly
- links are created correctly
- returned data contains correct service account

### DIFF
--- a/rbac/management/group/view.py
+++ b/rbac/management/group/view.py
@@ -44,7 +44,7 @@ from management.permissions import GroupAccessPermission
 from management.principal.it_service import ITService
 from management.principal.model import Principal
 from management.principal.proxy import PrincipalProxy
-from management.principal.serializer import PrincipalSerializer
+from management.principal.serializer import PrincipalSerializer, ServiceAccountSerializer
 from management.principal.view import ADMIN_ONLY_KEY, USERNAME_ONLY_KEY, VALID_BOOLEAN_VALUE
 from management.querysets import get_group_queryset, get_role_queryset
 from management.role.view import RoleViewSet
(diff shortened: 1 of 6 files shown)


MikelAlejoBR (approved):
This one looks cleaner to me!

rbac/management/group/view.py
```diff
@@ -797,8 +797,8 @@ def principals(self, request, uuid=None):
                     )
 
                 # Prettify the output payload and return it.
-                page = self.paginate_queryset(service_accounts)
-                serializer = PrincipalSerializer(page, many=True)
+                self.paginate_queryset(service_accounts)
```
abaiken, line 800: Do we still need to call this if we are not assigning it to a variable?

petracihalova: first I wanted to answer like this ... yes we need this will set correctly the `self.paginator` and this is needed for `meta` and `links` to be created correctly in the `StandardResultsSetPagination` class `when self.get_paginated_response(service_accounts)` is called 2 lines below

then I realised how we handle user based principals in this endpoint and added a serializer for service account so logic for displaying correct data according limit and offset will be handled by rest-framework's serializer ... I added this change as fixup to be visible what I changed but I will rebase it after PR review ... let me please know, what you think

abaiken: Thanks Petra! I see what you did & it makes much more sense to me now.

abaiken (approved):
LGTM! 🎉

PR #1020, opened 2024-02-05
[RHCLOUD-30836] flake8 config update

petracihalova:
[RHCLOUD-30836](https://issues.redhat.com/browse/RHCLOUD-30836)

flake8 supports configuration in the `tox.ini` or `.flake8` files ([docs](https://flake8.pycqa.org/en/latest/user/configuration.html#configuration-locations))
the `tox.ini` has the preference ([link](https://github.com/PyCQA/flake8/blob/5c52d752e64f210369b37654e7b26ce25e2266d6/src/flake8/options/config.py#L32))
-> in our repo we had both so only config in the `tox.ini` was used = we don't need both

and

upgraded flake8 version in pre-commit config + removed duplicated config (it will be used config from `tox.ini`)

and

I did audit of the flake8 rules we ignored

-> [E203](https://www.flake8rules.com/rules/E203.html) Whitespace before ':' - we had only 1 occurrence - I removed it from ignore list and added comment in the line where we want to ignore this rule (it is false positive output because of bug in flake8 - link part of the comment)

-> [Q000](https://github.com/zheller/flake8-quotes/) Remove bad quotes - we had 2337 occurrences but we can set that double quotes are fine too with `inline-quotes = "` so now we can remove this rule from ignore list

-> [D106](https://www.pydocstyle.org/en/stable/error_codes.html) = Missing docstring in public nested class - we have 11 occurrences (example [this](https://github.com/RedHatInsights/insights-rbac/blob/e130f4219d7b575cc12d34b996cffc5c2b67ec9c/rbac/management/group/model.py#L76) Meta class doesn't contain docstring) ... I believe we can easily add the missing docstrings if we want

-> [W503](https://www.flake8rules.com/rules/W503.html) = Line break before binary operator - we have 8 occurrences but the flake8 contains the rule [W504](https://www.flake8rules.com/rules/W504.html) too and these two rules (W503, W504) are mutually exclusive, so we have to ignore one (according the flake8 docs the W503 should be disabled by default but it is not working - so before this will be solved, we have to ignore it manually)

-> [C901](https://www.flake8rules.com/rules/C901.html) = Function is too complex - we have 11 occurrences (example method [principals](https://github.com/RedHatInsights/insights-rbac/blob/e130f4219d7b575cc12d34b996cffc5c2b67ec9c/rbac/management/group/view.py#L502) in the GroupViewSet class was evaluated as too complex = lots of cycles and conditions) ... I believe that is why we ignore it, so this will be not changed


petracihalova:
rebased, conflicts solved

astrozzc (approved):
LGTM！

petracihalova:
rebase and conflicts solved

PR #1024, opened 2024-02-07
Service Accounts principals/ endpoint - fix for "count"

petracihalova:
## Link(s) to Jira
- fix for comment in the [RHCLOUD-30665](https://issues.redhat.com/browse/RHCLOUD-30665)

it looks like the count is not calculate correctly because after it the data are filtered so we have to move the logic for it in the end of the function

and

added fields "limit" and "offset" into response of principals/ endpoint


vbelchio:
LGTM

PR #1031, opened 2024-02-16
Update rbac.settings in order to support multiple Kafka Brokers 

Ellen-Yi-Dong:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-29527

The what, why and how.
- Creating a list of brokers to connect to 
## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/rbac/settings.py
+++ b/rbac/rbac/settings.py
@@ -399,38 +399,38 @@ KAFKA_SERVERS = []
 if KAFKA_ENABLED:
     KAFKA_AUTH = {}
     if ENVIRONMENT.bool("CLOWDER_ENABLED", default=False):
-        kafka_broker = LoadedConfig.kafka.brokers
-        if len(kafka_broker) == 0:
-            ValueError("No kafka brokers available")
-        for i in range(0, len(kafka_broker)):
-            kafka_host = kafka_broker[i].hostname
-            kafka_port = kafka_broker[i].port
-            kakfka_info = f"{kafka_host}:{kafka_port}"
-            KAFKA_SERVERS.append(kakfka_info)
+        kafka_brokers = LoadedConfig.kafka.brokers
+        if not kafka_brokers:
+            raise ValueError("No kafka brokers available")
+        for broker in kafka_brokers:
+            kafka_host = broker.hostname
+            kafka_port = broker.port
+            kafka_info = f"{kafka_host}:{kafka_port}"
+            KAFKA_SERVERS.append(kafka_info)
 
-            try:
-                if kafka_broker[i].authtype.value == "sasl":
-                    KAFKA_AUTH.update(
-                        {
-                            "bootstrap_servers": f"{kafka_host}:{kafka_port}",
-                            "sasl_plain_username": kafka_broker[i].sasl.username,
-                            "sasl_plain_password": kafka_broker[i].sasl.password,
-                            "sasl_mechanism": kafka_broker[i].sasl.saslMechanism.upper(),
-                            "security_protocol": kafka_broker[i].sasl.securityProtocol.upper(),
-                        }
-                    )
-                if kafka_broker[i].cacert:
-                    KAFKA_AUTH["ssl_cafile"] = LoadedConfig.kafka_ca()
-            except AttributeError:
-                KAFKA_AUTH = {}
+        try:
+            if kafka_brokers.authtype.value == "sasl":
+                KAFKA_AUTH.update(
+                    {
+                        "bootstrap_servers": KAFKA_SERVERS,
+                        "sasl_plain_username": kafka_brokers.sasl.username,
+                        "sasl_plain_password": kafka_brokers.sasl.password,
+                        "sasl_mechanism": kafka_brokers.sasl.saslMechanism.upper(),
+                        "security_protocol": kafka_brokers.sasl.securityProtocol.upper(),
+                    }
+                )
+            if kafka_brokers.cacert:
+                KAFKA_AUTH["ssl_cafile"] = LoadedConfig.kafka_ca()
+        except AttributeError:
+            KAFKA_AUTH = {}
     else:
         kafka_host = "localhost"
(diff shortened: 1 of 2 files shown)


lindgrenj6 (approved):

rbac/rbac/settings.py
```diff
@@ -397,28 +397,33 @@
...
+            except AttributeError:
+                KAFKA_AUTH = {}
     else:
-        KAFKA_HOST = "localhost"
-        KAFKA_PORT = "9092"
-    KAFKA_SERVER = f"{KAFKA_HOST}:{KAFKA_PORT}"
+        kafka_host = "localhost"
+        kafka_port = "9092"
+
+    KAFKA_SERVER = f"{kafka_host}:{kafka_port}"
```
lpichler, line 426: @Ellen-Yi-Dong does this contains array ?

I did not run code but It looks we need to build array here from `LoadedConfig.kafka.brokers` but this looks to me that it takes just last element from that array. Same for KAFKA_AUTH. 

This var  KAFKA_SERVER is passed [here](https://github.com/RedHatInsights/insights-rbac/blob/9095381e70e2e0ac86c728e2985b654594499288/rbac/core/kafka.py#L44)  to kafka client and maybe it should be called accordingly:  `KAFKA_SERVERS`.

Also If it was not yet -  I recommend to test this change in eph. env.  - let me know if you need some guidance.

Thanks!

lpichler: FROM [docs](https://kafka-python.readthedocs.io/en/master/apidoc/KafkaProducer.html):  

> bootstrap_servers – ‘host[:port]’ string (or list of ‘host[:port]’ strings) that the producer should contact to bootstrap initial cluster metadata. This does not have to be the full node list. It just needs to have at least one broker that will respond to a Metadata API Request. Default port is 9092. If no servers are specified, will default to localhost:9092.

And I believe that list(array) what we exactly need for this change as I was reading related email.

Let me know what do you  think about this, thanks!

Ellen-Yi-Dong: Hello Libor! Thank you for the review and information! ! 
I think that it makes sense, and I'll get those changes in! (Though might reach out again for questions)

petracihalova (changes requested):

lpichler (approved):

rbac/rbac/settings.py
```diff
@@ -394,31 +394,40 @@
...
+            raise ValueError("No kafka brokers available")
+        for broker in kafka_brokers:
+            kafka_host = broker.hostname
+            kafka_port = broker.port
+            kafka_info = f"{kafka_host}:{kafka_port}"
+            KAFKA_SERVERS.append(kafka_info)
+
         try:
-            if kafka_broker.authtype.value == "sasl":
+            if kafka_brokers.authtype.value == "sasl":
```
petracihalova, line 412: `kafka_brokers` is list of brokers created few lines above (L402), I believe this list doesn't have attribute `authtype`

petracihalova (changes requested):

rbac/rbac/settings.py
Ellen-Yi-Dong: Thank you for catching that!!

petracihalova: there are more lines like that

rbac/rbac/settings.py
```diff
@@ -394,31 +394,40 @@
...
-            if kafka_broker.authtype.value == "sasl":
+            if kafka_brokers[0].authtype.value == "sasl":
                 KAFKA_AUTH.update(
                     {
-                        "bootstrap_servers": f"{KAFKA_HOST}:{KAFKA_PORT}",
-                        "sasl_plain_username": kafka_broker.sasl.username,
-                        "sasl_plain_password": kafka_broker.sasl.password,
-                        "sasl_mechanism": kafka_broker.sasl.saslMechanism.upper(),
-                        "security_protocol": kafka_broker.sasl.securityProtocol.upper(),
+                        "bootstrap_servers": KAFKA_SERVERS,
```
petracihalova, line 415: not sure if I understand what is expected here? here we are using only the first broker but we set bootstrap servers param to all brokers?

Ellen-Yi-Dong: From what I understand,  the  kafka producer can only have one type of authentication. So  if there is one broker that has that information, then KAFKA_AUTH can use that information (like the sasl_plain_password, sasl_mechanism and security protocol), and the other fields, while the bootstrap servers can take in a list of servers, which is why we have the kafka_servers variable there! 
Which is why I was thinking that if we can just the first one, but if the first broker doesn't have it, but the third broker has that information...hmmm I might have to make a different variable then...

Ellen-Yi-Dong: Here is what the brokers look like in terms of the authentication for rerference: 
Screenshot from 2024-02-21 08-10-54

petracihalova: yes that is what is not clear to me, why we set KAFKA_AUTH from first broker and not from e.g. second ... but I dont have knowledge or preference here

Ellen-Yi-Dong: Oh! All of the kafka_auth are the same for each of the brokers! If you look at the screenshot, you can see how the fields are the same for each one (and the data for each one is the same)  so that's why if we set from KAFKA_AUTH from the first broker, it should work for the other 2 brokers! If it helps, I'm also down to call/meet/talk about it anytime too!!

petracihalova: I see 🤩 that's the part I haven't seen before!!! thank you Ellen for the explanation

abaiken (approved):
LGTM!

PR #1039, opened 2024-02-22
cryptography package upgrade (to solve grype check failure)

petracihalova:


abaiken (commented):
This LGTM! We also have the rtd_requirements (it should be updated if you run `make requirements` although i'm not sure if its really needed.

abaiken (approved):

PR #1043, opened 2024-02-28
fix for query param 'service_account_clinet_ids' used together with the 'limit' or 'offset' param

petracihalova:
JIRA [RHCLOUD-30299](https://issues.redhat.com/browse/RHCLOUD-30299)
fix for [PR#999](https://github.com/RedHatInsights/insights-rbac/pull/999)

the query param  `service_account_client_ids` should be used alone (without other query params) and this case doesn't support pagination ... but we have to handle situation when `limit` and/or `offset` param are used too


MikelAlejoBR (approved):
LGTM!

PR #1045, opened 2024-02-29
updated schema in open api specification

petracihalova:
updated schema for `/groups/{uuid}/principals/` endpoint and 'service_account_client_ids' query param

the query param should be comma separated list of UUIDs

### DIFF
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -1279,10 +1279,8 @@
             "required": false,
             "description": "By specifying a list of client IDs with this query parameter, RBAC will return an object with the specified client ID and it's matching boolean value to flag whether the client ID is present in the group or not. This query parameter cannot be used along with any other query parameter.",
             "schema": {
-              "type": "array",
-              "items": {
-                "type": "string"
-              }
+              "type": "string",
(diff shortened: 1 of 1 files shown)


MikelAlejoBR (approved):
LGTM!

docs/source/specs/openapi.json
```diff
@@ -1279,10 +1279,8 @@
             "required": false,
             "description": "By specifying a list of client IDs with this query parameter, RBAC will return an object with the specified client ID and it's matching boolean value to flag whether the client ID is present in the group or not. This query parameter cannot be used along with any other query parameter.",
             "schema": {
-              "type": "array",
-              "items": {
-                "type": "string"
-              }
+              "type": "string",
+              "example": "e4b74850-b934-013c-244b-6aa2427b506c,7ef40690-b936-013c-244c-6aa2427b506c"
```
mfindra, line 1283: I dont think the example is appropriate here, as it is a default value and this functionality is not primary for this endpoint, so everyone needs to remove it, and the example is also set for admin_only so it will return 400 from base page.

petracihalova: solved in #1046

PR #1047, opened 2024-02-29
removed example for 'usernames' query param

petracihalova:
**this PR is draft for open api spec improvements**

open api spec updates
- removed schema's example for `usernames` query param -> this value was used as default value in the swagger si it always had to be manually removed (=annoying)


petracihalova:
this change will be replaced by https://github.com/RedHatInsights/insights-rbac/pull/1076

PR #1051, opened 2024-02-29
move the queryset pagination after principals check in BOP - to calculate the count properly

petracihalova:
before this change we ran the removed code only because we need the list of usernames
but we can get the usernames directly from `principals_from_params`
when this variable doesnt contains any principal, it returns the empty list `[]` and that is what we need
so for this case we just get list of usernames and then validate it via BOP

this code set the `count` value
`self.paginate_queryset(resp.get("data"))`


vbelchio:
LGTM

abaiken (approved):
🎉

PR #1052, opened 2024-03-01
RHCLOUD-31275 | fix requiring tokens for access

MikelAlejoBR:
## Link(s) to Jira
- [[RHCLOUD-31275]](https://issues.redhat.com/browse/RHCLOUD-31275)

We were attempting to validate the service account usernames with the
bearer token from the user, but that was creating some problems with the
access endpoint.

For example, when an organization administrator was querying the access
endpoint to check the permissions of a particular service account, due
to how the "get_principal_from_request" and "get_principal" functions
work, we were requesting a bearer token for that. Another case was when the
user that generated the request was a service account.

However, in both of these cases we should be validating the service
account username not with an incoming token we were expecting, but with
a "behind the curtains" process as the one we have for the regular
principals. Because, otherwise, we are forcing the users to send us a
particular token with the "api.iam.service_accounts" scope when they're
not really trying to manage service accounts.

This, however, is not ready from the IT side so until that is provided
to us, we are going to skip the service accounts username validation in
these two functions. This is fine because the handlers that take care of
managing service accounts are still validating that a token with the
proper "api.iam.service_accounts" scope is coming with the request.

Also, sometimes the token validator might throw exceptions from places that
are not views per se, and that caused 500 errors because the views could
not be extracted from the context to generate the error.

### How can the bug be exploited and fix confirmed?
1. Create a service account and grab its client ID.
2. As an org admin, send a request to `GET /access/?username=service-account-${CLIENT_ID}`.

Or

1. Create a service account and grab its client ID.
2. Send a request to `GET /access`.

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/api/common/exception_handler.py
+++ b/rbac/api/common/exception_handler.py
@@ -61,6 +61,25 @@ def _generate_errors_from_dict(data, **kwargs):
     return errors
 
 
+def _generate_error_data_payload_response(detail: str, context, http_status_code: int) -> dict:
+    """Generate the payload for the "data" parameter of the response."""
+    data = {
+        "errors": [
+            {
+                "detail": detail,
+                "status": str(http_status_code),
+            }
+        ]
+    }
+
+    # Some exceptions might be raised from places that are not views.
+    view = context.get("view")
+    if view:
+        data["errors"][0]["source"] = view.basename
+
+    return data
+
+
 def custom_exception_handler(exc, context):
     """Create custom response for exceptions."""
     response = exception_handler(exc, context)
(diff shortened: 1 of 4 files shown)


vbelchio:
LGTM

petracihalova (approved):

PR #1058, opened 2024-03-08
refactor: removed redundant initialization of variable res_list

petracihalova:
the variable `res_list` is initialised few rows below so the first initialisation is not needed


vbelchio:
LGTM

PR #1059, opened 2024-03-22
[RHCLOUD-30108] all principals can list the system roles under a tenant

petracihalova:
## Link(s) to Jira
- [RHCLOUD-31592](https://issues.redhat.com/browse/RHCLOUD-31592)
- [RHCLOUD-30108](https://issues.redhat.com/browse/RHCLOUD-30108)

we found that behaviour in [RHCLOUD-30108](https://issues.redhat.com/browse/RHCLOUD-30108) is caused because the `/roles/` endpoint is called with the `?system=true` query param and this action requires to be org admin or RBAC admin (= user with the 'User Access administrator' role) ... after discussion we decided that all users should have an access to the system roles (defined in the [rbac-config](https://github.com/RedHatInsights/rbac-config) repo)

## Local Testing
send the request to the `roles` endpoint with or without `system` query param
```
GET /api/rbac/v1/roles/
GET /api/rbac/v1/roles/?system=true
GET /api/rbac/v1/roles/?system=false
```
you need to be an org admin or user with 'User Access administrator' role to list roles under a tenant
or 
with `?system=true` query param all users can list roles under a tenant
(detailed AC Checklist here [RHCLOUD-31592](https://issues.redhat.com/browse/RHCLOUD-31592))

### DIFF
--- a/rbac/management/permissions/role_access.py
+++ b/rbac/management/permissions/role_access.py
@@ -31,6 +31,9 @@ class RoleAccessPermission(permissions.BasePermission):
         if request.user.admin:
             return True
         if request.method in permissions.SAFE_METHODS:
+            system_param = request.query_params.get("system")
+            if system_param and system_param.lower() == "true":
+                return True
             if is_scope_principal(request):
                 return True
             role_read = request.user.access.get("role", {}).get("read", [])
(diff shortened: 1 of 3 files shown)


astrozzc (approved):
LGTM! Need QE's help to check if it breaks any existing feature and we can start from there.

PR #1062, opened 2024-03-28
[RHCLOUD-31691] Groups list update

petracihalova:
## Link(s) to Jira
- [RHCLOUD-31691](https://issues.redhat.com/browse/RHCLOUD-31691)

this PR removes an exception for `GET /groups/` that allowed all principals to list the groups under a tenant
listing all groups will be now possible only for users with proper permissions (org admins, rbac admins)

the exception was set before we had the "My User Access" page in the Console UI but now it not needed anymore
the principal scope will be not affected with this change (`GET /groups/?scope=principal `will continue to operate as it does today)

## Local Testing
**without the change**
all principals (org admin, rbac admin, non admin) can list groups under a tenant
`GET /groups/ `returns all groups under a tenant (with one exception -> only org admin can see the 'Default admin access' group)

**with the change**
only org admin and rbac admin can list groups under a tenant
`GET /groups/ `returns all groups under a tenant (with one exception -> only org admin can see the 'Default admin access' group)

org admin = user with `org_admin=True`, member of 'Default admin access' group
rbac admin = non org admin, member of a group with the 'User Access administrator' role
non admin = non org admin, non rbac admin

all types of principals can list groups they are in (with principal scope), this is not affected by the change in this PR
`GET /groups/?scope=principal` is working same before and after change for all principal types

### DIFF
--- a/tests/management/group/test_view.py
+++ b/tests/management/group/test_view.py
@@ -2691,11 +2691,16 @@ class GroupViewNonAdminTests(IdentityRequest):
         self.user_access_admin_role_err_message = (
             "Non-admin users cannot add 'User Access administrator' role to groups."
         )
+        self.invalid_value_for_scope_query_param = (
+            "scope query parameter value foo is invalid. [account, org_id, principal] are valid inputs."
+        )
 
     def tearDown(self):
         """Tear down group view tests."""
         Group.objects.all().delete()
         Principal.objects.all().delete()
+        Permission.objects.all().delete()
+        Access.objects.all().delete()
         Role.objects.all().delete()
         Policy.objects.all().delete()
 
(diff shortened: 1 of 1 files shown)


astrozzc (commented):
If the scope is principal, have we checked if the username is himself? If he is filtering a username other than himself, that would be a problem

petracihalova:
> If the scope is principal, have we checked if the username is himself? If he is filtering a username other than himself, that would be a problem

thats a really good point !!! yes you are right I didnt realise that you can combine `username` and `scope` params

petracihalova:
tests are failing because this PR change the logic for who can or cannot list roles under a tenant and we have to reflect this in the IQE tests ... cc: @vbelchio @mfindra

PR #1071, opened 2024-04-05
[RHCLOUD-31874] Only org admins are allowed to manipulate with the RBAC admin role

petracihalova:
JIRA: [RHCLOUD-31874](https://issues.redhat.com/browse/RHCLOUD-31874)

we have two [RBAC system roles](https://github.com/RedHatInsights/rbac-config/blob/master/configs/prod/roles/rbac.json):
- User Access administrator
- User Access principal viewer

only Organisation Administrators are allowed to manipulate with the `User Access administrator` role
and at the same time it is not possible to create a custom role with RBAC permission
- `rbac:*:*`
- `rbac:principal:read`

### DIFF
--- a/rbac/management/group/definer.py
+++ b/rbac/management/group/definer.py
@@ -36,7 +36,6 @@ from rest_framework import serializers
 
 from api.models import Tenant
 
-USER_ACCESS_ADMINISTRATOR_ROLE_KEY = "User Access administrator"
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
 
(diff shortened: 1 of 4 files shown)


astrozzc:
We want user with `rbac:*:*` to be allowed to manipulate the e.g. `rbac:principal:read` right? Here it seems it is restricting all of that

petracihalova:
> We want user with `rbac:*:*` to be allowed to manipulate the e.g. `rbac:principal:read` right? Here it seems it is restricting all of that

I believe we want only org admin can manipulate with roles that contains RBAC permission ... btw the system role with `rbac:principal:read` is part of the `Default Access` group that user based principals are in

petracihalova:
rebased + fixcommits squashed

petracihalova:
added tests

PR #1075, opened 2024-04-06
updated rbac viewer role description to be alligned with role defined in the rbac-config repo

petracihalova:
goal is to have same description as here: https://github.com/RedHatInsights/rbac-config/blob/master/configs/prod/roles/rbac.json


astrozzc (approved):
LGTM! 
Seems no need to change the version to update the role

astrozzc (approved):

PR #1076, opened 2024-04-08
updated open api spec to use the array for comma separated list of strings

petracihalova:
this is open api spec improvement to use the array of strings in cases when value can be comma separated list of strings


vbelchio (approved):
Warning: this will affect directly all the clients who are using OpenAPI to generate the classes and methods to deal with RBAC API. These parameters need to be changed to an array format in every call.
But it seems necessary to have more compatibility whit the latest versions of OpenAPI.

LGTM

PR #1087, opened 2024-05-02
[RHCLOUD-31133] Service Account filter by username fix

petracihalova:
## Link(s) to Jira
- [RHCLOUD-31133](https://issues.redhat.com/browse/RHCLOUD-31133)

- the UI needs to filter SA by client ID, the RBAC backend doesn't support this 
- because the value of service account username = prefix `service-account-` + client ID and RBAC backend supports filtering by username we decided to ignore the prefix `service-account-` for service accounts filtering by username (= with query param `principal_username` in the request)

## Local Testing
create a group with 2 service account based principals
- one SA contains character "a" in the uuid (e.g. `07f2420a-424e-4c2f-a52f-0fe160097d7e`)
- second SA doesn't contain character "a" in the uuid (e.g. `e78dc6b2-5930-4649-999f-266f4b926f3e`)
try to filter service accounts by username = "a" and only SA with "a" in the client ID should occurred in the response
```
http://localhost:8000/api/rbac/v1/groups/{{group-uuid}}/principals/?principal_type=service-account&principal_username=a
```

(because we ignore the `service-account-` prefix, the "a" found in this prefix is ignored)

### DIFF
--- a/rbac/management/principal/it_service.py
+++ b/rbac/management/principal/it_service.py
@@ -331,10 +331,12 @@ class ITService:
             group_service_account_principals = group_service_account_principals.order_by(order_by)
 
         # Check if we should filter the service accounts by the username that the user specified.
+        # In this case we want to ignore the prefix "service-account-" in the SA username and
+        # filter records only by SA client ID (uuid).
         principal_username = options.get("principal_username")
         if principal_username:
             group_service_account_principals = group_service_account_principals.filter(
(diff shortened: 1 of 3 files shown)


petracihalova:
rebased

lpichler (approved):

abaiken (approved):
LGTM 👍

PR #1094, opened 2024-05-09
ephemeral roles+permissions configmaps update

petracihalova:
this PR align roles and permissions configmaps with [rbac-config ](https://github.com/RedHatInsights/rbac-config)(stage)


petracihalova:
we dont need the configmaps here anymore - https://github.com/RedHatInsights/insights-rbac/pull/1100

PR #1101, opened 2024-05-20
add '/openapi.json' endpoint into internal open api spec

petracihalova:
we provide the internal endpoint /openapi.json but we don't have it in internal open api spec

test:
run RBAC locally and send request
```
GET http://localhost:8000/_private/api/v1/openapi.json
```
with internal user in `x-rh-identity` (`"type":"Associate"`)

### DIFF
--- a/rbac/internal/specs/openapi.json
+++ b/rbac/internal/specs/openapi.json
@@ -1037,6 +1037,28 @@
           }
         }
       }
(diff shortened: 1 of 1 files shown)


petracihalova:
rebased

lpichler (approved):

PR #1108, opened 2024-05-22
RBAC pr-check simplification

petracihalova:
we dont need to install all dependencies because the `tox` will create own virtual environment and install everything is needed https://github.com/RedHatInsights/insights-rbac/blob/281ad3a730f588eeeee6af891df67ba5003d09ba/tox.ini#L40

**how to test:**
pr-check is running successfully with unit tests + linters + mypy check


astrozzc (approved):
LGTM!

PR #1127, opened 2024-06-25
use org id instead of the tenant name in the logs to be aligned with rest of seeding logs

petracihalova:
during seeding process we use the tenant name in logs like [here](https://github.com/RedHatInsights/insights-rbac/blob/ebb2593a50ab8a6a2180d78140e95484eeaeb0b8/rbac/management/seeds.py#L29)

but on other places we use org id and because of async process it is hard to check what logs belong together 

this PR adds org id instead of the tenant name in the logs

### DIFF
--- a/rbac/management/seeds.py
+++ b/rbac/management/seeds.py
@@ -26,11 +26,11 @@ logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
 def on_complete(progress, tenant):
     """Explicitly close the connection for the thread."""
-    logger.info(f"Purging policy cache for tenant {tenant.tenant_name} [{progress}].")
+    logger.info(f"Purging policy cache for tenant {tenant.org_id} [{progress}].")
     cache = AccessCache(tenant.org_id)
     cache.delete_all_policies_for_tenant()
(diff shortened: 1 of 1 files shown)


lpichler (approved):

rbac/management/seeds.py
```diff
@@ -26,11 +26,11 @@
 
 def on_complete(progress, tenant):
     """Explicitly close the connection for the thread."""
-    logger.info(f"Purging policy cache for tenant {tenant.tenant_name} [{progress}].")
+    logger.info(f"Purging policy cache for tenant {tenant} [{progress}].")
```
lpichler, line 29: HOw this log look like ? it dumps whole tenant object ?

petracihalova: this returns the string representation of the tenant https://github.com/RedHatInsights/insights-rbac/blob/3756478b2b63dd7238fe73852d61030ece52eda2/rbac/api/models.py#L46 ... so in this case something like "Tenant 12345" ... but now I can see that better will be use `tenant.org_id` to be more readable

lpichler (approved):

PR #1130, opened 2024-06-26
add temporary logs for debugging the purge_cache() issue

petracihalova:


rbac/management/cache.py
```diff
@@ -196,8 +196,18 @@ def delete_all_policies_for_tenant(self):
         with self.delete_handler(err_msg):
             logger.info("Deleting entire policy cache for tenant %s", self.tenant)
             keys = self.connection.keys(self.key_for("*"))
-            if keys:
-                self.connection.delete(*keys)
+            logger.info(f"Content of 'keys' variable: {keys}")
+            try:
+                if keys:
```
astrozzc, line 201: The try block can be put into the if statement

astrozzc (approved):

rbac/management/cache.py
petracihalova: done

PR #1132, opened 2024-06-27
temporary logs for debugging the purge_cache() issue

petracihalova:


lpichler (approved):

rbac/management/cache.py
```diff
@@ -197,17 +197,15 @@ def delete_all_policies_for_tenant(self):
             logger.info("Deleting entire policy cache for tenant %s", self.tenant)
             try:
                 key = self.key_for("*")
-                logger.info("value of key", key)
+                logger.info(f"value of key: {key}")
             except Exception as e:
-                logger.error('error from key = self.key_for("*")', e)
-                raise e
+                logger.error(f'error from key = self.key_for("*") with error message: {e}')
```
lpichler, line 202: shouldnot be here `"` instead of ` ?

petracihalova: no because we already use `"` for `"*"`
so one option is to use single quotes or we have to escape the inner double quotes
`logger.error(f"error from key = self.key_for(\"*\") with error message: {e}")`

PR #1134, opened 2024-06-28
Basic workspace support: spec, model and api

lpichler:
## Link(s) to Jira
https://issues.redhat.com/browse/RHCLOUD-33460
https://issues.redhat.com/browse/RHCLOUD-33525

- v2 openapi spec for workspaces
- workspace data model
- v2 API for workspace

This adds for workspace basic fields:
- name
- description
- uuid
- tenant_id
- created, modified,.. 

**Note: Currently API accessible only for org admins** 

Example of API requests:

```
GET api/rbac/v2/workspaces/


{
  "meta": {
    "count": 3,
    "limit": 10,
    "offset": 0
  },
  "links": {
    "first": "http://localhost:8000/api/rbac/v2/workspaces/?limit=10&offset=0",
    "next": null,
    "previous": null,
    "last": "http://localhost:8000/api/rbac/v2/workspaces/?limit=10&offset=0"
  },
  "data": [
    {
      "name": "C",
      "uuid": "ba81d9e0-acfe-482b-92dc-05472cc532c9",
      "description": ""
    },
    {
      "name": "Workspace A",
      "uuid": "ebe6f7ea-62c6-486c-906a-41a3f8a607b0",
      "description": "A description of Workspace A"
    },
    {
      "name": "Workspace C",
      "uuid": "28efe55d-50ad-481e-a4c5-75dbd17f4f8a",
      "description": "A description of Workspace A"
    }
  ]
}
```

```
POST api/rbac/v2/workspaces/

{
  "name": "Test Workspace"
}


{
  "name": "Test Workspace",
  "uuid": "af2a4f56-32cf-4eaa-bf3c-da6b688c70ee",
  "description": ""
}
```

```
PUT api/rbac/v2/workspaces/ba81d9e0-acfe-482b-92dc-05472cc532c9/

{
  "name": "Workspace C"
}

{
  "errors": [
    {
      "detail": "Workspace already exist in tenant",
      "source": "workspace",
      "status": "400"
    }
  ]
}
```

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


lpichler:
@coderbydesign @astrozzc if you can take a look on spec and model part (first 2 commit ) it would be helpful!

rbac/management/workspace/model.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class Workspace(TenantAwareModel):
+    """A workspace."""
+
+    name = models.CharField(max_length=100)
+    uuid = models.UUIDField(default=uuid4, editable=False, unique=True, null=False)
+    description = models.TextField(null=True)
+    created = models.DateTimeField(default=timezone.now)
+    modified = AutoDateTimeField(default=timezone.now)
```
alechenninger, line 34: What fields do inventory groups have, for comparison?

rbac/management/workspace/view.py
```diff
@@ -0,0 +1,87 @@
...
+    viewsets.GenericViewSet,
+):
+    """Workspace View.
+
+    A viewset that provides default `create()`, `destroy`, `retrieve()`,
+    and `list()` actions.
+
+    """
+
+    permission_classes = (WorkspaceAccessPermission,)
```
alechenninger, line 44: We should leverage Kessel itself for access controls on these APIs (essentially, Kessel-on-Kessel).

I don't have a complete schema for this yet, but I would consider this placeholder until we can do that. It would be a good way to exercise the Relations infrastructure ourselves by using it for RBAC v2 itself.

rbac/management/permissions/workspace_access.py
```diff
@@ -0,0 +1,26 @@
...
+"""Defines the Audit Log Access Permissions class."""
+from rest_framework import permissions
+
+
+class WorkspaceAccessPermission(permissions.BasePermission):
+    """Determines if a user is an Account Admin."""
+
+    def has_permission(self, request, view):
+        """Check permission based on Account Admin property."""
+        return request.user.admin
```
alechenninger, line 26: See comment below – before production I think we'll want this to use the Relations API itself to check this.

rbac/management/migrations/0044_workspace_workspace_unique_workspace_name_per_tenant.py
```diff
@@ -0,0 +1,37 @@
...
+                ("modified", management.rbac_fields.AutoDateTimeField(default=django.utils.timezone.now)),
+                ("tenant", models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to="api.tenant")),
+            ],
+            options={
+                "ordering": ["name", "modified"],
+            },
+        ),
+        migrations.AddConstraint(
+            model_name="workspace",
+            constraint=models.UniqueConstraint(fields=("name", "tenant"), name="unique workspace name per tenant"),
```
alechenninger, line 35: I think this may be too strict. We may want to allow workspaces be named the same _within_ a parent workspace. This might be one reason to store some amount of hierarchy in RBAC data. Imagine trying to create a "Production" workspace within the workspace named after a project, except someone else already created a "Production."

If we had a "parent" field, this would be unique on name and the parent foreign key.

For future consideration, this would also mean moving a Workspace could fail (i.e. 400 error) if the name conflicted with one already underneath the new parent.

rbac/management/workspace/model.py
alechenninger: We should also reference their validation rules (e.g. if they allow names longer than 200 chars, we could have a migration problem).

rbac/management/migrations/0044_workspace_workspace_unique_workspace_name_per_tenant.py
alechenninger: I brought this up on our UXD call today and we'll discuss it more Wednesday so we can come back to this.

rbac/management/workspace/model.py
alechenninger: Another note: we may want to add an integer `version` field that gets incremented with every update.

I'm still noodling on this. But the idea would be to assist in synchronizing spicedb & RBAC db's accurately, without out-of-order updates. It would also help with just end user concurrency control (e.g. two users try to update a workspace at the same time).

It may not be necessary. It depends on how we want to synchronize data. I'll work on this some more and get back to you.

lpichler: @alechenninger 

Inventory Groups:
DB:
id - uuid
org_id - varchar(36) 
account - varchar(10)
name - varchar(255)
created_on - TIMESTAMP WITH TZ
modified_on - TIMESTAMP WITH TZ


API:
```
   {
      "id": "f34402bf-a002-49d4-b1d6-e89afd18e6fa",
      "org_id": "11789772",
      "account": "6089719",
      "name": "abc_237464d5-890c-4a4c-ae4a-dc8ecda04637",
      "host_count": 1, (calculate in their python app)
      "created": "2024-03-27T12:31:16.441718+00:00",
      "updated": "2024-04-10T09:12:24.313976+00:00"
    }
```

[HBI Group Model](https://github.com/RedHatInsights/insights-host-inventory/blob/0c8cb31702b9d8c3fd91bb139afdab91346b3b84/app/models.py#L506-L512)
[HBI Group Migration](https://github.com/RedHatInsights/insights-host-inventory/blob/master/migrations/versions/ada40da60e4e_add_groups_and_hosts_groups_tables.py#L22)
[HBI group API](https://console.redhat.com/docs/api/inventory/v1#operations-groups-api\.group\.get_group_list())
[HBI group validation for creation](https://github.com/RedHatInsights/insights-host-inventory/blob/master/app/models.py#L885)

lpichler: @alechenninger 

Based on that I will change length of name field to 255.

DB changes about sync - I would do this in follow up PR - once we will discuss this(I think we have meeting in calendar about  it)

OrgID/Account is stored in separated table in api_tenant as there is reference for it workspace#tenant_id (thanks to TenantAwareModel )


As I understand we need also migrate hosts to DB or just to Relation API ? Anyway If so we can also add it in followup - there M:N between inv. group and hosts - so it would need [joining table](https://github.com/RedHatInsights/insights-host-inventory/blob/master/migrations/versions/ada40da60e4e_add_groups_and_hosts_groups_tables.py#L36).

rbac/management/permissions/workspace_access.py
lpichler: https://github.com/RedHatInsights/insights-rbac/pull/1134#discussion_r1662075500

rbac/management/workspace/view.py
lpichler: (I guess we need ticket for it - at least from PRBAC side)

rbac/management/migrations/0044_workspace_workspace_unique_workspace_name_per_tenant.py
lpichler: Yes you are right we should consider UXD perspective -  let's wait here for that discussion in this case.


NOTE: IG have this constraint as well [here](https://github.com/RedHatInsights/insights-host-inventory/blob/master/migrations/versions/ada40da60e4e_add_groups_and_hosts_groups_tables.py#L30)

lpichler:
@alechenninger also question is what filtering option we need here ? name only ? (name - exact match, substring ) - this could also added in follow up PR.

rbac/management/workspace/model.py
alechenninger: > As I understand we need also migrate hosts to DB or just to Relation API ? Anyway If so we can also add it in followup - there M:N between inv. group and hosts - so it would need joining table.

Hosts: Just to Relations API. We may need to migrate more for Common Inventory but TBD.

Inventory Groups: To both RBAC DB and Relations API.

> There is uuid on the workspace model,
> 
> I believe we should also another column for inventory group uuid - like inventory_group_uuid.

I think we might rather want to migrate those UUIDs as is for compatibility. Is there any reason to avoid reusing those UUIDs? That way Workspace API can be forwarded to directly from HBI and not have to include two ways of referring to a Workspace.

**Additional note on validation**

I would like to consider adding a _minimum_ Workspace name length. When searching for a workspace, if you need to find a workspace that's only named "a" then this is somewhat of an operational problem because we have to pull every workspace that matches a very generic criteria. If you know every workspace is at least 3-4 letters or whatever, then you would never need to specify less for filters.

The problem is this might conflict with existing Inventory Groups, since looking through their code I do not see a minimum enforced. Is it possible to find out what the current minimum Inventory Group name length is in practice?

It's not a hard requirement I'd say but something we might thank ourselves for later.

alechenninger:
> @alechenninger also question is what filtering option we need here ? name only ? (name - exact match, substring ) - this could also added in follow up PR.

Yeah good q. This is really another UXD question. See this thread: https://redhat-internal.slack.com/archives/C064X43CMLK/p1719914264363419

Searching a substring e.g. `LIKE '%xxx%'` can get expensive in my experience, though I think PG has some tricks up its sleeves. On the other hand, with a limited number of workspaces per tenant, we might be okay with an index scan.

rbac/management/workspace/model.py
alechenninger: > Shared vs separated UUID
> My the only point is possible conflict due to same UUIDs for our UUIDs for non IG workspace and theirs.
> What if they change this uuid identification ?
> Basically we would rely on their UUIDs generation and their validations for it.

I would like to avoid this problem by basically doing a one time migration from their data, and then after which they no longer have any inventory group data of their only. We replace inventory groups with workspaces and therefore there isn't a separate inventory group ID any more. There is just workspaces which replaced inventory groups.

lpichler:
@alechenninger @coderbydesign @astrozzc 

I updated openapi v2 spec and I added [typespec](https://typespec.io) source file and script which generates openapi spec.
Follow ups from PR's feedback:
- Create Ticket to solve access to workspace endpoint with Kessel [Comment](https://github.com/RedHatInsights/insights-rbac/pull/1134#discussion_r1662077597)
- Filtering, Matching

Spec with typespec file to review is available here in playground:

[TypeSpec playground](https://typespec.io/playground?options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40typespec%2Fhttp%2Fall%22%5D%7D%7D&c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7CtIZcmVzdNUZb3BlbmFwaTPVHXZlcnNpb25pbmciOwoKdXNpbmcgVHlwZVNwZWMuSHR0cDvQFVJlc3TRFVbJRzsKCkBzZXJ2aWNlKHt0aXRsZToiUm9sZSBCYXNlZCBBY2Nlc3MgQ29udHJvbCBmb3IgS2Vzc2VsIFByb2plY3QifSnHQmVyKCLkAO5zOi8vbG9jYWxob3N0OnvkAMJ9L3tiYXNlUGF0aH0iLCAiRGV2ZWxvcG1lbnQgU8U7IiwgeyDIJDogc3RyaW5nPSJhcGkvcmJhYy92MiIsIMRLyhw4MDAwIiB9KdJ%2BY29uc29sZS5zdGFnZS5yZWRoYXQuY29t7wCGU8Qf%2FwCA6wCA3WvZZVByb2R1Y3Rpb27fatBq5wHqZWQo5wGtcykKCm5hbWVzcGFjZecBhlJCQUM7CgplbnVtIMgnewogICAgdjIsCn0KCm1vZGVsIExpbmsgxhhAZm9ybWF0KCJ1cmkiKcUTZmlyc%2BkBWCA9ICIv5ACNdjIvKHJlc291cmNlcykvP29mZnNldD0wJmxpbWl0PTEwIjvYUXByZXZpb3Vz31TJVDLfVcUTbmV4%2FwClylE031HFE2xh%2FwD2y1E5zVHpAVhNZXRh5wFYY291bnQ6IGludDY0ID0gMzDKJkxpc3RSZXBvbnNlPFQ%2BxzBkYXRhOiBUW13mAKBsaW5rczrlAanGEW1lxCDEYctwVGltZXN0YW1wc8dGQGRvYygiMjAxOS0wMS0yMVQxNzozMjoyOFrnAOVjcmVhdGVkOiDmAMlEYXRlxEXGZN8%2BICBtb2RpZmnUP%2B0A3VF1ZXJ5UGFyYW3uAJjFE2V0ZXLlBDhzZWxlY3TkAW10aGUgYW3kATEgb2blAQogcmV0dXJuZWQux3pAbWF4VmFsdWUoMTAwMMgUaW7HFMcRccR5IOUBmj%2FlAXxlZ2VyID0gMTAszCHmAL3NIsgh8wCob3JkZeUCFGJ5IHbEby4gRm9yIGlu5AOSZckfLCBzdXBwbHkgJy0nIGJlZm9yZeUA2XDETcY5xCNjaCBhczogP8U3X2J5PS3kA8ToAObnAKDHHD86IE%2FEC0J5RGVmYXVsdOUA7nPlAW7qAsF1aWQiKQpzY2FsYXIgVVVJRCBleHRlbmRz5wLM6AQQ1EvnAZ%2FkAIPmAQHoAeLEY%2BcB0k9rUmVz5QKuRm9yTcQTPFQ%2BxjhAc3RhdHVzQ29kZcUQyg86IDIw5wFULi4uVMtPQ%2BYCgt9U2FQxzlQvKioKICogUucDoiBEZWxldGVkCiAqL8dvTm%2FkBqxlbnTIccgsxyvfdch1fewFbEVycm9y6QLbZcQO6AE9ZWzGHk1lc3NhZ2U0MDDHJ8QBZGV0YWlsP%2BwEZUpTT07kAiNzZSDFSCAtIEV4cOcDGOUCNTrkBAxlIDEgY29sdW1uIDEgKGNoYXIgMCki5gEM6gCwzVo0MDTMIeUA9M0h5gCOIsUjfQr%2FAMzlAMwz%2FgDMWW91IGRvIG5vdCBoYXZlIHBlcm1pc%2BQGhCB0b8QO5AKzIHRoaXMgYeUHAi7%2FAMIwM%2F8Awv8AwvIAwjT%2BAMJOb3QgZm91bmT%2FAJrqAVzdd0Jhc2lj6AJ4PFNUQVRVU%2BcC5u8CdSDPbsQUOiDGMuYFnN1pyGQ0MDDVX9VeNOgDocQBxVdzOvACvVtd333OfUFscmVhZHlFeGlzdOkD0f8AjPoAjFTffs9%2BM9Vv9gFaNDAz%2FwD8MDPffs9%2BNN9%2Bzn40334wNN9%2Bz34x337OfucFRcdc5wY6xVLFNvAGP30KCuQHKXJm5AT85gW%2F6AVFPEJvZHnkDEboCMfmCGA85QXTxRgoQGJvZHkgxAU6IMguKTr5BgfKOiB85gDA5gVuLvAA8iB82x4z3R4w0x7uA6w1MDA%2B7wD55gm25wD36Qm3bGlzdCguLi7vCPcp5QK%2B%2FQCF%2FwDB%2FwCj9QCjUmVh5AFR7wCj5AOTKOoCH%2BUI4lVuaXF1ZSBpbmRlbnRpZmljYXRvcucIhcUmcGF0aCDkCFs65QhT5gSE5ADW%2FwDU3B7%2BAZU0%2FwDy9QDyVXBk7QKOLCBSLCBF6AD6dcUe%2FwD8%2FwD86AD8xRrsAsvoARLzCR9SPv4Cf%2B0FJ0XeLf8BU%2F8BNfUBNeYI%2BukJNOcMgMUV%2FwEs%2FwEs6wIo%2Fwl3ZP8A9Nwe%2FwD07AD0QHJvdXRlKCIvd29ya%2BUJvHMiKQpAdGFnKCJXzRPlAO5PcGVyYeQImXMgYWJvdXQgzTfqCgDKO%2BcBPeYFlMkXRuQAr2Fk6whsxmDJISBB6wFExFjsCF3MJusCfMVCRGVzY3JpcOUQP29mynLNUWTKJe0Itts97Qhk7wDO7ADHa2V59QMaLi4u8AD%2BzR3qDqn%2FCM7EEOoIf8lU7AMy%2FgnDyTMgYcY0IGXFNWluIHRlbmFudP8J2jAw%2Fgp06QIT7goBYWxpYXPvBx1BxRQ9zxY88AEpPuwPfcYs6gJ56wC45w4jc3VtbWFyed8rxStwb3N0IG9w5weVIGn1AKEuxh7qAJc%2BO%2FUCB%2BcHAOUDYsRL7BDGzDzqAQ7sBzLoAQzMFMo3zFztAQ9CeSBk5g6wLCByxyFzIGFyZSBzb3LkBEdpbiBhc2NlbmTkAZ%2FlDzjkD6fqARTsD13pATrEb8okc%2BUP5u4BOmdl5QE55Afe6gDs7AE1xBrnASjmAO3sB3zoAO3MFPEBJOYA4UdldCBh%2FwHv5gC13yrpALTkAvXkALTMfeYAtMQa7QC07gc%2B6AC2zhbxAq8s6gITLP8DsOYDsO0A8EVkae0Ax%2FEA58Uh5BE47gHEIG9y7ATYyDdwdeUA9OYH6eQA9vMA0y7GHucA%2BugFSOQGhv0Aiccj7QCLyCXmBy3EescKaXPvB1YuxhnkBoY%3D&e=%40typespec%2Fopenapi3)

NOTE: you can switch it to display tsp in swagger UI:

Screenshot 2024-07-08 at 14 48 02

How to install and compile tsp file is covered Readme update.

docs/source/specs/typespec/main.tsp
```diff
@@ -0,0 +1,251 @@
...
+
+/**
+ * Resource Deleted
+ */
+model NoContentResponseResourceDeleted{
+    @statusCode
+    statusCode: 201,
+}
+
+namespace Errors {
```
alechenninger, line 91: Would it be possible to make our error objects comply with https://www.rfc-editor.org/rfc/rfc9457.html ?

If possible it would be nice to define a generic type which models this, and allows extension to define the schema of additional fields.

I sketched out something like this:

```
@error
model Problem<T extends ProblemDetails> {
  ...T,
}

enum ProblemType {
  "https://project-kessel.org/problems/oops",
}

model ProblemDetails {
  type: ProblemType,
  status: int16,
  title?: string,
  detail?: string,
}

model Oopsy extends ProblemDetails {
  type: ProblemType.`https://project-kessel.org/problems/oops`,
  oopsy: string,
  @statusCode
  status: 500,
  title: "There ways an oopsy",
}
```

There is an open upstream issue to include something native in Typespec for this which may be slated for August: https://github.com/microsoft/typespec/issues/2553

As long as what we did is compatible, shouldn't matter in the meantime.

docs/source/specs/typespec/main.tsp
```diff
@@ -0,0 +1,251 @@
...
+    @get op read is ReadResourceAlias.read;
+
+    alias UpdateResourceAlias = UpdateResource<WorkspaceForRead, Workspace, ErrorMessage400WorkspaceAlreadyExist>;
+    @doc("Edit a workspace")
+    @summary("Edit the workspace name or description")
+    @put op update is UpdateResourceAlias.update;
+
+    @doc("Delete a workspace")
+    @summary("Delete the workspace")
+    @delete op delete is DeleteResource.delete;
```
alechenninger, line 250: IMO there is a little bit too much templating / aliasing going on here. It's a little hard to understand and might hide errors.

For example, `ListResource.list` current returns type `T[]`, but is instantiated with a WorkspaceListResponse, which already includes an array of data. So it looks like an array of objects with metadata and an array of objects.

e.g. this is generated:

```
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/Workspaces.WorkspaceListResponse'
```

I think you want

```
interface ListResource<T extends ListResponse> {
    list(...ListQueryParams): T |
    Errors.ErrorResponse401 |
    Errors.ErrorResponse403 |
    Errors.ErrorBasicResponse<500>;
}
```

It might be enough to just define reusable models for different request/response patterns, but drop the common interface operations. We could probably define an alias instead for some of the common error types to make those more reusable, for example.

```
alias CommonErrors = Errors.ErrorResponse401 | Errors.ErrorResponse403 | Errors.ErrorBasicResponse<500>;

// ...

  op create(@body body: CreateWorkspaceRequest): CreateWorkspaceResponse | CommonErrors | Problem.SomeUniqueProblemToWorkspaceCreation;
```

There is also some built-in stuff for REST interfaces in Typespec which might be interesting. Nothing is jumping out at me right away other than what you're already using basically but worth looking into I think. It has some interfaces which are meant to be extended I guess.

docs/source/specs/typespec/main.tsp
```diff
@@ -0,0 +1,251 @@
...
+
+model Meta {
+    count: int64 = 30
+}
+
+model ListResponse<T> {
+    data: T[];
+    links: Link;
+    meta: Meta;
+}
```
alechenninger, line 42: I'm not sure about this ListResponse schema for a couple of reasons:

- it relies on offset pagination rather than [keyset/cursor pagination](https://use-the-index-luke.com/no-offset) (Django Rest has [this](https://www.django-rest-framework.org/api-guide/pagination/#cursorpagination) to say about it.) I'm not sure this matters really, though because...
- I think the shape of this API might not warrant pagination regardless, since it is sort of "built-in" paginated to the children of one parent of the hierarchy at a time. Replace "parent" with "page" and it's similar considering limits to the total number of workspaces and limits to the number of children per node. Those limits effectively define the page size.

This may change considering the query problem, so another option which might be pragmatic for now is to simply omit the List endpoint since it is still an active research topic.

docs/source/specs/typespec/main.tsp
```diff
@@ -0,0 +1,251 @@
...
+    alias CreateResourceAlias = CreateResource<WorkspaceForRead>;
+    @doc("Create workspace in tenant")
+    @summary("Create workspace in tenant")
+    @post op create is CreateResourceAlias.create<Workspace>;
+
+    model WorkspaceListResponse is ListResponse<Workspace>;
+    alias ListResourceAlias = ListResource<WorkspaceListResponse>;
+    @doc("By default, responses are sorted in ascending order by workspace name")
+    @summary("List workspaces for tenant")
+    @get op list is ListResourceAlias.list;
```
alechenninger, line 236: I'm not sure how you add additional typed query parameters to this (probably a template parameter to ListResource, but see comment below about templating), but one of the things we probably need will need here is a "parent" parameter.

On the other hand, as mentioned above, we may want to omit this endpoint entirely for now and come back after our spikes.

docs/source/specs/typespec/main.tsp
```diff
@@ -0,0 +1,251 @@
...
+    Errors.ErrorResponse401 |
+    Errors.ErrorResponse403 |
+    Errors.ErrorBasicResponse<500>;
+}
+
+@route("/workspaces")
+@tag("Workspaces")
+@doc("Operations about workspaces")
+namespace Workspaces {
+    model WorkspaceForRead {
```
alechenninger, line 207: This is named `WorkspaceForRead` but I think it is actually more `WorkspaceForWrite`?

docs/source/specs/typespec/main.tsp
```diff
@@ -0,0 +1,251 @@
...
+        @doc("Workspace A")
+        name: string = "Workspace A";
+        @doc("Description of Workspace A")
+        description?: string = "Description of Workspace A";
+    }
```
alechenninger, line 212: Probably makes sense to include `parent?: UUID`.

We may want to change that to `parent` (not optional) later but will need to think through how tenants are registered in the API / backend.

docs/source/specs/typespec/main.tsp
alechenninger, line 250: For delete we probably want an error/problem type for "cannot delete, it isn't empty."

docs/source/specs/typespec/main.tsp
```diff
@@ -0,0 +1,251 @@
...
+    statusCode: 201,
+    ...T
+}
+
+/**
+ * Resource Deleted
+ */
+model NoContentResponseResourceDeleted{
+    @statusCode
+    statusCode: 201,
```
coderbydesign, line 88: Should we change this to a `204` status code? Without a supplemental response message (which would support a `200`), it seems like this is the best option: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/DELETE#responses

```suggestion
    statusCode: 204,
```

docs/source/specs/typespec/main.tsp
coderbydesign: We may want to get input from platex on this and see if there are other standards/patterns in other platform APIs (I don't believe there is, but it might be worth guidance). Right now the pattern for RBAC in v1 is limit/offset pagination, which would primarily have implications on the UI clients if we moved away from that. Not suggesting we shouldn't, but a consideration.

rbac/management/migrations/0044_workspace_workspace_unique_workspace_name_per_tenant.py
coderbydesign: Yeah it does feel like we'd need it to be unique based on the parent/tenant. HBI groups don't currently support hierarchy so it makes sense that they'd have it enforced at that level.

tests/management/workspace/test_view.py
```diff
@@ -0,0 +1,340 @@
...
+            "description": "New Workspace - description",
+            "tenant_id": self.tenant.id,
+        }
+
+        workspace = Workspace.objects.create(**workspace_data)
+
+        url = reverse("workspace-detail", kwargs={"uuid": workspace.uuid})
+        client = APIClient()
+
+        workspace_data["name"] = "Updated name"
```
coderbydesign, line 143: I think the validation currently requires that the `name` be updated in a `PUT`. For instance, I've tried the following example requests:

Given the following workspace record:
```json
{"name": "foo", "description": "foo workspace", "uuid": "1234"}
```

- `PUT` request `/api/rbac/v2/workspaces/1234/` with payload `{"name": "foo", "description": "UPDATED"}` I get back `{"errors":[{"detail":"Workspace already exist in tenant","source":"workspace","status":"400"}]}` -- it's checking uniqueness when I don't want/need to update the name, but only want to update description here
- `PUT` request `/api/rbac/v2/workspaces/1234/` with payload `{"description": "UPDATED"}` I get back `{"errors":[{"detail":"This field is required.","source":"name","status":"400"}]}` -- should actually be a `PATCH` to update only certain fields, so this one probably makes sense.

I'd expect

rbac/management/workspace/view.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def retrieve(self, request, *args, **kwargs):
+        """Get a workspace."""
+        return super().retrieve(request=request, args=args, kwargs=kwargs)
+
+    def destroy(self, request, *args, **kwargs):
+        """Delete a workspace."""
+        return super().destroy(request=request, args=args, kwargs=kwargs)
+
+    def update(self, request, *args, **kwargs):
```
coderbydesign, line 70: Should we support `PATCH` with `partial_update`?

rbac/management/workspace/view.py
coderbydesign: @lpichler curious if we have a ticket for this now or if we still need to create one?

rbac/rbac/urls.py
```diff
@@ -37,6 +37,7 @@
 urlpatterns = [
     re_path(r"^{}v1/".format(API_PATH_PREFIX), include("api.urls")),
     re_path(r"^{}v1/".format(API_PATH_PREFIX), include("management.urls")),
+    re_path(r"^{}v2/".format(API_PATH_PREFIX), include("management.v2_urls")),
```
coderbydesign, line 40: Out of curiosity, were you involved in some of the initial discussion around standards for versioning APIs on the platform? I thought so, but couldn't remember where that all landed.

rbac/management/workspace/view.py
lpichler: @coderbydesign I created ticket [here](https://issues.redhat.com/browse/RHCLOUD-33934). I put it into same epic but I am not sure whether it is good place. At this moment I created ticket as SPIKE.(we can change once details around schema will be clear )

Thanks for note!

rbac/management/workspace/view.py
lpichler: Done 👍

rbac/rbac/urls.py
lpichler: I couldn't remember as well and I also did not find anything regard this. Let me know if you will find something!

rbac/management/migrations/0044_workspace_workspace_unique_workspace_name_per_tenant.py
lpichler: parent column added - uniqueness is based on combination: (name, parent, tenant)

docs/source/specs/typespec/main.tsp
lpichler: Done 👍

docs/source/specs/typespec/main.tsp
lpichler: Done 👍

docs/source/specs/typespec/main.tsp
lpichler: - I removed list endpoint now.

@coderbydesign 
Good point! Consumers will need to change how they will be call v2 endpoints than they did it before.
As those endpoints are new - I think that it will not break anything.

docs/source/specs/typespec/main.tsp
lpichler: list action removed 👍

docs/source/specs/typespec/main.tsp
lpichler: changed according to other comments 👍

docs/source/specs/typespec/main.tsp
lpichler: Added 👍

docs/source/specs/typespec/main.tsp
lpichler: Done 👍

lpichler:
@coderbydesign @alechenninger I addressed all your comments, main things:
- reworked errors in types file (also backend API changes were added to fit it)
- add parent columns, actions in  backend API changes, validations added
- added PATCH method, PUT fixes added
- ...

rbac/management/workspace/view.py
coderbydesign: That makes sense, thanks for logging this!

tests/management/workspace/test_view.py
coderbydesign: 😄

coderbydesign (approved):
Thanks for the conversation here, @lpichler and @alechenninger! Approving on my end based on comments and feedback.

docs/source/specs/typespec/main.tsp
```diff
@@ -0,0 +1,170 @@
...
+    @doc("2019-01-21T17:32:28Z")
+    modified: offsetDateTime;
+}
+
+@format("uuid")
+scalar UUID extends string;
+
+namespace Problems {
+    enum ProblemType {
+        InsufficientPermission:"http://project-kessel.github.io",
```
alechenninger, line 35: > http://project-kessel.github.io

This should be unique per problem type as I understand it.

docs/source/specs/typespec/main.tsp
```diff
@@ -0,0 +1,170 @@
...
+    @error
+    model Problem404 extends ProblemDetails<Status = 404>{
+        detail: "Not found"
+    }
+    
+    alias Problem500 = ProblemDetails<Status = 500>;
+
+    @error
+    model Problem400 extends ProblemDetails<Status = 400>{
+        detail: "JSON parse error - Expecting value: line 1 column 1 (char 0)",
```
alechenninger, line 70: This detail value I think is trying to say that in the schema, this "detail" field will always have that value as the string, hence it generates OpenAPI like

```
      properties:
        detail:
          type: string
          enum:
            - 'JSON parse error - Expecting value: line 1 column 1 (char 0)'
```

That's probably not what we want. It looks like maybe you're trying to annotate an example? Not sure how to do that.

docs/source/specs/typespec/main.tsp
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    alias Problem401 = ProblemDetails<Status = 401>;
+
+    @error
+    model Problem404 extends ProblemDetails<Status = 404>{
+        detail: "Not found"
+    }
+    
+    alias Problem500 = ProblemDetails<Status = 500>;
```
alechenninger, line 66: I think these all need ProblemType's, too, but I suggest we add a follow up task to enumerate the error types as we go.

docs/source/specs/typespec/main.tsp
```diff
@@ -0,0 +1,170 @@
...
+        @key uuid: UUID;
+        parent?: UUID;
+        ...BasicWorkspace;
+        ...Timestamps;
+    }
+
+    alias CreateWorkspaceRequest = BasicWorkspace;
+
+    model CreateWorkspaceResponse extends BasicWorkspace{
+        @statusCode _: 204;
```
alechenninger, line 105: 204 means "No Content" so I think you want a 201 here.
```suggestion
        @statusCode _: 201;
```

docs/source/specs/typespec/main.tsp
```diff
@@ -0,0 +1,170 @@
...
+
+    model GetWorkspaceResponse extends BasicWorkspace{
+        @statusCode _: 200;
+    }
+
+    model PutWorkspaceRequest {
+        @doc("Workspace A")
+        name: string = "Workspace A";
+        @doc("Description of Workspace A")
+        description: string = "Description of Workspace A";
```
alechenninger, line 116:
```suggestion
        description?: string = "Description of Workspace A";
```

Should be optional?

docs/source/specs/typespec/main.tsp
```diff
@@ -0,0 +1,170 @@
...
+
+    model PatchWorkspaceRequest {
+        @doc("Workspace A")
+        name?: string = "Workspace A";
+        @doc("Description of Workspace A")
+        description?: string = "Description of Workspace A";
+    }
+
+    model PutWorkspaceResponse extends BasicWorkspace{
+        @statusCode _: 201;
```
alechenninger, line 127: Since this is an update, I think you want
```suggestion
        @statusCode _: 200;
```

docs/source/specs/typespec/main.tsp
```diff
@@ -0,0 +1,170 @@
...
+        @doc("Unique identification")
+        @path uuid: UUID;
+        @body body: PutWorkspaceRequest
+    ): PutWorkspaceResponse | Problems.CommonProblems | Problems.Problem400AlreadyExists ;
+
+    @patch op patch(
+        @doc("Unique identification")
+        @path uuid: UUID;
+        @body body: PutWorkspaceRequest
+    ): PatchWorkspaceResponse | Problems.CommonProblems | Problems.Problem400AlreadyExists ;
```
alechenninger, line 162: IMO I would just omit this API, there isn't a UI for it that I know of. Not a big deal tho.

rbac/management/utils.py
```diff
@@ -337,3 +337,33 @@ def get_admin_from_proxy(username, request):
...
+    path_prefix = os.getenv("API_PATH_PREFIX", "api/")
+    if path_prefix != "":
+        if path_prefix.startswith("/"):
+            path_prefix = path_prefix[1:]
+        if not path_prefix.endswith("/"):
+            path_prefix = path_prefix + "/"
+    return path_prefix
+
+
+def v2response_error_from_errors(errors, exc=None, context=None):
```
alechenninger, line 353: Can't quite tell but it doesn't look like the "type" field is being added to the error response ever?

docs/source/specs/typespec/main.tsp
alechenninger:
```suggestion
        InsufficientPermission:"http://project-kessel.org/problems/insufficient-permission",
```

docs/source/specs/typespec/main.tsp
```diff
@@ -0,0 +1,170 @@
...
+        ...Timestamps;
+    }
+
+    alias CreateWorkspaceRequest = BasicWorkspace;
+
+    model CreateWorkspaceResponse extends BasicWorkspace{
+        @statusCode _: 204;
+    }
+
+    model GetWorkspaceResponse extends BasicWorkspace{
```
alechenninger, line 108: nit: the request/response objects are not always consistent with whether they are using the HTTP verb or the operation name. I would go with the operation name (so CreateWorkspace*, ReadWorkspace*, UpdateWorkspace*, etc).

rbac/management/workspace/model.py
```diff
@@ -0,0 +1,44 @@
...
+
+from api.models import TenantAwareModel
+
+
+class Workspace(TenantAwareModel):
+    """A workspace."""
+
+    name = models.CharField(max_length=100)
+    uuid = models.UUIDField(default=uuid4, editable=False, unique=True, null=False)
+    parent = models.UUIDField(null=True, blank=True, editable=True)
```
alechenninger, line 32: Still TBD what we do with queries and if we need this to be duplicated in both services but we can come back to that later.

rbac/management/workspace/model.py
```diff
@@ -0,0 +1,44 @@
...
+from django.utils import timezone
+from management.rbac_fields import AutoDateTimeField
+
+from api.models import TenantAwareModel
+
+
+class Workspace(TenantAwareModel):
+    """A workspace."""
+
+    name = models.CharField(max_length=100)
```
alechenninger, line 30: Was this was meant to be 255 max to match HBI?

rbac/management/workspace/model.py
```diff
@@ -0,0 +1,44 @@
...
+    uuid = models.UUIDField(default=uuid4, editable=False, unique=True, null=False)
+    parent = models.UUIDField(null=True, blank=True, editable=True)
+    description = models.TextField(null=True)
```
alechenninger, line 33: There should probably be a max length here too? (we don't want a 50mb zip base64 encoded as a description ;P)

rbac/management/workspace/serializer.py
```diff
@@ -0,0 +1,45 @@
...
+"""Serializer for workspace management."""
+from rest_framework import serializers
+
+from .model import Workspace
+
+
+class WorkspaceSerializer(serializers.ModelSerializer):
+    """Serializer for the Workspace model."""
+
+    name = serializers.CharField(required=False, max_length=150)
```
alechenninger, line 27: Should these constraints match those on the model?

rbac/management/workspace/view.py
```diff
@@ -0,0 +1,144 @@
...
+        """Get a workspace."""
+        return super().retrieve(request=request, args=args, kwargs=kwargs)
+
+    def destroy(self, request, *args, **kwargs):
+        """Delete a workspace."""
+        instance = self.get_object()
+        if Workspace.objects.filter(parent=instance.uuid, tenant=instance.tenant).exists():
+            message = "Unable to delete due to workspace dependencies"
+            error = {"workspace": [_(message)]}
+            raise serializers.ValidationError(error)
```
alechenninger, line 71: This will need to get much smarter in the future re: assets but fine as a basic starting point

alechenninger (commented):
Looks great overall, a few minor things we should be able to touch up in collab.

docs/source/specs/typespec/main.tsp
lpichler: yes `@example` - finally they[ released it](https://typespec.io/docs/release-notes/release-2024-07-16) - it was not here.  

I put this into case example.

docs/source/specs/typespec/main.tsp
lpichler: I will create ticket for it.

docs/source/specs/typespec/main.tsp
lpichler: No, PUT contains entire resource , PATCH can contain partial data.

rbac/management/workspace/view.py
lpichler: yes, I believe we need ticket for it.

rbac/management/workspace/model.py
lpichler: Done 👍

docs/source/specs/typespec/main.tsp
```diff
@@ -0,0 +1,171 @@
...
+    model Problem404 extends ProblemDetails<Status = 404>{
+        detail: "Not found"
+    }
+    
+    alias Problem500 = ProblemDetails<Status = 500>;
+
+    @error
+    model Problem400 extends ProblemDetails<Status = 400>{
+        @example("JSON parse error - Expecting value: line 1 column 1 (char 0)")
+        detail: string = "The request payload contains invalid syntax.",
```
alechenninger, line 71: Swap detail for title in Problem400 and Problem404

docs/source/specs/typespec/main.tsp
```diff
@@ -0,0 +1,171 @@
...
+    @put op update(
+        @doc("Unique identification")
+        @path uuid: UUID;
+        @body body: PutWorkspaceRequest
+    ): PutWorkspaceResponse | Problems.CommonProblems | Problems.Problem400AlreadyExists ;
+
+    @patch op patch(
+        @doc("Unique identification")
+        @path uuid: UUID;
+        @body body: PutWorkspaceRequest
```
alechenninger, line 162:
```suggestion
        @body body: PatchWorkspaceRequest
```

rbac/management/utils.py
lpichler: I will create ticket for it.

docs/source/specs/typespec/main.tsp
lpichler: Done 👍

docs/source/specs/typespec/main.tsp
lpichler: Done 👍

rbac/management/workspace/model.py
lpichler: Done 👍 Changed to 255 and to varchar

docs/source/specs/typespec/main.tsp
lpichler: Done 👍

docs/source/specs/typespec/main.tsp
lpichler: https://issues.redhat.com/browse/RHCLOUD-34229

docs/source/specs/typespec/main.tsp
lpichler: https://issues.redhat.com/browse/RHCLOUD-34229

docs/source/specs/typespec/main.tsp
lpichler: https://issues.redhat.com/browse/RHCLOUD-34230

rbac/management/utils.py
lpichler: https://issues.redhat.com/browse/RHCLOUD-34231

rbac/management/workspace/serializer.py
lpichler: Done 👍

rbac/management/workspace/view.py
lpichler: https://issues.redhat.com/browse/RHCLOUD-34232

docs/source/specs/typespec/main.tsp
lpichler: Done 👍 Titles were added

alechenninger:
LGTM. Follow ups noted in slack. Thanks for your patience on this one.

alechenninger (approved):

PR #1136, opened 2024-07-02
Consume gRPC python client

astrozzc:
-----END CERTIFICATE-----

## Link(s) to Jira
-

The what, why and how.

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


rbac/migration_tool/utils.py
```diff
@@ -0,0 +1,30 @@
...
+import grpc
+from django.conf import settings
+from relations.v0 import common_pb2
+from relations.v0 import relation_tuples_pb2
+from relations.v0 import relation_tuples_pb2_grpc
+
+
+def create_relationship(resource_name, resource_id, subject_name, subject_id, relation):
+    """Create a relationship between a resource and a subject."""
+    return common_pb2.Relationship(
```
lpichler, line 11: There is support for validation at proto level - and thanks to that you can validate request before request is sent

There is example:
https://github.com/project-kessel/relations-client-python/blob/main/examples/client_validations.py#L75C20-L91C51
(Unfortunately you need to validate each part of request separately as is in example - validations did not work for whole request)

This could help catch error like empty data sooner.

rbac/migration_tool/utils.py
```diff
@@ -0,0 +1,30 @@
...
+    """Write relationships to the relation API server."""
+    with grpc.insecure_channel(settings.RELATION_API_SERVER) as channel:
+        stub = relation_tuples_pb2_grpc.KesselTupleServiceStub(channel)
+
+        request = relation_tuples_pb2.CreateTuplesRequest(
+            upsert=True,
+            tuples=relationships,
+        )
+
+    return stub.CreateTuples(request)
```
lpichler, line 30: I believe we can add some try catch block to catch errors.

@vbelchio can you show example how do you handle errors  for gRPC requests ?
We were discussing and I can't find link to that discuss.
I would prefer to see error message from gRPC which in which attribute proper error message is, ....
(With that I will also add such examples in repo client)

rbac/migration_tool/migrate.py
```diff
@@ -119,9 +105,10 @@ def migrate_role(role: Role):
     v1_to_v2_mapping = shared_system_role_replicated_role_bindings_v1_to_v2_mapping
     permissioned_role_migrator = Migrator(v1_to_v2_mapping)
     v2_roles = [v2_role for v2_role in permissioned_role_migrator.migrate_v1_roles(v1_role)]
-    spicedb_rel_summary = spicedb_relationships(frozenset(v2_roles))
-    for rel in spicedb_rel_summary:
+    relationships = spicedb_relationships(frozenset(v2_roles))
+    for rel in relationships:
         logger.info(stringify_spicedb_relationship(rel))
+    # write_relationships(relationships)
```
lpichler, line 111: 1. Maybe we could have environment variable to be enabled/disable - or maybe rather add parameter in request ? (or both :) )

2. If we will run it for all orgs all - it would lot of relations and I curious what would happen :) and I would definitely try it :) 

We can also implement some mechanisms like batching with retry logic.

3. Maybe it would be great to log how many relation are we going to write.

rbac/migration_tool/utils.py
astrozzc: Updated

rbac/migration_tool/utils.py
astrozzc: Updated

rbac/migration_tool/migrate.py
astrozzc: Updated with a parameter, which seems more convinient

lpichler:
@astrozzc this is awesome, can you remove `bonfire_env` folder ?  🙏

lpichler (approved):

PR #1137, opened 2024-07-03
Remove Principal Check 

Ellen-Yi-Dong:
## Link(s) to Jira
-https://issues.redhat.com/browse/RHCLOUD-33727

The what, why and how.
Need to remove principal check in order to make sure that the PR checks work and is in relation to the updates that were made by performance team 

"We need to determine if the principal check in the audit log model at [this location](https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/audit_log/model.py#L104) is necessary.

This check causes issues in the development environment or when custom data is populated, particularly when the user is not yet created in the database.

 

Current Problem:

    The principal check leads to issues when the performance team runs their script, which first creates roles and then principals. This sequence causes errors as the principal is not found.
        Alternative is to discuss with the performance team to understand the sequence of their script and explore the possibility of changing the order to create principals before roles.
    There is a possibility that the principal existence check is already being handled elsewhere in the codebase.

 

Proposed Solution:

Consider bypassing the principal check by directly using the username and ID from the request." 
## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/management/audit_log/model.py
+++ b/rbac/management/audit_log/model.py
@@ -98,7 +98,7 @@ class AuditLog(TenantAwareModel):
     def log_create(self, request, resource):
         """Audit Log when a role or a group is created."""
         self.principal_id = request.user.user_id
-        self.principal_username = request.user.username 
+        self.principal_username = request.user.username
         self.resource_type = resource
 
         self.resource_id, resource_name = self.get_resource_item(resource, request)
--- a/rbac/management/audit_log/model.py
+++ b/rbac/management/audit_log/model.py
@@ -95,13 +95,10 @@ class AuditLog(TenantAwareModel):
             # TODO: update for permission related items
             return None
 
-        elif r_type == "principal":
-            principal_object = get_object_or_404(Principal, username=request.user.username, tenant=verify_tenant)
-            return principal_object.id, principal_object.username
-
     def log_create(self, request, resource):
         """Audit Log when a role or a group is created."""
-        self.principal_id, self.principal_username = self.get_resource_item("principal", request)
+        self.principal_id = request.user.user_id
+        self.principal_username = request.user.username 
         self.resource_type = resource
 
         self.resource_id, resource_name = self.get_resource_item(resource, request)


rbac/management/audit_log/model.py
```diff
@@ -95,13 +95,10 @@ def get_resource_item(self, r_type, request, *args, **kwargs):
...
             return None
 
-        elif r_type == "principal":
-            principal_object = get_object_or_404(Principal, username=request.user.username, tenant=verify_tenant)
-            return principal_object.id, principal_object.username
-
     def log_create(self, request, resource):
         """Audit Log when a role or a group is created."""
-        self.principal_id, self.principal_username = self.get_resource_item("principal", request)
+        self.principal_id = request.user.user_id
```
petracihalova, line 100: curious question, here the `get_resource_time` method is not called anymore ... do we still need then this part? but I dont have strong preferences ❤️ https://github.com/RedHatInsights/insights-rbac/blob/34cb1b4d3398d254a599675eb83f84cf40d36a79/rbac/management/audit_log/model.py#L98-L100

rbac/management/audit_log/model.py
```diff
@@ -95,13 +95,10 @@ def get_resource_item(self, r_type, request, *args, **kwargs):
             # TODO: update for permission related items
             return None
 
-        elif r_type == "principal":
-            principal_object = get_object_or_404(Principal, username=request.user.username, tenant=verify_tenant)
-            return principal_object.id, principal_object.username
-
     def log_create(self, request, resource):
```
vbelchio, line 98: @Ellen-Yi-Dong Checking the PR_check failure we have a constraint violation in the `management_auditlog`.
Using a fresh database, I tried to create a new role and got this error due to not have the user requester as a principal yet:
```
{
  "errors": [
    {
      "detail": "insert or update on table \"management_auditlog\" violates foreign key constraint \"management_auditlog_principal_id_512ce4d0_fk_managemen\"\nDETAIL:  Key (principal_id)=(51834776) is not present in table \"management_principal\".\n",
      "source": "role",
      "status": "400"
    }
  ]
}
```

Maybe we can check and add the user requester as a principal inside the `log_create` method.

vbelchio (changes requested):

PR #1141, opened 2024-07-17
set cpu and memory resources for the worker init container

petracihalova:
the rbac worker init container runs migrations and seeding jobs that are very cpu and memory intensive so we want to set resources separately for init container and for the worker container itself

image
https://grafana.stage.devshift.net/d/HXbU7G1Iz/aandm-resources-health?orgId=1&from=1721143635069&to=1721145097089&var-Datasource=crcp01ue1-prometheus


astrozzc (approved):
LGTM

PR #1142, opened 2024-07-17
update example value for cross account requests in open api spec

petracihalova:
`POST + PUT + PATCH /cross_account_requests` endpoints uses role display name in the json payloads (and if the role doesn't have a `display_name` then rbac [populates](https://github.com/RedHatInsights/insights-rbac/blob/5a71008586d0ef495dfbdcd98aea39589e762190/rbac/management/role/model.py#L62-L66) the `name` property

### DIFF
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -3718,7 +3718,7 @@
             "type": "array",
             "items": {
               "type": "string",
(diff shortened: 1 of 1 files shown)


vbelchio (approved):
LGTM
Thanks to update ;D

PR #1146, opened 2024-07-19
Set a finer control over destructive operations

petracihalova:
the goal is to have a finer control over destructive operations via internal endpoints and seeding

now the `RBAC_DESTRUCTIVE_API_ENABLED_UNTIL` param will control destructive operations for internal endpoints
and `RBAC_DESTRUCTIVE_SEEDING_ENABLED_UNTIL` param will control destructive operations for seeding job

motivation:
it can happen that we allow destructive operations for the removal of one specific role via internal request and in the meantime some event (rbac deploy, rbac config deploy, app-interface deploy yml update ....) will trigger pods reloading and that will run the seeding job with destructive operations, even though this was not our intention


astrozzc (approved):

lpichler:
Does this PR require corresponding changes in app interface ?

petracihalova:
> Does this PR require corresponding changes in app interface ?

yes we have `RBAC_DESTRUCTIVE_ENABLED_UNTIL` in [deploy-perf.yml](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/insights/rbac/deploy-perf.yml) and in [deploy-clowder.yml](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/insights/rbac/deploy-clowder.yml) and we should remove them or replace them by `RBAC_DESTRUCTIVE_API_ENABLED_UNTIL` and before we change it the default value from past will be used for new variables, so we should be fine here

lpichler (approved):

lpichler (approved):

PR #1153, opened 2024-08-01
[RHCLOUD-33738] Group principals username only

petracihalova:
## Link(s) to Jira
- [RHCLOUD-33738](https://issues.redhat.com/browse/RHCLOUD-33738)

we support the `username_only` parameter for the endpoint `GET /groups/{uuid}/principals/?username_only=true` to return only usernames for principals (in that case the call to IT is skipped)

and we should support the same for the service account based principals
`GET /groups/{uuid}/principals/?principal_type=service-account&username_only=true`

so in this case the call to IT will be skipped and we will return only service account usernames from RBAC database

## Local Testing
create a group with at least 1 service account
send the request `GET /groups/{uuid}/principals/?principal_type=service-account`
check the response contains whole service account object

then send the request `GET /groups/{uuid}/principals/?principal_type=service-account&username_only=true`
check the response contains only usernames

### DIFF
--- a/rbac/management/group/view.py
+++ b/rbac/management/group/view.py
@@ -787,13 +787,14 @@ class GroupViewSet(
                 # Get the "principal username" parameter.
                 options[PRINCIPAL_USERNAME_KEY] = request.query_params.get(PRINCIPAL_USERNAME_KEY)
 
+                # Validate the token only if username_only is false (default value)
+                if username_only == "false":
+                    token_validator = ITSSOTokenValidator()
+                    request.user.bearer_token = token_validator.validate_token(
+                        request=request,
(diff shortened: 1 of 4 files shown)


astrozzc (approved):
LGTM

petracihalova:
rebased

coderbydesign (approved):
Functionally works as expected locally, and I see we already have this in the [internal spec](https://github.com/RedHatInsights/insights-rbac/blob/af1685ad3644c622677dab9e30ce4ecfe98d930f/rbac/internal/specs/openapi.json#L706). 🌴

Ellen-Yi-Dong (approved):
LGTM!

petracihalova:
rebased

PR #1155, opened 2024-08-05
[RHCLOUD-32905] validate and get the scope query params

petracihalova:
## Link(s) to Jira
- [RHCLOUD-32905](https://issues.redhat.com/browse/RHCLOUD-32905)

this PR adds the validation for the `scope` query param, the valid values are `[org_id, principal] `

**context:**
we support the `scope` query param in several endpoints for filtering the resource by scope with default value `org_id` and another option `principal`
- `GET /groups/`
- `GET /roles/`
- `GET /policies/`

when org admin or RBAC admin requests `GET /groups/` the default `scope=org_id` is used and all groups under a org id are returned
otherwise with `GET /groups/?scope=principal` only groups the user belongs in are returned
for the invalid value `GET /groups/?scope=foo` the API returns an validation error
```
{
    "errors": [
        {
            "detail": "scope query parameter value 'foo' is invalid. ['org_id', 'principal'] are valid inputs.",
            "source": "detail",
            "status": "400"
        }
    ]
}
```

## Local Testing

- as org admin send `GET /groups/` request witnout additional query params and check that all groups under a tenant are returned
- as org admin send `GET /groups/?scope=org_id` request and check that all groups under a tenant are returned (same as the case above)
- as org admin send `GET /groups/?scope=principal` request and check that only groups the principal belongs in are returned
- as org admin send `GET /groups/?scope=foo` request and check that API returns validation error with status code 400

### DIFF
--- a/rbac/management/querysets.py
+++ b/rbac/management/querysets.py
@@ -36,6 +36,7 @@ from management.utils import (
     policies_for_principal,
     queryset_by_id,
     roles_for_principal,
+    validate_and_get_key,
 )
 from rest_framework import permissions, serializers
 from rest_framework.request import Request
@@ -97,7 +98,7 @@ def _gather_group_querysets(request, args, kwargs):
     """Decide which groups to provide for request."""
     username = request.query_params.get("username")
 
-    scope = request.query_params.get(SCOPE_KEY, ORG_ID_SCOPE)
+    scope = validate_and_get_key(request.query_params, SCOPE_KEY, VALID_SCOPES, ORG_ID_SCOPE)
     if scope != ORG_ID_SCOPE and not username:
         return get_object_principal_queryset(request, scope, Group)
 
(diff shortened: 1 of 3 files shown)


rbac/management/querysets.py
```diff
@@ -97,7 +98,7 @@ def _gather_group_querysets(request, args, kwargs):
     """Decide which groups to provide for request."""
     username = request.query_params.get("username")
 
-    scope = request.query_params.get(SCOPE_KEY, ORG_ID_SCOPE)
+    scope = validate_and_get_key(request.query_params, SCOPE_KEY, VALID_SCOPES, ORG_ID_SCOPE, required=False)
```
astrozzc, line 101: Since you provided the default value, you can get rid off that required=False

astrozzc (approved):
Overall LGTM!

rbac/management/querysets.py
petracihalova: thank you Jay, it is updated

Ellen-Yi-Dong (approved):
LGTM!

PR #1159, opened 2024-08-05
[RHCLOUD-31875] Update the group permissions

petracihalova:
## Link(s) to Jira
- [RHCLOUD-31875](https://issues.redhat.com/browse/RHCLOUD-31875)

the non admin (non org admin and non rbac admin) user can retrieve the group detail via `GET /groups/{uuid}/?scope=principal`
but only org admin or rbac admin can do that
the `scope` query param is not valid param for this endpoint and should be ignored

**how to reproduce it:**
1. find or create new user based principal (non org admin, non rbac admin)
2. create new group (without roles) and add the user from (1) into this group
3. with the user from step (1) in the identity header send the `GET /groups/{uuid}/` request with the group uuid created in the previous step (2) -> expected 403 response
4. with the user from step (1) in the identity header send the request with principal scope `GET /groups/{uuid}/?scope=principal` with the group uuid created in the step (2) -> expected 403 response but you get 200 response with group details

403 Response example
```
{
    "errors": [
        {
            "detail": "You do not have permission to perform this action.",
            "source": "detail",
            "status": "403"
        }
    ]
}
```

the issue exists and should be fixed for the service account base principal too

**the broader context:**
the non org admin and non rbac admin have an access to Group APIs only to list groups the principal belongs in
- with `scope` query param with principal value - `GET /groups/?scope=principal `
- with `usename` query param when the username is principal's own username - `GET /groups/?username=<value> `

## Local Testing

### DIFF
--- a/rbac/management/permissions/group_access.py
+++ b/rbac/management/permissions/group_access.py
@@ -38,11 +38,12 @@ class GroupAccessPermission(permissions.BasePermission):
             group_read = request.user.access.get("group", {}).get("read", [])
             if group_read:
                 return True
-            username = request.query_params.get("username")
-            if username:
-                return username == request.user.username
-            if not username and is_scope_principal(request):
-                return True
+            if view.basename == "group" and view.action == "list":
+                username = request.query_params.get("username")
+                if username:
+                    return username == request.user.username
+                if not username and is_scope_principal(request):
+                    return True
         else:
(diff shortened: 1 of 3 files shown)


astrozzc (approved):

Ellen-Yi-Dong (approved):

petracihalova:
rebased

PR #1163, opened 2024-08-05
[RHCLOUD-23970] add the system query param in the internal open api spec for the principals' groups endpoint

petracihalova:
## Link(s) to Jira
- [RHCLOUD-23970](https://issues.redhat.com/browse/RHCLOUD-23970)

internal endpoint: `GET /integrations/tenant/<orgId>/principal/<username>/groups/ `
the `system` parameter is already working for this endpoint, we need just to update the internal openapi spec

### DIFF
--- a/rbac/internal/specs/openapi.json
+++ b/rbac/internal/specs/openapi.json
@@ -490,6 +490,15 @@
                 "modified"
               ]
             }
+          },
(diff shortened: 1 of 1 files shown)


vbelchio (approved):
LGTM

PR #1170, opened 2024-08-15
Catch error if already subscribed

astrozzc:
https://glitchtip.devshift.net/insights/issues/341451


petracihalova (approved):
awesome, thank you

PR #1172, opened 2024-08-20
[WIP] Dual Write: Generate replication event for Role endpoints

lpichler:
-


rbac/management/role/view.py
```diff
@@ -371,7 +394,7 @@ def update(self, request, *args, **kwargs):
         @apiParam (Path) {String} id Role unique identifier
 
         @apiParam (Request Body) {String} name Role name
-        @apiParam (Request Body) {Array} access Access definition
+        @apiParam (Request Body) {ArRray} access Access definition
```
lennysgarage, line 397: Might be a small nitpick not sure if the docstrings affect things here but should `ArRray` just be `Array`?

rbac/management/role/model.py
```diff
@@ -138,6 +139,8 @@ class BindingMapping(models.Model):
 
     id = models.UUIDField(default=uuid4, primary_key=True)
     v1_role = models.ForeignKey(Role, on_delete=models.CASCADE)
+    v2_role = models.ForeignKey(V2Role, on_delete=models.CASCADE)
+    permissions = ArrayField(models.CharField(max_length=200), blank=True, null=True)
```
alechenninger, line 143: This current model has a race condition which could result in permanently inconsistent permissions data between RBAC and Relations.

**NOTE**: We can address this in a follow-up PR if you like.

The issue can be seen if we look at the SQL. Imagine this is from a transaction TX1, but concurrently there is a transaction TX2 that updates the same role.

```sql
[2024-08-22 11:23:22,461] DEBUG: (0.001) SELECT "api_tenant"."id", "api_tenant"."ready", "api_tenant"."tenant_name", "api_tenant"."account_id", "api_tenant"."org_id" FROM "api_tenant" WHERE "api_tenant"."tenant_name" = 'public' LIMIT 21; args=('public',); alias=default
[2024-08-22 11:23:22,464] DEBUG: (0.001) SELECT "management_role"."id", "management_role"."tenant_id", "management_role"."uuid", "management_role"."name", "management_role"."display_name", "management_role"."description", "management_role"."system", "management_role"."platform_default", "management_role"."version", "management_role"."created", "management_role"."modified", "management_role"."admin_default", COUNT(DISTINCT "management_policy_roles"."policy_id") AS "policyCount", COUNT(DISTINCT "management_access"."id") AS "accessCount" FROM "management_role" LEFT OUTER JOIN "management_policy_roles" ON ("management_role"."id" = "management_policy_roles"."role_id") LEFT OUTER JOIN "management_access" ON ("management_role"."id" = "management_access"."role_id") WHERE ("management_role"."tenant_id" IN (2, 1) AND "management_role"."uuid" = '42c4106a-e1be-4b14-8ab0-7a74bb6df67b'::uuid) GROUP BY "management_role"."id" LIMIT 21; args=(2, 1, UUID('42c4106a-e1be-4b14-8ab0-7a74bb6df67b')); alias=default
[2024-08-22 11:23:22,465] DEBUG: (0.000) SELECT "management_access"."id", "management_access"."tenant_id", "management_access"."permission_id", "management_access"."role_id" FROM "management_access" WHERE "management_access"."role_id" IN (99); args=(99,); alias=default
[2024-08-22 11:23:22,467] DEBUG: (0.002) SELECT "management_v2role"."id", "management_v2role"."is_system" FROM "management_v2role" WHERE "management_v2role"."id" IN (SELECT U0."v2_role_id" FROM "management_bindingmapping" U0 WHERE U0."v1_role_id" = 99); args=(99,); alias=default
[2024-08-22 11:23:22,469] DEBUG: (0.002) DELETE FROM "management_rolemapping" WHERE "management_rolemapping"."v2_role_id" IN ('2c983c5d-3763-4054-9396-b42666af7847'::uuid, '1e67063a-e3ff-465a-b042-907f6d830594'::uuid); args=(UUID('2c983c5d-3763-4054-9396-b42666af7847'), UUID('1e67063a-e3ff-465a-b042-907f6d830594')); alias=default
[2024-08-22 11:23:22,470] DEBUG: (0.001) DELETE FROM "management_bindingmapping" WHERE "management_bindingmapping"."v2_role_id" IN ('2c983c5d-3763-4054-9396-b42666af7847'::uuid, '1e67063a-e3ff-465a-b042-907f6d830594'::uuid); args=(UUID('2c983c5d-3763-4054-9396-b42666af7847'), UUID('1e67063a-e3ff-465a-b042-907f6d830594')); alias=default
[2024-08-22 11:23:22,471] DEBUG: (0.000) DELETE FROM "management_v2role" WHERE "management_v2role"."id" IN ('2c983c5d-3763-4054-9396-b42666af7847'::uuid, '1e67063a-e3ff-465a-b042-907f6d830594'::uuid); args=(UUID('2c983c5d-3763-4054-9396-b42666af7847'), UUID('1e67063a-e3ff-465a-b042-907f6d830594')); alias=default

TX2 commit -> adds new rows to the mapping

[2024-08-22 11:23:22,472] DEBUG: (0.000) SELECT "management_resourcedefinition"."id", "management_resourcedefinition"."tenant_id", "management_resourcedefinition"."attributeFilter", "management_resourcedefinition"."access_id" FROM "management_resourcedefinition" WHERE "management_resourcedefinition"."access_id" = 527; args=(527,); alias=default
[2024-08-22 11:23:22,472] DEBUG: (0.000) SELECT "management_permission"."id", "management_permission"."tenant_id", "management_permission"."application", "management_permission"."resource_type", "management_permission"."verb", "management_permission"."permission", "management_permission"."description" FROM "management_permission" WHERE "management_permission"."id" = 187 LIMIT 21; args=(187,); alias=default
[2024-08-22 11:23:22,473] DEBUG: (0.000) SELECT "management_resourcedefinition"."id", "management_resourcedefinition"."tenant_id", "management_resourcedefinition"."attributeFilter", "management_resourcedefinition"."access_id" FROM "management_resourcedefinition" WHERE "management_resourcedefinition"."access_id" = 528; args=(528,); alias=default
[2024-08-22 11:23:22,474] DEBUG: (0.000) SELECT "management_permission"."id", "management_permission"."tenant_id", "management_permission"."application", "management_permission"."resource_type", "management_permission"."verb", "management_permission"."permission", "management_permission"."description" FROM "management_permission" WHERE "management_permission"."id" = 186 LIMIT 21; args=(186,); alias=default
[2024-08-22 11:23:22,476] DEBUG: (0.001) SELECT "management_policy"."id", "management_policy"."tenant_id", "management_policy"."uuid", "management_policy"."name", "management_policy"."description", "management_policy"."group_id", "management_policy"."system", "management_policy"."created", "management_policy"."modified" FROM "management_policy" INNER JOIN "management_policy_roles" ON ("management_policy"."id" = "management_policy_roles"."policy_id") WHERE "management_policy_roles"."role_id" = 99 ORDER BY "management_policy"."name" ASC, "management_policy"."modified" ASC; args=(99,); alias=default
[2024-08-22 11:23:22,478] DEBUG: (0.002) SELECT "management_v2role"."id", "management_v2role"."is_system" FROM "management_v2role" WHERE ("management_v2role"."id" = '799ab52e-6249-473f-a256-c7377391ce13'::uuid AND NOT "management_v2role"."is_system") LIMIT 21; args=(UUID('799ab52e-6249-473f-a256-c7377391ce13'),); alias=default
[2024-08-22 11:23:22,479] DEBUG: (0.000) SAVEPOINT "s12935360512_x2"; args=None; alias=default
[2024-08-22 11:23:22,480] DEBUG: (0.001) INSERT INTO "management_v2role" ("id", "is_system") VALUES ('799ab52e-6249-473f-a256-c7377391ce13'::uuid, false); args=(UUID('799ab52e-6249-473f-a256-c7377391ce13'), False); alias=default
[2024-08-22 11:23:22,480] DEBUG: (0.000) RELEASE SAVEPOINT "s12935360512_x2"; args=None; alias=default
[2024-08-22 11:23:22,482] DEBUG: (0.001) SELECT "management_rolemapping"."v1_role_id" FROM "management_rolemapping" WHERE ("management_rolemapping"."v1_role_id" IN (99) AND "management_rolemapping"."v2_role_id" = '799ab52e-6249-473f-a256-c7377391ce13'::uuid); args=(99, UUID('799ab52e-6249-473f-a256-c7377391ce13')); alias=default
[2024-08-22 11:23:22,483] DEBUG: (0.001) INSERT INTO "management_rolemapping" ("v1_role_id", "v2_role_id") VALUES (99, '799ab52e-6249-473f-a256-c7377391ce13'::uuid) RETURNING "management_rolemapping"."id"; args=(99, UUID('799ab52e-6249-473f-a256-c7377391ce13')); alias=default
[2024-08-22 11:23:22,484] DEBUG: (0.001) INSERT INTO "management_bindingmapping" ("id", "v1_role_id", "v2_role_id", "permissions") VALUES ('14afe2a7-ebbb-4a08-b1d9-76748e511822'::uuid, 99, '799ab52e-6249-473f-a256-c7377391ce13'::uuid, ARRAY['inventory_hosts_read']::varchar(200)[]); args=(UUID('14afe2a7-ebbb-4a08-b1d9-76748e511822'), 99, UUID('799ab52e-6249-473f-a256-c7377391ce13'), ['inventory_hosts_read']); alias=default
[2024-08-22 11:23:22,496] DEBUG: (0.011) SELECT "management_v2role"."id", "management_v2role"."is_system" FROM "management_v2role" WHERE ("management_v2role"."id" = '873cc590-4d96-4e2d-9844-77b007111167'::uuid AND NOT "management_v2role"."is_system") LIMIT 21; args=(UUID('873cc590-4d96-4e2d-9844-77b007111167'),); alias=default
[2024-08-22 11:23:22,496] DEBUG: (0.000) SAVEPOINT "s12935360512_x3"; args=None; alias=default
[2024-08-22 11:23:22,497] DEBUG: (0.000) INSERT INTO "management_v2role" ("id", "is_system") VALUES ('873cc590-4d96-4e2d-9844-77b007111167'::uuid, false); args=(UUID('873cc590-4d96-4e2d-9844-77b007111167'), False); alias=default
[2024-08-22 11:23:22,497] DEBUG: (0.000) RELEASE SAVEPOINT "s12935360512_x3"; args=None; alias=default
[2024-08-22 11:23:22,498] DEBUG: (0.001) SELECT "management_rolemapping"."v1_role_id" FROM "management_rolemapping" WHERE ("management_rolemapping"."v1_role_id" IN (99) AND "management_rolemapping"."v2_role_id" = '873cc590-4d96-4e2d-9844-77b007111167'::uuid); args=(99, UUID('873cc590-4d96-4e2d-9844-77b007111167')); alias=default
[2024-08-22 11:23:22,499] DEBUG: (0.001) INSERT INTO "management_rolemapping" ("v1_role_id", "v2_role_id") VALUES (99, '873cc590-4d96-4e2d-9844-77b007111167'::uuid) RETURNING "management_rolemapping"."id"; args=(99, UUID('873cc590-4d96-4e2d-9844-77b007111167')); alias=default
[2024-08-22 11:23:22,500] DEBUG: (0.000) INSERT INTO "management_bindingmapping" ("id", "v1_role_id", "v2_role_id", "permissions") VALUES ('89db096d-a7a6-4f2e-b06b-1ab94dfd3da8'::uuid, 99, '873cc590-4d96-4e2d-9844-77b007111167'::uuid, ARRAY['inventory_hosts_write']::varchar(200)[]); args=(UUID('89db096d-a7a6-4f2e-b06b-1ab94dfd3da8'), 99, UUID('873cc590-4d96-4e2d-9844-77b007111167'), ['inventory_hosts_write']); alias=default
[2024-08-22 11:23:22,504] DEBUG: (0.004) COMMIT; args=None; alias=default

TX1 commits -> 
```

In this case, TX1 will commit and (when we have outbox events) publish an outbox event which removes and adds tuples, however it will miss the tuples that were added in TX2 which committed before TX1. This means the state in Relations API will reflect permissions that were actually removed according to the latest state in RBAC. This is a very bad correctness violation with security implications.

Ultimately this stems from a "[phantom read](https://www.postgresql.org/docs/current/transaction-iso.html#TRANSACTION-ISO)" phenomenon: the query in `[2024-08-22 11:23:22,467] DEBUG: (0.002) SELECT "management_v2role"."id", ...` reads one result set at one point in the transaction, but later on this would return a different set after the concurrent TX2 committed.

There are three ways I can think to fix this:

1. Change the isolation level for "Serializable" for role updating transactions. While Postgres's version of Repeatable Read does prevent phantom reads, I do not think it prevents them in a way that would help us here. What we need is a "predicate lock" (described in [the Serializable level](https://www.postgresql.org/docs/current/transaction-iso.html#XACT-SERIALIZABLE)) which does not allow a new mapping row for the same v1role to be inserted after we have selected the mapping rows for deletion.
2. Use a `JSONField` (https://docs.djangoproject.com/en/5.1/ref/models/fields/#jsonfield) for storing mappings with a single unique key on the v1role ID (a 1-1 relationship instead of 1-m) and lock this row with `select ... for update` (https://docs.djangoproject.com/en/5.1/ref/models/querysets/#select-for-update).
3. Use a `JSONField` (https://docs.djangoproject.com/en/5.1/ref/models/fields/#jsonfield) for storing mappings with a single unique key on the v1role ID and use the "Repeatable read" transaction isolation level. The use of a JSONField with a 1-1 relationship means that we turn the "phantom read" into a "lost update" problem and "lost update" is cheaper to prevent (prevented by Repeatable Read in postgres–not true for all databases.)

For all options, the main risks are performance and ensuring we have prevented any concurrency issues correctly. It's not clear to me which would be a better choice. We may want to just try one and see how it goes.

For Option 1, PG is very good at this due to its use of "snapshot isolation." It achieves serializability without blocking locks. And we know there would never be any correctness issue no matter what shape or code we wrote, and users could know when there was a concurrent modification. It still has an overhead cost though. Also, I couldn't find out how to enable this for _only_ the role updating transactions where it would be necessary. If we can figure that out, it might reduce the cost a bit more. One thing that could help would be to declare any "read only" transactions (e.g. the access endpoint) so PG can optimize them, however I'm not quite sure how to do that with Django. (I found this [forum thread](https://groups.google.com/g/django-developers/c/A2X6YMoZFdw/m/f9M6iTZOAgAJ) which might have a suggestion that works.)

For Option 2, it would involve a blocking lock. This probably has a similar performance cost due to locking the row but can at least be very targeted. However, because it blocks it's not always guaranteed to be better. It is also harder to use and guarantee there won't be issues (i.e. did we lock the right rows?)

For option 3, this would definitely perform better than option 1 since it aborts less. Like option 1 it does not involve blocking locks, and still detects the race condition. The question to me is if option 2 would be better anyway, since we know if TX2 in the example above commits, then TX1 will eventually abort in repeatable read. With option 2, TX2 would just wait until TX1 committed. One benefit of this option though is that, like option 1, at least a user can be alerted to the fact that there was a concurrent modification, whereas option 2 would hide this.

After thinking through all of this, I feel like the best option to start might be option 2 or 3 since they both use the same data schema. If one doesn't work out, we can try the other easily. Option 3 is perhaps the simplest because we just change one config flag and don't have to worry about getting the application code perfectly right. We could also try out the change (database isolation) in production ASAP without having to get the dual write code in. If it causes a performance issue or too many rollbacks, we could revert and try option 2.

I am open to starting with option 2 as well.

WDYT? cc @coderbydesign

rbac/management/role/relation_api_dual_write_handler.py
```diff
@@ -0,0 +1,156 @@
...
+            relations_to_remove.append(relationship_to_json(relation))
+
+        replication_event = {"relations_to_add": relations_to_add, "relations_to_remove": relations_to_remove}
+        return replication_event
+
+    def generate_replication_event_to_outbox(self, role):
+        """Generate replication event to outbox table."""
+        if not self.replication_enabled():
+            return
+        self.set_role(role)
```
alechenninger, line 131: Why does this method accept a `role` if the `role` is already passed when it is initialized? You could drop that and the `set_role` method I think, which makes this class less error prone. For example you could technically pass a `role` with a different tenant, and then `self.tenant_id` etc would be out of sync with `self.role`.

rbac/management/role/model.py
alechenninger: Also note that since most operations are performed in their own isolated transaction (autocommit / not inside `atomic()`) then I think the perf overhead of repeatable read (or serializable for that matter) should be negligible (however you will also not see any of the benefits of the increased isolation level in those cases–it will only matter to `atomic()` blocks).

rbac/management/role/relation_api_dual_write_handler.py
lpichler: I need new state of `role` but I can remove `set_role` method

rbac/management/role/model.py
lpichler: Option 2 was added.

rbac/management/role/view.py
```diff
@@ -418,15 +440,45 @@ def update(self, request, *args, **kwargs):
...
 
+    def update_with_relation_api_replication(self, request, *args, **kwargs):
+        """Update a role with replicating data into Relation API."""
+        try:
+            role = self.get_object()
+            dual_write_handler = RelationApiDualWriteHandler(role, "UPDATE")
+            dual_write_handler.generate_relations_from_current_state_of_role()
+            with transaction.atomic():
+                response = super().update(request=request, args=args, kwargs=kwargs)
+                dual_write_handler.generate_replication_event_to_outbox(self.get_object())
```
alechenninger, line 460: So there is another race here that could get us dead in the water. Basically, all of the `RelationApiDualWriteHandler` operations need to be done within a single transaction that is either row level locked for the Role, or uses repeatable read isolation.

Annotated code with what the problem is:

```python
            # This is outside a tx, so uses current snapshot and does not block other transactions
            role = self.get_object()
            dual_write_handler = RelationApiDualWriteHandler(role, "UPDATE")
            # At this point, Role state may be changed in DB,
            # and another outbox event could have been written with different role state
            # now unknown to this method
            dual_write_handler.generate_relations_from_current_state_of_role()
            with transaction.atomic():
                # Now we lock the Role, but too late–the Role has already changed
                response = super().update(request=request, args=args, kwargs=kwargs)
                # Now we generate mappings based on update, but we do not remove the right relations
                # because those relations were generating concurrently.
                dual_write_handler.generate_replication_event_to_outbox(self.get_object())
```

Fixing this is the same thing as we talked about before – we need to bring the whole thing into the same transaction and either (a) blocking lock the row or (b) use repeatable read.

We discussed a bit about (a) in slack if we want to go down that route but I'm okay either way

alechenninger: So this needs to be applied everywhere the dual write handler is used I think.

lpichler:
LGTM 👍

astrozzc (approved):

rbac/migration_tool/utils.py
```diff
@@ -91,6 +90,15 @@ def stringify_spicedb_relationship(rel: common_pb2.Relationship):
     )
 
 
+def relationship_to_json(rel):
```
merlante, line 93: Could we add
```
from google.protobuf.json_format import MessageToJson

json_rel = MessageToJson(rel)
```
(e.g. https://stackoverflow.com/questions/19734617/protobuf-to-json-in-python)
Instead?

This would standardise going to and from json.

alechenninger (approved):
IMO this is ready to ship

PR #1176, opened 2024-08-26
Don't create root workspace in database in migration tool

lpichler:
- existence of root workspace in RBAC doesn't have usage and it becomes prerequisite to enable [dual writes ](https://github.com/RedHatInsights/insights-rbac/pull/1172) 
  - this means that migration has to be done first before enabling dual writes

- it adds complexity: it is need to create root workspace when new organisation needs to be created rbac
- Name of root workspace could be determined from `org_id`  only which simplifies code and reduces issues which be caused by existence/non-existence of root workspace

### DIFF
--- a/rbac/migration_tool/migrate.py
+++ b/rbac/migration_tool/migrate.py
@@ -21,7 +21,6 @@ from typing import FrozenSet
 
 from django.conf import settings
 from management.role.model import BindingMapping, Role, V2Role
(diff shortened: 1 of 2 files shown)


coderbydesign (approved):
Approving based on the collab this morning, knowing we'll likely need to consider the IDs being used for all workspace relations moving forward.

PR #1179, opened 2024-08-29
[RHCLOUD-24782] Add field user_id to principal

astrozzc:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-24782

We need the uuid to create the relationships between users and group etc in AuthZ project

## Local Testing
Unit test

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


alechenninger (commented):
Nice! Looks pretty good overall I am just wondering about that ID choice in the group view.

I was looking around places Principal objects are created. I noticed cross-account request principals do not have user_ids set. Maybe we want to keep it that way, since these principals will be special cased anyway? This is a good example of why I think user IDs in Relations should have some "issuer" embedded in the ID. So that we can have users from different authorities in the same space. We can come back to this one.

rbac/management/group/view.py
```diff
@@ -428,7 +427,9 @@ def add_principals(self, group, principals, org_id=None):
             try:
                 principal = Principal.objects.get(username__iexact=username, tenant=tenant)
             except Principal.DoesNotExist:
-                principal = Principal.objects.create(username=username, tenant=tenant)
+                principal = Principal.objects.create(
+                    username=username, tenant=tenant, user_id=self.request.user.user_id
```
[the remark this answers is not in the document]
astrozzc: Oh, good catch! Right, the requester is different. Will fix this and add a unit test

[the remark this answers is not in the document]
astrozzc: Updated.
The cross account users are only created in this place: https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/api/cross_access/util.py#L65-L66
Will leave the user_id out for them

rbac/management/group/view.py
```diff
@@ -424,11 +423,12 @@ def add_principals(self, group, principals, org_id=None):
                 "errors": [{"detail": "User(s) {} not found.".format(users), "status": "404", "source": "principals"}],
             }
         for item in resp.get("data", []):
+            # cross-account request principals won't be in the resp from BOP since they don't exist
             username = item["username"]
             try:
                 principal = Principal.objects.get(username__iexact=username, tenant=tenant)
             except Principal.DoesNotExist:
-                principal = Principal.objects.create(username=username, tenant=tenant)
+                principal = Principal.objects.create(username=username, tenant=tenant, user_id=item["user_id"])
```
alechenninger, line 431: Will the data have `"user_id"` in it?

I was looking through the PrincipalProxy and the structure of the response in BOP. It looked like you might need to include an `options` dict with `"return_id"` set when calling `request_filtered_principals`.

astrozzc: Oops! Good catch! Somehow I was reading this the opposite way, default to True. https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/principal/proxy.py#L278 😅 My brain might stopped working at that time

rbac/management/utils.py
```diff
@@ -115,7 +115,7 @@ def get_principal(
         else:
             # Avoid possible race condition if the user was created while checking BOP
             principal, _ = Principal.objects.get_or_create(
-                username=username, tenant=tenant
+                username=username, tenant=tenant, user_id=request.user.user_id
```
coderbydesign, line 118: Will we need to ensure we backfill data first? Wondering if this would cause a false-negative lookup and end up creating duplicates in the case where:

- We have a principal with `username="Foo", user_id=null` because we've merged this but not backfilled `user_id` values
- This lookup gets called and now `reqeust.user.user_id` is populated, so the lookup might be on `username="Foo", user_id=123` which would create a new record?

astrozzc:
I removed the db_index, because it would slow down the object creation. I will add it back once we populate the users data into RBAC

lpichler (approved):

rbac/management/principal/model.py
```diff
@@ -26,14 +26,19 @@
 class Principal(TenantAwareModel):
     """A principal."""
 
+    class Types(models.TextChoices):
+        USER = "user", "User"
+        SERVICE_ACCOUNT = "service-account", "Service Account"
```
coderbydesign, line 31: Sort of a nit, but we could probably trim the values down on these to be 1-3 characters, which probably doesn't do a ton for optimization, but might not hurt? Like `SERVICE_ACCOUNT = "SA", "Service Account"` for instance, and then we could have a `max_length` on the column definition below. 🤷 Again probably not a big deal, but since the type is effectively an enum with a human-readable counterpart it might be worth it?

coderbydesign: Also, with this since we don't use it in a form or anything, we could probably just get away with:

```python
USER = "user"
SERVICE_ACCOUNT = "service-account"
```
We'd still have the `label` and `name` attributes based off the constant name:
```python
>>> Types.USER.__dict__
{'_value_': 'user', '_name_': 'USER', '__objclass__': <enum 'Types'>, '_label_': 'User'}

>>> Types.SERVICE_ACCOUNT.__dict__
{'_value_': 'service-account', '_name_': 'SERVICE_ACCOUNT', '__objclass__': <enum 'Types'>, '_label_': 'Service Account'}
```

astrozzc: Yea, we still need db_index to optimize the query. The type is existing data, so I am using service-account and user. We need a migration to migrate existing data before we can use `SA` though, I will add a JIRA task for this to consider more. At least add a db_index to the field to help optimize the query speed

- Being explicit can help reading the code and db records; 
- The short form can help reducing some db space;

PR #1180, opened 2024-08-30
RHCLOUD-35015: Bring mapping reads within locked transaction to prevent out of order replication

alechenninger:
**NOTE: This is not a PR to merge into main. It is a PR into the branch in #1172 (dual-writes). Merging will update that PR only.**

---

This got larger than I initially intended as I dug deeper and realized a few things:

Initially we had just wanted to fix that the mappings, access, and policy for a Role were queried prior to the lock, and the first commit should theoretically fix that under certain assumptions (namely, that the Role is always locked before any of its policy, access, or mappings are updated, which currently isn't the case but is fixable).

However, I later realized that the access and policy tables are queried off of the initial Role object. It wasn't clear that those weren't already fetched, and if they were (or the queryset was updated to eagerly fetch), it would cause those to be read outside of the lock (and lead to possible inconsistency again).

This is pretty fragile, and the whole thing is complicated, so I decided to try a little bit more refactoring to simplify things. With the latest, things work like this:

1. During update & destroy, the role and mapping are retrieved and locked by modifying get_queryset logic (and so no additional get_objects are necessary, and django rest's innate logic will do the right thing)
2. During create, update, and destroy, we use an atomic block at the view level so that the lock lasts the duration of all of our updates
3. The dual write logic is handled in `perform_` hooks provided by django rest because these have access to the serializer and/or instance and can run before and after inserts/updates/deletes happen in SQL (which is what we need). This keeps us from having to do more queries and kinda organizes all the SQL logic in (mostly) one place.
4. The same current mapping object is used inside the dual write handler throughout (since this is 1-1 with the Role)

Unfortunately we're not done – what's left now is to handle similar locking and mapping updates for add/remove role to group APIs. But I think that may warrant a follow-up PR.

### DIFF
--- a/rbac/internal/views.py
+++ b/rbac/internal/views.py
@@ -480,7 +480,7 @@ def get_param_list(request, param_name):
 def data_migration(request):
     """View method for running migrations from V1 to V2 spiceDB schema.
 
-    POST /_private/api/utils/data_migration/?exclude_apps=cost_management,rbac&orgs=id_1,id_2&write_db=True
+    POST /_private/api/utils/data_migration/?exclude_apps=cost_management,rbac&orgs=id_1,id_2&write_relationships=True
     """
     if request.method != "POST":
         return HttpResponse('Invalid method, only "POST" is allowed.', status=405)
@@ -489,7 +489,7 @@ def data_migration(request):
     args = {
         "exclude_apps": get_param_list(request, "exclude_apps"),
         "orgs": get_param_list(request, "orgs"),
-        "write_db": request.GET.get("write_db", "False") == "True",
+        "write_relationships": request.GET.get("write_relationships", "False") == "True",
     }
     migrate_data_in_worker.delay(args)
     return HttpResponse("Data migration from V1 to V2 are running in a background worker.", status=202)
--- a/rbac/management/role/relation_api_dual_write_handler.py
+++ b/rbac/management/role/relation_api_dual_write_handler.py
@@ -124,6 +124,9 @@ class RelationApiDualWriteHandler:
 
             self.role_relations = relations
 
+            if self.binding_mapping is None:
+                self.binding_mapping = BindingMapping.objects.create(role=self.role)
+
             self.binding_mapping.mappings = mappings
             self.binding_mapping.save()
 
--- a/rbac/management/role/relation_api_dual_write_handler.py
+++ b/rbac/management/role/relation_api_dual_write_handler.py
@@ -57,7 +57,7 @@ class RelationApiDualWriteHandler:
         """Check whether replication enabled."""
         return ENVIRONMENT.get_value("REPLICATION_TO_RELATION_ENABLED", default=False, cast=bool)
 
-    def generate_relations_from_current_state_of_role(self):
+    def get_relations_from_current_state_of_role(self):
         """Generate relations from current state of role and UUIDs for v2 role and role binding from database."""
         if not self.replication_enabled():
             return
--- a/rbac/management/role/relation_api_dual_write_handler.py
+++ b/rbac/management/role/relation_api_dual_write_handler.py
@@ -125,10 +125,12 @@ class RelationApiDualWriteHandler:
             self.role_relations = relations
 
             if self.binding_mapping is None:
-                self.binding_mapping = BindingMapping.objects.create(role=self.role)
-
-            self.binding_mapping.mappings = mappings
-            self.binding_mapping.save()
+                self.binding_mapping = BindingMapping.objects.create(
+                    role=self.role,
+                    mappings=mappings)
+            else:
+                self.binding_mapping.mappings = mappings
+                self.binding_mapping.save(force_update=True)
 
             return relations
         except Exception as e:
--- a/rbac/management/role/relation_api_dual_write_handler.py
+++ b/rbac/management/role/relation_api_dual_write_handler.py
@@ -23,6 +23,7 @@ from migration_tool.migrate import migrate_role
 from migration_tool.utils import relationship_to_json
 
 from rbac.env import ENVIRONMENT
+from rbac.management.role.model import BindingMapping
 
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
@@ -44,6 +45,7 @@ class RelationApiDualWriteHandler:
             self.role_relations = []
             self.current_role_relations = []
             self.role = role
+            self.binding_mapping = None
             self.tenant_id = role.tenant_id
             self.org_id = role.tenant.org_id
             self.root_workspace = Workspace.objects.get(
@@ -57,7 +59,11 @@ class RelationApiDualWriteHandler:
         """Check whether replication enabled."""
         return ENVIRONMENT.get_value("REPLICATION_TO_RELATION_ENABLED", default=False, cast=bool)
 
-    def get_relations_from_current_state_of_role(self):
+    def get_current_role_relations(self):
+        """Get current roles relations."""
+        return self.current_role_relations
+
+    def load_relations_from_current_state_of_role(self):
         """Generate relations from current state of role and UUIDs for v2 role and role binding from database."""
         if not self.replication_enabled():
             return
@@ -65,36 +71,67 @@ class RelationApiDualWriteHandler:
             logger.info(
                 "[Dual Write] Generate relations from current state of role(%s): '%s'", self.role.uuid, self.role.name
             )
-            relations = migrate_role(
-                self.role, False, str(self.root_workspace.uuid), self.org_id, True, True, True, False
+
+            self.binding_mapping = self.role.binding_mapping
+
+            if self.binding_mapping is None:
+                raise Exception(f"BindingMapping for role {self.role.uuid} not found.")
+
+            relations, _ = migrate_role(
+                self.role,
+                write_relationships=False,
+                root_workspace=str(self.root_workspace.uuid),
+                default_workspace=self.org_id,
+                current_bindings=self.binding_mapping,
             )
+
             self.current_role_relations = relations
         except Exception as e:
             raise DualWriteException(e)
 
-    def regenerate_relations_and_mappings_for_role(self):
-        """Delete and generated relations with mapping for a role."""
+    def generate_replication_event_to_outbox(self, role):
+        """Generate replication event to outbox table."""
         if not self.replication_enabled():
-            return []
-        return self.generate_relations_and_mappings_for_role()
+            return
+        self.role = role
+        self._generate_relations_and_mappings_for_role()
+        return self.save_replication_event_to_outbox()
 
-    def generate_relations_and_mappings_for_role(self):
+    def save_replication_event_to_outbox(self):
+        """Generate and store replication event to outbox table."""
+        if not self.replication_enabled():
+            return {}
+        try:
+            replication_event = self._build_replication_event()
+            self._save_replication_event(replication_event)
+        except Exception as e:
+            raise DualWriteException(e)
+        return replication_event
+
+    def _generate_relations_and_mappings_for_role(self):
         """Generate relations and mappings for a role with new UUIDs for v2 role and role bindings."""
         if not self.replication_enabled():
             return []
         try:
             logger.info("[Dual Write] Generate new relations from role(%s): '%s'", self.role.uuid, self.role.name)
-            relations = migrate_role(self.role, False, str(self.root_workspace.uuid), self.org_id, True)
+
+            relations, mappings = migrate_role(
+                self.role,
+                write_relationships=False,
+                root_workspace=str(self.root_workspace.uuid),
+                default_workspace=self.org_id,
+                current_bindings=self.binding_mapping)
+
             self.role_relations = relations
+
+            self.binding_mapping.mappings = mappings
+            self.binding_mapping.save()
+
             return relations
         except Exception as e:
             raise DualWriteException(e)
 
-    def get_current_role_relations(self):
-        """Get current roles relations."""
-        return self.current_role_relations
-
-    def build_replication_event(self):
+    def _build_replication_event(self):
         """Build replication event."""
         if not self.replication_enabled():
             return {}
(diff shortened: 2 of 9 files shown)


rbac/management/role/serializer.py
```diff
@@ -395,15 +395,22 @@ def validate_role_update(instance, validated_data):
...
+        update_fields.append("name")
+    if "updated_display_name" in update_data:
+        role.display_name = update_data["updated_display_name"]
+        update_fields.append("display_name")
+    if "updated_description" in update_data:
+        role.description = update_data["updated_description"]
+        update_fields.append("description")
+
+    role.save(update_fields=update_fields)
+
```
alechenninger, line 413: This wasn't strictly required for the change but removes extraneous locking. Can revert back to prior code if desired for any reason.

rbac/internal/views.py
```diff
@@ -480,7 +480,7 @@ def get_param_list(request, param_name):
 def data_migration(request):
     """View method for running migrations from V1 to V2 spiceDB schema.
 
-    POST /_private/api/utils/data_migration/?exclude_apps=cost_management,rbac&orgs=id_1,id_2&write_db=True
+    POST /_private/api/utils/data_migration/?exclude_apps=cost_management,rbac&orgs=id_1,id_2&write_relationships=True
```
alechenninger, line 483: I renamed this parameter throughout to make it a little clearer.

Now that migration involves writes to the local DB (BindingMapping), "write_db" was confusing to me and seemed to refer to the _local_ DB, even though what this really means is whether relation tuples will be written out directly (or just persisted in outbox).

So hopefully this name is clearer but happy to revert or try something else.

rbac/management/role/relation_api_dual_write_handler.py
```diff
@@ -57,44 +59,79 @@ def replication_enabled(self):
         """Check whether replication enabled."""
         return ENVIRONMENT.get_value("REPLICATION_TO_RELATION_ENABLED", default=False, cast=bool)
 
-    def generate_relations_from_current_state_of_role(self):
+    def get_current_role_relations(self):
```
alechenninger, line 62: I reordered / renamed some methods in this class just for clarity while trying to understand how it was meant to be used.

rbac/management/role/relation_api_dual_write_handler.py
```diff
@@ -57,44 +59,79 @@ def replication_enabled(self):
...
+
+            if self.binding_mapping is None:
+                raise Exception(f"BindingMapping for role {self.role.uuid} not found.")
+
+            relations, _ = migrate_role(
+                self.role,
+                write_relationships=False,
+                root_workspace=str(self.root_workspace.uuid),
+                default_workspace=self.org_id,
+                current_bindings=self.binding_mapping,
```
alechenninger, line 85: I changed this up a bit more than I intended for a few reasons:

- We were querying the binding mapping redundantly
- Now it's locked from the get-go, so it can be retrieved from the Role object
- Passing it around ended up cleaning up some of the parameters I found. Instead of controlling inner behavior with flags, I moved some of that behavior out here, and some of it is now controlled by passing the current bindings or not. The intent is just that it is maybe easier to use / harder to end up with inconsistent data now.

rbac/management/role/relation_api_dual_write_handler.py
```diff
@@ -57,44 +59,79 @@ def replication_enabled(self):
...
         except Exception as e:
             raise DualWriteException(e)
 
-    def regenerate_relations_and_mappings_for_role(self):
-        """Delete and generated relations with mapping for a role."""
+    def generate_replication_event_to_outbox(self, role):
+        """Generate replication event to outbox table."""
         if not self.replication_enabled():
-            return []
-        return self.generate_relations_and_mappings_for_role()
```
alechenninger, line 79: I inlined this since all it does really is call `generate_relations_and_mappings_for_role` and it already has the replication_enabled check, too.

rbac/management/role/relation_api_dual_write_handler.py
```diff
@@ -57,44 +59,79 @@ def replication_enabled(self):
...
+                self.role,
+                write_relationships=False,
+                root_workspace=str(self.root_workspace.uuid),
+                default_workspace=self.org_id,
+                current_bindings=self.binding_mapping)
+
             self.role_relations = relations
+
+            self.binding_mapping.mappings = mappings
+            self.binding_mapping.save()
```
alechenninger, line 128: I moved the mapping update and persist out of this method to (at least for me) make the logic clearer and simplify `migrate_role` arguments and implementation.

rbac/management/role/view.py
```diff
@@ -138,8 +138,39 @@ class RoleViewSet(
...
+            # Because we reuse this Role object when reading and
+            # determining current relations to remove,
+            # this lock prevents any accidental and non-obvious race conditions from occuring.
+            # (such as if this was innocently changed to select related access or policy rows)
+            # Getting the mapping here (which is necessary to lock it) also saves us
+            # redundant queries later.
+
+            return (Role.objects.filter(tenant=self.request.tenant)
+                    .select_related("mappings")
+                    .select_for_update())
```
alechenninger, line 173: This is a key bit

rbac/management/role/view.py
```diff
@@ -336,27 +362,13 @@ def destroy(self, request, *args, **kwargs):
...
                 response = super().destroy(request=request, args=args, kwargs=kwargs)
-                dual_write_handler.save_replication_event_to_outbox()
         except DualWriteException as e:
             return self.dual_write_exception_response(e)
 
-        if response.status_code == status.HTTP_204_NO_CONTENT:
-            role_obj_change_notification_handler(role, "deleted", request.user)
-
-            auditlog = AuditLog()
-            auditlog.log_delete(request, AuditLog.ROLE, role)
```
alechenninger, line 359: Audit log stuff seemed to make more sense in the `perform_` hooks now that we added those, since they already have access to the latest instance. But happy to revert back if desired.

rbac/management/role/view.py
```diff
@@ -336,27 +362,13 @@ def destroy(self, request, *args, **kwargs):
...
-        role = self.get_object()
-        if role.system or role.platform_default:
-            key = "role"
-            message = "System roles cannot be deleted."
-            error = {key: [_(message)]}
-            raise serializers.ValidationError(error)
+
         try:
             with transaction.atomic():
-                self.delete_policies_if_no_role_attached(role)
```
alechenninger, line 347: This needed to come after dual write handler reads current state. See code comments below.

rbac/management/role/serializer.py
```diff
@@ -139,7 +139,6 @@ def create(self, validated_data):
         role = Role.objects.create(name=name, description=description, display_name=display_name, tenant=tenant)
         create_access_for_role(role, access_list, tenant)
 
-        role_obj_change_notification_handler(role, "created", self.context["request"].user)
```
alechenninger, line 142: This was moved to `perform_` hooks just for consistency. It has to be there for delete (because delete doesn't use the serializer), and we already have "outbox-y" style things there now (so auditing, outbox, and notifications are all together).

rbac/management/role/view.py
```diff
@@ -138,8 +138,45 @@ class RoleViewSet(
...
+
+            # NOTE: If we want to try REPEATABLE READ isolation instead of READ COMMITTED,
+            # this should work with that as well.
+            # You would be able to remove `select_for_update` here,
+            # and instead rely on REPEATABLE READ's lost update detection to abort the tx.
+            # Nothing else should need to change.
+
+            return (Role.objects.filter(tenant=self.request.tenant)
+                    .select_related("mappings")
+                    .select_for_update())
```
alechenninger, line 179: This is the key bit.

rbac/management/role/relation_api_dual_write_handler.py
```diff
@@ -57,44 +59,82 @@ def replication_enabled(self):
...
+                current_bindings=self.binding_mapping)
+
             self.role_relations = relations
+
+            if self.binding_mapping is None:
+                self.binding_mapping = BindingMapping.objects.create(role=self.role)
+
+            self.binding_mapping.mappings = mappings
+            self.binding_mapping.save()
+
```
alechenninger, line 132: I moved the mapping update and persist out of `migrate_role` to (at least for me) make the logic clearer and simplify `migrate_role` arguments and implementation. All mapping SQL is now performed inside the dual write handler.

rbac/management/role/relation_api_dual_write_handler.py
```diff
@@ -57,44 +59,84 @@ def replication_enabled(self):
...
             self.role_relations = relations
+
+            if self.binding_mapping is None:
+                self.binding_mapping = BindingMapping.objects.create(
+                    role=self.role,
+                    mappings=mappings)
+            else:
+                self.binding_mapping.mappings = mappings
+                self.binding_mapping.save(force_update=True)
```
alechenninger, line 133: I moved the mapping update and persist out of migrate_role to (at least for me) make the logic clearer and simplify migrate_role arguments and implementation. All mapping SQL is now performed inside the dual write handler.

rbac/management/role/view.py
alechenninger: From review: add back user access check from original role queryset

rbac/migration_tool/sharedSystemRolesReplicatedRoleBindings.py
```diff
@@ -97,13 +97,12 @@ def set_system_roles(cls):
 
 def v1_role_to_v2_mapping(
```
coderbydesign, line 98: It looks like there are some other callers still passing in more arguments we'd need to update the signature for: https://github.com/RedHatInsights/insights-rbac/pull/1180/files#diff-1dbf098bdd583306097f07e14314357a5949c2c93a1cfd1d562cbc426f6099d9R260

rbac/migration_tool/sharedSystemRolesReplicatedRoleBindings.py
```diff
@@ -165,14 +158,16 @@ def v1_role_to_v2_mapping(
     return frozenset(v2_role_bindings)
 
 
-candidate_system_roles = {}
 custom_roles_created = 0
 
 
-def extract_system_roles(perm_groupings, v1_role, db_role_id, use_mapping_from_db=False):
+def extract_system_roles(perm_groupings: dict[V1resourcedef, list[str]], v1_role: V1role, binding_mapping: BindingMapping | None):
```
coderbydesign, line 164: It looks like this syntax for union types is supported in Python 3.10 + https://peps.python.org/pep-0604/ and [we're currently requiring 3.9](https://github.com/RedHatInsights/insights-rbac/blob/master/Pipfile#L62) 

Maybe this could work prior to an upgrade? https://docs.python.org/3/library/typing.html#typing.Optional

rbac/migration_tool/sharedSystemRolesReplicatedRoleBindings.py
alechenninger: Ah, nice catch. Will fix.

rbac/migration_tool/sharedSystemRolesReplicatedRoleBindings.py
alechenninger: Looks like that did the trick ty

rbac/management/role/relation_api_dual_write_handler.py
```diff
@@ -57,44 +59,87 @@ def replication_enabled(self):
...
-                self.role, False, str(self.root_workspace.uuid), self.org_id, True, True, True, False
+
+            try:
+                self.binding_mapping = self.role.binding_mapping
+            except BindingMapping.DoesNotExist:
+                logger.warning(
+                    "[Dual Write] Binding mapping not found for role(%s): '%s'. Creating new binding mapping.",
+                    self.role.uuid,
+                    self.role.name,
+                )
```
alechenninger, line 82: This was not getting hit before for how the test cases were set up, but now this case is exposed so I added this warning instead of failing. Whether or not we fail this depends how we want to handle migration. I don't actually think allowing this will cause any inconsistency to happen as long as if there are no mappings stored in rbac, it also means there are no corresponding relations in the Relations API. So I think a warning is appropriate.

rbac/management/role/view.py
```diff
@@ -440,28 +453,71 @@ def update(self, request, *args, **kwargs):
...
+        """
+        Update the role and publish outbox, notification, and audit events.
+
+        Assumes concurrent updates are prevented (e.g. with atomic block and locks).
+        """
+        if self.action != "partial_update":
+            dual_write_handler = RelationApiDualWriteHandler(serializer.instance, "UPDATE")
+            dual_write_handler.load_relations_from_current_state_of_role()
+
+        role = serializer.save()
```
alechenninger, line 488: One thing to note about this happening before serializer.save() here is that this load will happen before some validation in the serializer happens. The serializer has a `validate_role_update` method which checks if the role is a system role. This could fail and we could have saved ourselves from building up current mappings.

It's tempting to move that validation here, which would make it more cohesive with the same validation that happens in `perform_destroy` but I'll leave that up to others reviewing.

rbac/migration_tool/sharedSystemRolesReplicatedRoleBindings.py
alechenninger: Looks like that other callsite actually wasn't used (it just forwarded to another method). So I was able to just remove it.

rbac/management/role/relation_api_dual_write_handler.py
coderbydesign: I think we could get rid of the nested `try/except` here and just catch the error to log the warning below, which might make it easier to read?

```python
try:
    logger.info(
        "[Dual Write] Generate relations from current state of role(%s): '%s'", self.role.uuid, self.role.name
    )

    self.binding_mapping = self.role.binding_mapping

    relations, _ = migrate_role(
        self.role,
        write_relationships=False,
        root_workspace=str(self.root_workspace.uuid),
        default_workspace=self.org_id,
        current_bindings=self.binding_mapping,
    )

    self.current_role_relations = relations
except Exception as e:
    raise DualWriteException(e)
except BindingMapping.DoesNotExist:
    logger.warning(
        "[Dual Write] Binding mapping not found for role(%s): '%s'. Creating new binding mapping.",
        self.role.uuid,
        self.role.name,
    )
```

alechenninger: I do like how that reads better. One problem is that we still need to run `migrate_role` in that case though. Wondering if there is another way 🤔

alechenninger: Maybe we don't actually? 🤔

alechenninger: Yeah I think this should work since in this case there _shouldn't_ be any relations that we have to remove. I just had to put the generic `Exception` catch block last. Good one!

astrozzc (approved):
LGTM!

PR #1184, opened 2024-09-05
RHCLOUD-35016: Resource to workspace relations from RBAC may be out of date; do not migrate

alechenninger:
## Link(s) to Jira
- RHCLOUD-35016: https://issues.redhat.com/browse/RHCLOUD-35016

Avoid migrating relations that we should not migrate due to race conditions with services that own the resources themselves and Inventory API. See Jira.

## Local Testing

Unit tests

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


rbac/rbac/settings.py
```diff
@@ -346,6 +346,15 @@
 
 ROLE_CREATE_ALLOW_LIST = ENVIRONMENT.get_value("ROLE_CREATE_ALLOW_LIST", default="").split(",")
 
+DEFAULT_DUAL_WRITE_APP_DENY_LIST = [
```
coderbydesign, line 349: A naming thing (which I'm fine either way with), but we do have an `exclude_apps` param in the migrator: https://github.com/RedHatInsights/insights-rbac/blob/d8051fd8ff1041f054bacdb657ce135e6b8977cc/rbac/migration_tool/migrate.py#L157C18-L157C30 I wonder does it make sense to name these `DUAL_WRITE_APP_EXCLUDE_LIST` or make them consistent?

rbac/rbac/settings.py
```diff
@@ -346,6 +346,15 @@
...
+    "approval",
+    "cost-management",
+    "playbook-dispatcher",
+]
+DUAL_WRITE_APP_DENY_LIST = ENVIRONMENT.get_value(
```
coderbydesign, line 354: So this will utilize the default list _or_ a list that you define, correct? We do have a pattern already for some other app lists: https://github.com/RedHatInsights/insights-rbac/blob/d8051fd8ff1041f054bacdb657ce135e6b8977cc/deploy/rbac-clowdapp.yml#L733-L734 where instead of hard-coding the default list, we can have a default in the template, parameterized, and then we can set it per environment as well. Thoughts?

Also, is this not being consumed (yet)?

rbac/migration_tool/ingest.py
```diff
@@ -47,7 +53,16 @@ def extract_info_into_v1_role(role: Role):
...
+
+    # With the replicated role bindings algorithm, role bindings are scoped by group, so we need to add groups
+    # TODO: We don't need to care about principals here – see RHCLOUD-35039
+    # Maybe not even groups? See RHCLOUD-34786
+    groups = set()
+    for policy in role.policies.all():
+        principals = [str(principal) for principal in policy.group.principals.values_list("uuid", flat=True)]
+        groups.add(V1group(str(policy.group.uuid), frozenset(principals)))
+
+    return V1role(role_id, frozenset(v1_perms), frozenset(groups))  # we don't get groups from the sheet
```
coderbydesign, line 65: Can we drop this comment now that we've got the groups?

alechenninger: Nice catch

rbac/rbac/settings.py
alechenninger: Love it. Thank you.

rbac/rbac/settings.py
alechenninger: Yeah I appreciate that consistency. It was hard coded before so I sort of ported it "as is" but that's not really an excuse!

It's a bit burred in the diff but it's used in sharedSystemRolesReplicatedRoleBindings.py:

```python
def is_for_enabled_app(perm: V1permission):
    """Return true if the permission is for an app that should migrate."""
    return perm.app not in settings.DUAL_WRITE_APP_DENY_LIST
```

I'll make the update!

astrozzc (approved):

rbac/rbac/settings.py
coderbydesign: Ahh!! Missed that, thanks! 😄

rbac/rbac/settings.py
alechenninger: Ended up doing a few things (will push momentarily):

- Refactored to put defaults in clowdapp as suggested
- Introduced a new environment variable to separate out app exclusion vs just resource definition exclusion (as discussed this morning)
- Renamed variables as I realized they are not exclusive to "dual write" but to v2 migration in general

Question: For defaults in the clowdapp, does this make its way into app-interface or do we need to also update configs there?

alechenninger:
@coderbydesign @astrozzc @lpichler  Alright take another look and let me know what you think.

Couple of notes::

- I expanded the exclusion feature slightly to allow excluding an app entirely (e.g. approval) vs excluding only the resource definitions (e.g. cost, playbook dispatcher)
- A test is failing right now due to https://issues.redhat.com/browse/RHCLOUD-35040 – I would like to fix that separately since it is a more involved fix (and this is still a PR into a feature branch)

rbac/rbac/settings.py
lpichler: >Question: For defaults in the clowdapp, does this make its way into app-interface or do we need to also update configs there?

We don't need to update app-interface configs. Presence in app-interface configs allows us to use same image - it overrides defaults in clowdapp yaml.

rbac/management/management/commands/migrate_relations.py
```diff
@@ -15,8 +17,13 @@ class Command(BaseCommand):
     def add_arguments(self, parser):
         """Add arguments to command."""
         parser.add_argument("--org-list", nargs="+", default=[])
-        parser.add_argument("--exclude-apps", nargs="+", default=[])
-        parser.add_argument("--write-to-db", default=False, action="store_true")
+        parser.add_argument(
+            "--exclude-apps",
+            nargs="+",
+            default=settings.V2_MIGRATION_APP_EXCLUDE_LIST,
```
coderbydesign, line 23: Should we add this in the internal/Turnpike endpoint defaults as well? https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/internal/views.py#L490

coderbydesign (approved):

PR #1189, opened 2024-09-11
Add GET /workspace/<uuid>/ detail ancestry in payload

coderbydesign:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-35031

To support the workspace UI, we'll need to supply ancestry for workspaces so that on the "detail" page for workspaces, breadcrumbs and links can be constructed in order to help with the UX for navigation and awareness.

**Questions:**

- Currently this adds ~all fields on the workspace (`description` may not be needed) to keep reuse and reduce complexity, do we feel like we should strip this down now?~ a custom serializer to just return `name`, `parent_id` and `uuid` in the `ancestry` list. 

- ~Do we think it's ok to always include `ancestry`, or would we want to optionally include it with a param/new endpoint like `/workspaces/<uuid>/ancestry` which in theory would require an additional call from the UI?~ We'll support an optional `?include_ancestry=true|false` query param which defaults to `false` to include the `ancestry: []` list.

## Local Testing
Query a workspace with ancestors (and children) `/workspaces/<uuid>/` and note that you should not see `ancestry` by default, but after querying `/workspaces/<uuid>/?include_ancestry=true` you'll see a list of workspace objects which can be used to build the tree back to the root parent, but no children.

```json
{
  "uuid": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
  "parent_id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
  "name": "Workspace A",
  "description": "Description of Workspace A",
  "created": "2024-09-11T18:58:54.083Z",
  "modified": "2024-09-11T18:58:54.083Z",
  "ancestry": [
    {
      "uuid": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
      "parent_id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
      "name": "Workspace A"
    }
  ]
}
```

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/docs/source/specs/typespec/main.tsp
+++ b/docs/source/specs/typespec/main.tsp
@@ -103,13 +103,17 @@ namespace Workspaces {
         ...Timestamps;
     }
 
+    model WorkspaceWithAncestry extends Workspace {
+        ancestry: Workspace[];
+    }
+
     alias CreateWorkspaceRequest = BasicWorkspace;
 
     model CreateWorkspaceResponse extends BasicWorkspace{
         @statusCode _: 201;
     }
 
-    model ReadWorkspaceResponse extends BasicWorkspace{
+    model ReadWorkspaceResponse extends WorkspaceWithAncestry{
         @statusCode _: 200;
     }
 
--- a/docs/source/specs/typespec/main.tsp
+++ b/docs/source/specs/typespec/main.tsp
@@ -205,7 +205,10 @@ namespace Workspaces {
 
     @doc("List workspaces in a tenant")
     @summary("List workspaces in a tenant")
-    @get op list(): WorkspaceListResponse | Problems.CommonProblems;
+    @get op list(
+        @query limit?: int64 = 10;
+        @query offset?: int64 = 0;
+    ): WorkspaceListResponse | Problems.CommonProblems;
 
     @doc("Create workspace in tenant")
     @summary("Create workspace in tenant")
--- a/docs/source/specs/typespec/main.tsp
+++ b/docs/source/specs/typespec/main.tsp
@@ -111,7 +111,7 @@ namespace Workspaces {
 
     alias CreateWorkspaceRequest = BasicWorkspace;
 
-    model CreateWorkspaceResponse extends BasicWorkspace{
+    model CreateWorkspaceResponse extends Workspace{
         @statusCode _: 201;
     }
 
@@ -127,7 +127,9 @@ namespace Workspaces {
         @doc("Workspace A")
         name: string = "Workspace A";
         @doc("Description of Workspace A")
-        description?: string = "Description of Workspace A";
+        description: string = "Description of Workspace A";
+        @doc("Description of Workspace A")
+        parent_id: UUID = "Parent UUID of Workspace A";
     }
 
     model PatchWorkspaceRequest {
@@ -137,11 +139,11 @@ namespace Workspaces {
         description?: string = "Description of Workspace A";
     }
 
-    model UpdateWorkspaceResponse extends BasicWorkspace{
+    model UpdateWorkspaceResponse extends Workspace{
         @statusCode _: 200;
     }
 
-    model PatchWorkspaceResponse extends BasicWorkspace{
+    model PatchWorkspaceResponse extends Workspace{
         @statusCode _: 200;
     }
 
--- a/docs/source/specs/typespec/main.tsp
+++ b/docs/source/specs/typespec/main.tsp
@@ -103,7 +103,7 @@ namespace Workspaces {
         ...Timestamps;
     }
 
-    model WorkspaceAncestry {
+    model WorkspaceAncestor {
         @key uuid: UUID;
         parent_id?: UUID;
         name: string;
@@ -120,7 +120,7 @@ namespace Workspaces {
     }
 
     model ReadWorkspaceWithAncestryResponse extends ReadWorkspaceResponse {
-        ancestry: WorkspaceAncestry[];
+        ancestry: WorkspaceAncestor[];
     }
 
     model UpdateWorkspaceRequest {
--- a/docs/source/specs/typespec/main.tsp
+++ b/docs/source/specs/typespec/main.tsp
@@ -103,8 +103,10 @@ namespace Workspaces {
         ...Timestamps;
     }
 
-    model WorkspaceWithAncestry extends Workspace {
-        ancestry: Workspace[];
+    model WorkspaceAncestry {
+        @key uuid: UUID;
+        parent_id?: UUID;
+        name: string;
     }
 
     alias CreateWorkspaceRequest = BasicWorkspace;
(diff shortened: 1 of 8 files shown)


rbac/management/workspace/model.py
```diff
@@ -38,3 +38,14 @@ class Workspace(TenantAwareModel):
 
     class Meta:
         ordering = ["name", "modified"]
+
+    def ancestors(self):
+        """Return a list of ancestors for a Workspace instance."""
+        ancestors = []
+        parent = self.parent
```
lpichler, line 45: if there will performance implications with this it looks we can also leverage  recursive queries which Alec  mentioned recently, something like: (and then query workspaces objects)

```
WITH RECURSIVE ancestors AS (
    SELECT uuid, parent_id
    FROM management_workspace
    WHERE uuid = 'f503650a-c72e-446d-825d-39beb9e75eff' <- get ancestors for this workspaces 
    UNION ALL
    SELECT w.uuid, w.parent_id
    FROM management_workspace w
             INNER JOIN ancestors a ON a.uuid = w.parent_id
)
SELECT
    parent_id
FROM
    ancestors WHERE parent_id IS NOT NULL;
```

coderbydesign: @lpichler I ended up implementing similar to what you suggested here in [9061c6c](https://github.com/RedHatInsights/insights-rbac/pull/1189/commits/9061c6c870d0342e77379f09b70aaed9ebe34e21) with a few changes outlined in the commit/query you should be able to see.

rbac/management/workspace/model.py
```diff
@@ -38,3 +38,14 @@ class Workspace(TenantAwareModel):
 
     class Meta:
         ordering = ["name", "modified"]
+
+    def ancestors(self):
```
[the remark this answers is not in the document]
coderbydesign: The way this logic _was_, yes, otherwise we'd potentially have the ability to create an infinite loop. With [9061c6c](https://github.com/RedHatInsights/insights-rbac/pull/1189/commits/9061c6c870d0342e77379f09b70aaed9ebe34e21) we should be ok without them. In testing we'd just simulate a root without a parent.

rbac/management/workspace/model.py
```diff
@@ -38,3 +38,29 @@ class Workspace(TenantAwareModel):
...
+            WITH RECURSIVE ancestors AS
+              (SELECT uuid,
+                      parent_id
+               FROM management_workspace
+               WHERE uuid = %s
+               UNION SELECT w.uuid,
+                                w.parent_id
+               FROM management_workspace w
+               JOIN ancestors a ON w.uuid = a.parent_id)
+            SELECT uuid AS uuid,
```
coderbydesign, line 61: We need the `uuid` value for the next query, but Django requires the PK `id`, so I'm aliasing it here. This will still only return `uuid` but allows for the query to succeed.

rbac/management/workspace/model.py
```diff
@@ -38,3 +38,29 @@ class Workspace(TenantAwareModel):
 
     class Meta:
         ordering = ["name", "modified"]
+
+    def ancestors(self):
+        """Return a list of ancestors for a Workspace instance."""
+        ancestor_ids = [a.uuid for a in self._ancestry_queryset() if a.uuid != self.uuid]
+        ancestors = Workspace.objects.filter(uuid__in=ancestor_ids)
```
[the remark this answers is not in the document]
coderbydesign: Pushed this change into the serializer to chain `.only()` to the queryset, in order to leave this untouched in the event full ancestor objects are needed while still optimizing the serialized object query.

docs/source/specs/typespec/main.tsp
```diff
@@ -103,16 +103,26 @@ namespace Workspaces {
...
 
+    model WorkspaceAncestry {
+        @key uuid: UUID;
+        parent_id?: UUID;
+        name: string;
+    }
+
     alias CreateWorkspaceRequest = BasicWorkspace;
 
     model CreateWorkspaceResponse extends BasicWorkspace{
```
vbelchio, line 114: @coderbydesign Maybe we need to change here to `Workspace` instead of `BasicWorkspace` once the JSON response has all the fields.

coderbydesign: Updated!

rbac/management/workspace/model.py
astrozzc, line 45: This query would help so we don't have to do extra query, and get rid off that only() from the serializer:
`        return Workspace.objects.raw(
            """
            WITH RECURSIVE ancestors AS
              (SELECT name, uuid, parent_id
               FROM management_workspace
               WHERE uuid = %s
               UNION SELECT w.name, w.uuid, w.parent_id
               FROM management_workspace w
               JOIN ancestors a ON w.uuid = a.parent_id)
            SELECT uuid AS uuid,
                   uuid AS id,
                   name,
                   parent_id
            FROM ancestors
            WHERE uuid != %s
        """,
            [self.uuid, self.uuid],
        )
`

astrozzc (approved):
LGTM

rbac/management/workspace/model.py
coderbydesign: Hey @astrozzc! I'd considered similar as well, but wanted to leave the base method for returning `ancestors` actually returning the full object/queryset in the event that we need those objects' values. The `only()` call should modify the queryset before the query is executed since what comes out of here is the queryset. Thoughts?

astrozzc: Got it. Just worried a bit about the extra query, seems redundant. But sounds good to me

vbelchio (approved):
QE: Passed

docs/source/specs/typespec/main.tsp
```diff
@@ -103,21 +103,33 @@ namespace Workspaces {
...
+    }
+
     model UpdateWorkspaceRequest {
         @doc("Workspace A")
         name: string = "Workspace A";
         @doc("Description of Workspace A")
-        description?: string = "Description of Workspace A";
+        description: string = "Description of Workspace A";
+        @doc("Description of Workspace A")
+        parent_id: UUID = "Parent UUID of Workspace A";
```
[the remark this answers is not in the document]
coderbydesign: @alechenninger that all makes total sense. This specifically was to fix discrepancies between the current API behavior and the API spec so that clients can actually be built (currently `parent_id` is required in the update). I totally agree with what you're saying, and if it's good with you I'd prefer to log a ticket to _prevent_ `parent_id` from being changed, and not require it in PUT/PATCH.

docs/source/specs/typespec/main.tsp
```diff
@@ -103,21 +103,33 @@ namespace Workspaces {
         ...Timestamps;
     }
 
+    model WorkspaceAncestry {
```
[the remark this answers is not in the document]
coderbydesign: Updated in definition and reference.

docs/source/specs/typespec/main.tsp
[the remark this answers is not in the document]
coderbydesign: Logged a shell ticket here: https://issues.redhat.com/browse/RHCLOUD-35415

PR #1190, opened 2024-09-11
Removing some fields as required to match the real responses

vbelchio:
## Link(s) to Jira
-

The fields `next` and `previous` from the `PaginationLinks` model could be `null` some cases.

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


docs/source/specs/typespec/main.tsp
```diff
@@ -163,12 +163,12 @@ namespace Workspaces {
         @doc("The next link in pagination")
         @example("/api/rbac/v2/(resource)/?limit=10&offset=10")
         @format("uri")
-        next: string;
+        next?: string;
```
coderbydesign, line 166: I _think_ what we'd want for these is to make them nullable, since the keys should always exists, where this would imply the key may be optional? Something like this would ensure it's always there, but make the spec update to `nullable: true`:

```suggestion
        next: string | null;
```

coderbydesign (approved):
🌴

PR #1198, opened 2024-09-16
[RHCLOUD-35039] Generate replication event for v1 Group add / remove principal endpoints

lpichler:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-35039

This PR generated replication event for adding and removing principals.  Replication event is written in outbox table.

- reorganize code to pull out API call from transaction
- lock group and principals records
- generate replication event for added principals and service accounts

TODOs:
- [x] tests
- [x] removal principals from groups

## Local Testing
```
POST http://localhost:8000/api/rbac/v1/groups/5f067122-0e23-49ad-a604-4c91c5fe071e/principals/
x-rh-identity: 

{
  "principals": [
    {
      "username": "user_name"
    },
    {
      "type": "service-account",
      "clientId": "<uuid>"
    }
  ]
}
```

Example of replication event:
```
{
  "relations_to_add": [
    {
      "resource": {
        "type": "group",
        "id": "5f067122-0e23-49ad-a604-4c91c5fe071e"
      },
      "relation": "member",
      "subject": {
        "type": "user",
        "id": "98c39729-1920-4fcf-ad6d-fce4a23432eb"
      }
    },
    {
      "resource": {
        "type": "group",
        "id": "5f067122-0e23-49ad-a604-4c91c5fe071e"
      },
      "relation": "member",
      "subject": {
        "type": "user",
        "id": "cbf1e515-80ed-4683-a6c1-ac10710473d2"
      }
    }
  ],
  "relations_to_remove": []
}
```

SQL log  add principals and services accounts

https://privatebin.corp.redhat.com/?f8551a6f459c36d9#3iXYJHnq264rpeYjAJLV4QW1GCGReoQ2bkZ5FHkhKM7Q

SQL logs remove principals and service accounts 

https://privatebin.corp.redhat.com/?13873ff898734d29#128ERA1UVHUNtV4q2gtSqCdKSnmyLPXBtjZQChy4SSgY

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/management/group/relation_api_dual_write_group_handler.py
+++ b/rbac/management/group/relation_api_dual_write_group_handler.py
@@ -1,0 +1,114 @@
+#
+# Copyright 2024 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+"""Class to handle Dual Write API related operations."""
+import logging
+from typing import Optional
+
+from django.conf import settings
+from management.role.relation_api_dual_write_handler import (
+    DualWriteException,
+    OutboxReplicator,
+    RelationReplicator,
+    ReplicationEvent,
+    ReplicationEventType
+)
+from migration_tool.utils import create_relationship
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+
+class RelationApiDualWriteGroupHandler:
+    """Class to handle Dual Write API related operations."""
+
+    def __init__(
+        self,
+        group,
+        event_type: ReplicationEventType,
+        principal_usernames,
+        service_accounts,
+        replicator: Optional[RelationReplicator] = None,
+    ):
+        """Initialize RelationApiDualWriteGroupHandler."""
+        if not self.replication_enabled():
+            return
+        try:
+            self.group_relations_to_add = []
+            self.group_relations_to_remove = []
+            self.principal_usernames = principal_usernames
+            self.service_accounts = service_accounts
+            self.group = group
+            self.event_type = event_type
+            self._replicator = replicator if replicator else OutboxReplicator(group)
+        except Exception as e:
+            raise DualWriteException(e)
+
+    def replication_enabled(self):
+        """Check whether replication enabled."""
+        return settings.REPLICATION_TO_RELATION_ENABLED is True
+
+    def _generate_relations(self):
+        """Generate user-groups relations."""
+        relations = []
+        for principal in self.principal_usernames + self.service_accounts:
+            relations.append(
+                create_relationship(
+                    ("rbac", "group"),
+                    str(self.group.uuid),
+                    ("rbac", "user"),
+                    str(principal.uuid),
+                    "member"
+                )
+            )
+
+        return relations
+
+    def replicate_new_principals(self):
+        """Replicate new principals into group."""
+        if not self.replication_enabled():
+            return
+        logger.info("[Dual Write] Generate new relations from Group(%s): '%s'", self.group.uuid, self.group.name)
+
+        self.group_relations_to_add = self._generate_relations()
+        self._replicate()
+
+    def replicate_removed_principals(self):
+        """Replicate removed principals from group."""
+        if not self.replication_enabled():
+            return
+        logger.info("[Dual Write] Generate new relations from Group(%s): '%s'", self.group.uuid, self.group.name)
+
+        self.group_relations_to_remove = self._generate_relations()
+
+        self._replicate()
+
+    def _replicate(self):
+        if not self.replication_enabled():
+            return
+        try:
+            self._replicator.replicate(
+                ReplicationEvent(
+                    type=self.event_type,
+                    # TODO: need to think about partitioning
+                    # Maybe resource id
+                    partition_key="rbactodo",
+                    remove=self.group_relations_to_remove,
+                    add=self.group_relations_to_add,
+                ),
+            )
+        except Exception as e:
+            raise DualWriteException(e)
--- a/rbac/management/group/relation_api_dual_write_group_handler.py
+++ b/rbac/management/group/relation_api_dual_write_group_handler.py
@@ -25,7 +25,7 @@ from management.role.relation_api_dual_write_handler import (
     OutboxReplicator,
     RelationReplicator,
     ReplicationEvent,
-    ReplicationEventType
+    ReplicationEventType,
 )
 from migration_tool.utils import create_relationship
 
@@ -67,11 +67,7 @@ class RelationApiDualWriteGroupHandler:
         for principal in self.principal_usernames + self.service_accounts:
             relations.append(
                 create_relationship(
-                    ("rbac", "group"),
-                    str(self.group.uuid),
-                    ("rbac", "user"),
-                    str(principal.uuid),
-                    "member"
+                    ("rbac", "group"), str(self.group.uuid), ("rbac", "user"), str(principal.uuid), "member"
                 )
             )
 
--- a/rbac/management/group/relation_api_dual_write_group_handler.py
+++ b/rbac/management/group/relation_api_dual_write_group_handler.py
@@ -20,6 +20,7 @@ import logging
 from typing import Optional
 
 from django.conf import settings
+from management.principal.model import Principal
 from management.role.relation_api_dual_write_handler import (
     DualWriteException,
     OutboxReplicator,
@@ -39,8 +40,7 @@ class RelationApiDualWriteGroupHandler:
         self,
         group,
         event_type: ReplicationEventType,
-        principal_usernames,
-        service_accounts,
+        principals: list[Principal],
         replicator: Optional[RelationReplicator] = None,
     ):
         """Initialize RelationApiDualWriteGroupHandler."""
@@ -49,8 +49,7 @@ class RelationApiDualWriteGroupHandler:
         try:
             self.group_relations_to_add = []
             self.group_relations_to_remove = []
-            self.principal_usernames = principal_usernames
-            self.service_accounts = service_accounts
+            self.principals = principals
             self.group = group
             self.event_type = event_type
             self._replicator = replicator if replicator else OutboxReplicator(group)
@@ -64,7 +63,7 @@ class RelationApiDualWriteGroupHandler:
     def _generate_relations(self):
         """Generate user-groups relations."""
         relations = []
-        for principal in self.principal_usernames + self.service_accounts:
+        for principal in self.principals:
             relations.append(
                 create_relationship(
                     ("rbac", "group"), str(self.group.uuid), ("rbac", "user"), str(principal.uuid), "member"
--- a/rbac/management/group/view.py
+++ b/rbac/management/group/view.py
@@ -702,7 +702,7 @@ class GroupViewSet(
                     group, new_principals = self.add_principals(group, principals_from_response, org_id=org_id)
 
                 dual_write_handler = RelationApiDualWriteGroupHandler(
-                    group, ReplicationEventType.ADD_PRINCIPALS_TO_GROUP, new_principals, new_service_accounts
+                    group, ReplicationEventType.ADD_PRINCIPALS_TO_GROUP, new_principals + new_service_accounts
                 )
                 dual_write_handler.replicate_new_principals()
 
@@ -918,8 +918,7 @@ class GroupViewSet(
                 dual_write_handler = RelationApiDualWriteGroupHandler(
                     group,
                     ReplicationEventType.REMOVE_PRINCIPALS_FROM_GROUP,
-                    principals_to_remove,
-                    service_accounts_to_remove,
+                    principals_to_remove + service_accounts_to_remove,
                 )
                 dual_write_handler.replicate_removed_principals()
 
--- a/rbac/management/group/view.py
+++ b/rbac/management/group/view.py
@@ -17,7 +17,7 @@
 
 """View for group management."""
 import logging
-from typing import Iterable, Optional
+from typing import Iterable, List, Optional, Tuple
 from uuid import UUID
 
 import requests
@@ -476,7 +476,7 @@ class GroupViewSet(
         group: Group,
         service_accounts: Iterable[dict],
         org_id: str = "",
-    ) -> Group:
+    ) -> Tuple[Group, List[Principal]]:
         """Add service accounts to the group."""
         # Get the tenant in order to fetch or store the service account in the database.
         tenant: Tenant = self.request.tenant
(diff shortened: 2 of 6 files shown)


rbac/management/group/view.py
```diff
@@ -669,19 +677,41 @@ def principals(self, request: Request, uuid: Optional[UUID] = None):
...
             if len(principals) > 0:
-                resp = self.add_principals(group, principals, org_id=org_id)
+                proxy_response = self.validate_principals_in_proxy_request(principals, org_id=org_id)
 
-            # Storing user principals might return an error structure instead of a group,
-            # so we need to check that before returning a response.
-            if isinstance(resp, dict) and "errors" in resp:
-                return Response(status=resp["status_code"], data=resp["errors"])
+            with transaction.atomic():
+                group = self._get_locked_object()
```
alechenninger, line 683: I'm actually not sure we need to lock the group?

The main thing we're worried about is the outbox being out of order.

In the role case, the outbox event was _dependent_ upon all events before it, because it referenced existing IDs that could have been changed by the very previous / concurrent tx.

In the group principal case, the IDs are always the same – group, user & service account ID – regardless of previous state. So I think we should be able to get away without locking the group too in this case.

For example, there can definitely be race conditions but I don't think they "matter" in that, though they may appear odd in some cases, they still result in consistent relations. For example:

1. tx1 add user1 start
2. tx2 add user1 start
3. tx1 add user commit – outbox add user1
4. tx3 remove user start
5. tx3 remove user commit – outbox remove user1
6. tx2 add user commit – outbox add user1 

This would result in a user added even though we just removed them, but the rbac db & relations would still be consistent about this. It's the same outcome as if tx2 just started a "little" bit later and was after tx3 instead of interleaving.

Locking would change the order (tx2 would block until tx1 committed, tx3 would block until tx2 committed, and therefore we'd end up with the user removed instead of added), but again this is only the difference in slight timings. It doesn't change the consistency of the data.

The main thing I think is just that any group changes MUST be committed with the associated outbox event wholly representing those changes in the same transaction.

Can you think of a case where we need to lock the group?

rbac/management/group/relation_api_dual_write_group_handler.py
```diff
@@ -0,0 +1,142 @@
...
+            service_account_client_ids = [item["clientId"] for item in self.service_account_client_ids]
+
+            filtered_principal_and_service_accounts_uuids = self.group.principals.filter(
+                Q(username__in=principal_usernames) | Q(service_account_id__in=service_account_client_ids)
+            ).values_list("uuid", flat=True)
+
+            for principal_uuid in filtered_principal_and_service_accounts_uuids:
+                relations.append(
+                    create_relationship("group", str(self.group.uuid), "user", str(principal_uuid), "member")
+                )
```
alechenninger, line 99: May I ask why the additional query here?

I'm suspicious it would cause race conditions (and require us to lock the group) although I haven't proven that yet. But regardless of that, it maybe seems unnecessary?

If another tx removed the users from the group, that tx should commit and write its own replication event.

I'm hoping (perhaps naively) that group principal add/remove can be very simple, essentially 3 sql statements:

For a req to add principals

1. insert principals & insert group fks from principals in request
2. insert outbox event w/ principals in request
3. commit

And likewise for remove.

lpichler: Group is locked but query may be seems unnecessary. It was about what id of principal and service account we will have in relation. 

Request payload contains username of principal and clientID service account. 

so I will keep UUID as id in relation - eventually we can change it later.

Anyway any IDs could be passed as they are available from additions - it just need to be extracted and passed here.

I will remove query and pass UUID it here.

rbac/management/group/view.py
```diff
@@ -423,25 +424,29 @@ def add_principals(self, group, principals, org_id=None):
...
+
+    def add_principals(self, group, principals_from_response, org_id=None):
+        """Add principals to the group."""
+        tenant = self.request.tenant
+
+        for item in principals_from_response:
             username = item["username"]
             try:
-                principal = Principal.objects.get(username__iexact=username, tenant=tenant)
+                principal = Principal.objects.select_for_update().get(username__iexact=username, tenant=tenant)
```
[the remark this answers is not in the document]
lpichler: I was assuming this: What if principal record is delete during this transaction by other transaction  ? Is this case ok ?

rbac/management/group/view.py
lpichler: Thanks @alechenninger for explaining! 

Mainly I was thinking if group is deleted during current transaction - do you think it is reason to lock group ?

1. tx1 add user1 start
2. tx2 delete group
3. tx1 sent data to outbox table with group which doesn't exist.

This was my reasoning

alechenninger: Yeah great question. The answer to this is basically the same for the principal case. I think you're talking about this scenario:

Tx1 | Tx2 | Result | Input | Output | Tuple state
-- | -- | -- | -- | -- | --
check if g1 exists |   | ok | g1 | exists |  
add user1 g1 |   | ok | u1, g1 |   |  
add outbox event |   | ok |   |   |  
  | check if g1 exists | ok |   | exists |  
  | remove g1 | ok | g1 |   |  
  | select for update role bindings w/ g1 | ok | g1 | role binding list |  
  | update / remove role bindings | ok | new role binding state |   |  
  | add outbox event | ok |   |   |  
  | commit | ok |   |   | ok
commit |   | ok |   |   | not ok! outbox event adds orphaned group#member@user tuple

I don't think that is actually what would happen. The reason is because of foreign key constraints. 

```
ALTER TABLE "management_group_principals" ADD CONSTRAINT "management_group_pri_group_id_0c1b3b8d_fk_managemen" FOREIGN KEY ("group_id") REFERENCES "management_group" ("id") DEFERRABLE INITIALLY DEFERRED; (params None)

ALTER TABLE "management_group_principals" ADD CONSTRAINT "management_group_pri_principal_id_47b7a6f8_fk_managemen" FOREIGN KEY ("principal_id") REFERENCES "management_principal" ("id") DEFERRABLE INITIALLY DEFERRED; (params None)
```

What should happen is this:

Tx1 | Tx2 | Result | Input | Output | Tuple state
-- | -- | -- | -- | -- | --
check if g1 exists |   | ok | g1 | exists |  
add user1 g1 |   | ok | u1, g1 |   |  
add outbox event |   | ok |   |   |  
  | check if g1 exists | ok |   | exists |  
  | remove g1 | ok | g1 |   |  
  | select for update role bindings w/ g1 | ok | g1 | role binding list |  
  | update / remove role bindings | ok | new role binding state |   |  
  | add outbox event | ok |   |   |  
  | commit | ok |   |   | ok
commit |   | rollback - fk constraint violation |   |   | ok – no event committed (rollback)

When tx1 goes to commit, it shouldn't succeed because of the foreign key constraints b/w the group and the principal. Might be worth confirming this by manually running sql against two concurrent sessions in the db but I think that's what should happen. In that case, as long as the outbox event and the user add are in the same tx (atomic block), the outbox event actually never gets written because the user addition never gets written, either, all without locking.

That said, we can consider locking the group _anyway_ since it's not too expensive and might save us in case there is a scenario we aren't thinking of. But I think we're okay.

For principals, it's the same thing. If a concurrent tx removes the principal and commits, and then we try to refer to that principal with another commit, that will rollback b/c the principal no longer exists.

lpichler:
Currently PR is done.

rbac/management/group/view.py
lpichler: Thanks ! I will leave here locking and principals here in that case.

alechenninger: Not sure I follow – what I am describing is why the lock is not necessary. The first table isn't actually what would happen, I was clarifying the difference between I think what you were expecting vs what I think would actually happen.

It's not a huge deal to leave the locks but shouldn't be necessary. Are you saying you want to leave it just in case?

rbac/management/group/view.py
```diff
@@ -464,15 +472,22 @@ def add_service_accounts(
             if len(invalid_service_accounts) > 0:
                 raise ServiceAccountNotFoundError(f"Service account(s) {invalid_service_accounts} not found.")
 
+    def add_service_accounts(
+        self,
+        group: Group,
+        service_accounts: Iterable[dict],
+        org_id: str = "",
+    ) -> Group:
+        """Add service accounts to the group."""
```
alechenninger, line 481: Looks like the type signature is wrong here 

```suggestion
    ) -> Tuple[Group, List[Principal]]:
```

rbac/management/group/view.py
```diff
@@ -1126,14 +1141,41 @@ def remove_service_accounts(
 
             raise ValueError(f"Service account(s) {service_account_ids_diff} not found in the group '{group.name}'")
 
+        removed_service_accounts = []
         # Remove service accounts from the group.
         with transaction.atomic():
```
alechenninger, line 1146: I _think_ this additional atomic block is unnecessary now

rbac/management/group/view.py
```diff
@@ -1126,14 +1141,41 @@ def remove_service_accounts(
...
             for service_account in valid_service_accounts:
                 group.principals.remove(service_account)
+                removed_service_accounts.append(service_account)
 
         logger.info(
             f"[Request_id:{request_id}] {valid_service_account_ids} "
             f"removed from group {group.name} for org id {org_id}."
         )
         for username in service_accounts:
             group_principal_change_notification_handler(self.request.user, group, username, "removed")
```
alechenninger, line 1156: One note around the notification handlers: previously these happend _after_ the writes committed. So the semantics you had prior to this were that you might not send a message (due to failure) even if you successfully committed something.

With this change, that's different. Now it's the other way around: because we haven't committed yet, we send the message regardless of whether our commit rolls back, which means you might instead send a message when you shouldn't have.

So it's whether you want to err on the side of sending when you shouldn't, vs not sending when you should've. I'm not sure which is better–maybe even this is an improvement–but just something to keep in mind for edge cases.

rbac/management/group/view.py
```diff
@@ -1126,14 +1141,41 @@ def remove_service_accounts(
...
+
+        return removed_service_accounts
+
+    def _get_add_principals_queryset(self):
+        """Obtain queryset for requesting user based on access to add principals."""
+        return Group.objects.filter(tenant=self.request.tenant).select_for_update()
```
alechenninger, line 1162: Do we need to port anything else from `get_group_queryset` here?

There are various filters that take into account user access and query params. I'm not sure if those are relevant here, but we did end up porting some of that over for the Role view at least.

This is maybe one cost of locking – if we don't really need to, we could consider keeping the same query and we wouldn't have to run a second query, either.

alechenninger (commented):
Nothing major, just a few questions / small things. Looking good!

rbac/management/group/view.py
lpichler: yes I understand what did you describe but according to statement and I also confirmed it with 2 sessions.
in TX1, it tries to insert a group ID that no longer exists because it was deleted by transaction TX2, resulting in an error due to a constraint violation

I intended to leave it here because of this sentence:

> That said, we can consider locking the group anyway since it's not too expensive and might save us in case there is a scenario we aren't thinking of. But I think we're okay.

Now I understand that you prefer not to use locking for groups and principals, 
so I will remove it. 

NOTE: With locks, the only difference is that TX2 will wait for TX1, but most likely it will fail because TX2 is attempting to delete a group that now has a reference in the management_group_principals table, which was added by TX1.

rbac/management/group/view.py
lpichler: good catch, thanks!

rbac/management/group/view.py
lpichler: This is good point to consider - in addition if sending into kafka will fail - I guess it will rollback whole transaction. Which is probably fine as it means that debezium will also not work :) 

(I will keep as it is - as I see role dual write has same change   )

rbac/management/group/view.py
lpichler: yes I am going to remove locking as I mentioned in previous commit.

rbac/management/group/view.py
lpichler: Thanks, done 👍

lpichler:
Thanks @alechenninger, I added your suggestions 👍

rbac/management/group/view.py
alechenninger: Thanks for clarifying and confirming locally!

I guess one another small downside to locking is that if there are other transactions trying to write to the group in some way, those will block while the group is locked (even if they are unrelated to dual write, e.g. group name or something like that). A small thing, though!

rbac/management/group/view.py
alechenninger: That's a good point. It means if Kafka is down, more functionality is impacted (severely!). I suppose that is another reason to consider moving notifications to use the outbox pattern, too. Not only would it be atomic, but it would also be more available (and probably reduce latencies on API calls, too).

We may want to reconsider the role changes that brought that into the TX as well, but I'll open an issue for it and we can discuss with the team.

alechenninger:
@lpichler Cool I think this is good to 🚢 then! I think I will wait for #1199 to be reviewed & merged before merging this one to keep that review isolated.

rbac/management/group/relation_api_dual_write_group_handler.py
```diff
@@ -0,0 +1,110 @@
...
+
+class RelationApiDualWriteGroupHandler:
+    """Class to handle Dual Write API related operations."""
+
+    def __init__(
+        self,
+        group,
+        event_type: ReplicationEventType,
+        principal_usernames,
+        service_accounts,
```
alechenninger, line 43: Since these are both just `List[Principal]` what do you think about combining the parameters? It makes it a little easier to use (e.g. in tests) because I don't have to explicitly sort users vs services ahead of time, and its not actualy needed as far as I can tell since we combine them into one below (`for principal in self.principal_usernames + self.service_accounts`).

alechenninger (commented):
Small minor things but with or without LGTM!

(Note: I pushed up some more tests and updates to fixtures in test_dual_write – just a way to test more scenarios. In the future we may be able to use some of those cases to also test against full stack relations + spicedb checks.)

rbac/management/group/view.py
```diff
@@ -668,20 +681,37 @@ def principals(self, request: Request, uuid: Optional[UUID] = None):
...
-                resp = self.add_principals(group, principals, org_id=org_id)
+                proxy_response = self.validate_principals_in_proxy_request(principals, org_id=org_id)
+                if len(proxy_response.get("data", [])) > 0:
+                    principals_from_response = proxy_response.get("data", [])
+                if isinstance(proxy_response, dict) and "errors" in proxy_response:
+                    return Response(status=proxy_response["status_code"], data=proxy_response["errors"])
+
+            with transaction.atomic():
+                new_service_accounts = []
+                group = self.get_object()
```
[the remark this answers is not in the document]
lpichler: you are right, it was not intent (probably leftover after I removed locking groups), good catch!

rbac/management/group/relation_api_dual_write_group_handler.py
lpichler: I did not combine this initially because I wasn't sure if we would need separate IDs for the service account and principals but as you mentioned it seems unnecessary for now. Anyway If it becomes necessary later, we can handle it in the dual write handler.

I updated also your tests and I removed service account creation as well - but if you intended to keep it for later use  we can revert/remove last commit.

lpichler:
@alechenninger I have addressed your latest suggestions, thanks!

rbac/management/group/relation_api_dual_write_group_handler.py
alechenninger: K cool! Let's keep the service account creation support in the test fixture – that is still a use case we want to test explicitly I think, regardless of the implementation.

As for IDs, a fair point. I am thinking both will end up using "user_id" field being added in #1179 though, so we might be okay.

alechenninger (approved):
Ship it!

PR #1205, opened 2024-09-18
Log when User/SA/Roles is added to a group

Ellen-Yi-Dong:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-34891

The what, why and how.

Log when a user/principal, service account, or role is added to a group

## Local Testing

- Added unit tests to test changes

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


rbac/management/audit_log/view.py
```diff
@@ -33,7 +33,13 @@ class AuditLogViewSet(mixins.ListModelMixin, viewsets.GenericViewSet):
     serializer_class = AuditLogSerializer
     permission_classes = (AuditLogAccessPermission,)
 
+    def filter_by_id(self, queryset):
+        """Filter queryset by id."""
+        return queryset.order_by("-id").values()
```
petracihalova, line 38: I am not sure if I understand what is the intention here ... by the method name and docstring it looks like something is supposed to be filtered but the queryset is sorted

rbac/management/audit_log/view.py
```diff
@@ -33,7 +33,13 @@ class AuditLogViewSet(mixins.ListModelMixin, viewsets.GenericViewSet):
...
+
     def list(self, request, *args, **kwargs):
         """List all of the audit logs within database by tenant."""
         self.queryset = filter_queryset_by_tenant(AuditLog.objects.all(), request.tenant)
+        if request.query_params.get("order_by") is not None:
+            self.queryset = self.filter_by_id(self.queryset)
```
petracihalova, line 44: I was not able to find any open api documentation to that view, it is in progress still? it is related with my previous comment, what is the expected usage of this `order_by` query param? because now the value is not validated, this only checks if that param is included or not, but maybe I dont see the context 😇

petracihalova (commented):
nice job 👍  I have few comments

rbac/management/audit_log/view.py
Ellen-Yi-Dong: Oh! I completley meant to be ordering by id! I wanted to have it ordered by ID so when I was testing the audit logs, I was able to see the most recent audit log appear on top so that I didn't need to go looking for it!! 
I'll update the method name and docstring!! 😅

rbac/management/group/view.py
```diff
@@ -1112,7 +1199,10 @@ def remove_service_accounts(
 
         # Get the group's service accounts that match the service accounts that the user specified.
         valid_service_accounts = Principal.objects.filter(
-            group=group, tenant=tenant, type="service-account", service_account_id__in=service_accounts
+            group=group,
+            tenant=tenant,
+            type="service-account",
```
astrozzc, line 1204: Let's use TYPE_SERVICE_ACCOUNT for this, I have a PR ongoing to replace these random type names with enums. So I can remember replacing this in my PR too

astrozzc (approved):
LGTM

rbac/management/audit_log/model.py
```diff
@@ -138,3 +153,17 @@ def log_edit(self, request, resource, object):
         self.action = AuditLog.EDIT
         self.tenant_id = self.get_tenant_id(request)
         super(AuditLog, self).save()
+
+    def log_add(self, request, resource, object, type_dict, user_type):
```
lpichler, line 157: maybe better name ?  it is not visible difference between `log_add` and `log_create`. 
something like "log_role_or_user_added_to_group" or `log_group_membership` or `log_group_assigment`.

astrozzc:
> @astrozzc we should discuss when this PR can be merged as it conflicts with some dual write PR as #1198,..

Oh, good point. That one is about to be merged, right? Let's merge that one first before we merge this one. That one is too big to do a rebase

lpichler:
> > @astrozzc we should discuss when this PR can be merged as it conflicts with some dual write PR as #1198,..
> 
> Oh, good point. That one is about to be merged, right? Let's merge that one first before we merge this one. That one is too big to do a rebase

yes it is merged now but into https://github.com/RedHatInsights/insights-rbac/pull/1198, this PR also contains migration. I am not sure what it is the best approach here.

if we will merge this, we would need to solve conflicts in https://github.com/RedHatInsights/insights-rbac/pull/1198 and also change names of migrations.

rbac/management/audit_log/model.py
Ellen-Yi-Dong: Oh! There's another task to add permissions to a role, so I was thinking of having this ` log_add ` function do that as well! I just have it for now grouped this way to split the prs.

rbac/management/audit_log/view.py
petracihalova: thank you Ellen, is there some plan to add these into open api spec as well? the endpoint `/api/rbac/v1/auditlogs/` is working right and can be used?

Ellen-Yi-Dong: Yes, the endpoint is working!! 
I do plan to add them into the open api spec!! Thank you for reminding me about it!! I'll update it within the next PR!!

petracihalova: awesome, thank you

Ellen-Yi-Dong:
For now, I think I'll delete the migration and also update this PR with also adding the permissions to a role because of the log_add function if that's alright?

rbac/management/audit_log/model.py
```diff
@@ -80,9 +80,8 @@ def get_resource_item(self, r_type, request, *args, **kwargs):
             group_object_name = "group: " + group_object.name
             return group_object_id, group_object_name
 
-        elif r_type == AuditLog.PERMISSION:
-            # TODO: update for permission related items
-            return None
+        else:
+            return TypeError("Wrong Resource Type")
```
petracihalova, line 84: I would prefer here the **ValueError** because we got an unexpected value of the resource type, the TypeError is more suitable for values that are not correct type, e.g. you expected a number and got a string

rbac/management/audit_log/model.py
```diff
@@ -98,6 +97,21 @@ def find_edited_field(self, resource, resource_name, request, object):
...
+        if user_type == AuditLog.USER:
+            for i in type_dict:
+                names_list.append(i["username"])
+        elif user_type == "service_account":
+            for i in type_dict:
+                names_list.append(i["clientId"])
+        elif user_type == AuditLog.ROLE:
+            names_list = type_dict
+        else:
+            return NameError("User type does not exist")
```
petracihalova, line 112: I would prefer here the ValueError because the NameError is more suitable for scenarios when you try to use/call some variable that is not defined

petracihalova (changes requested):

petracihalova (approved):

vbelchio:
@Ellen-Yi-Dong When I tried to add an inexistent Role to an existent Group, we got this logged into `auditlog` table, probably we need to validate if the role exists before logging it.

rbac/management/audit_log/model.py
```diff
@@ -98,6 +97,22 @@ def find_edited_field(self, resource, resource_name, request, object):
                 description = description + "edited access (permissions/resources)"
         return description
 
+    def find_specific_list_of_users(self, type_dict, user_type):
+        """Create list of principals/roles/service accounts for description."""
+        names_list = []
+        if user_type == "user" or user_type == "User":
+            for i in type_dict:
```
astrozzc, line 104: Nit: can we use a more meaningful name to replace these "i" here?
Such as principal_data, role_data, etc.

astrozzc (approved):

vbelchio (approved):
LGTM

But we need to solve the related bug below to have the best behavior of the audit logs and the process to add a non-existent role to a group.

JIRA Bug: https://issues.redhat.com/browse/RHCLOUD-35765

lpichler (approved):

rbac/management/group/view.py
```diff
@@ -771,6 +785,27 @@ def principals(self, request: Request, uuid: Optional[UUID] = None):
             # Serialize the group...
             output = GroupSerializer(group)
             response = Response(status=status.HTTP_200_OK, data=output.data)
+
+            if status.is_success(response.status_code):
+                if len(principals) > 0:
```
lpichler, line 790: this should be part of transaction block which start on 715.
you can put audit log creation under those related conditions (`len(service_accounts|principals) > 0`).

if operation fails, transaction is rolled back and record into audit log is not created.

rbac/management/audit_log/model.py
```diff
@@ -138,3 +153,17 @@ def log_edit(self, request, resource, object):
         self.action = AuditLog.EDIT
         self.tenant_id = self.get_tenant_id(request)
         super(AuditLog, self).save()
+
+    def log_group_assignment(self, request, resource, object, type_dict, user_type):
```
lpichler, line 157: resource and object names are confusing 

it looks that resource is rather `resource_type` and then object can be `resource` :) 

Does it make sense ?

lpichler, line 157: `user_type` could be user, service or role.  What about name `target_resource_type` or `assigned_resource_type` ? or something like that

rbac/management/group/view.py
```diff
@@ -1083,10 +1132,23 @@ def roles(self, request, uuid=None, principals=None):
...
+
             response_data = GroupRoleSerializerIn(group)
+            response = Response(status=status.HTTP_200_OK, data=response_data.data)
+            if status.is_success(response.status_code):
+                auditlog = AuditLog()
+                auditlog.log_group_assignment(
+                    request,
+                    AuditLog.GROUP,
+                    object=group,
+                    type_dict=response_data.data,
```
lpichler, line 1148: what about to pass just values for description ? Something like:
`[response_data.data["uuid"] for role in response_data.data]`

then type_dict could be renamed to something like `values_for_description`
and it will allow to remove method `self.find_specific_list_of_users` and it is possible to join(",") in log assignment method.

vbelchio:
QE: Passed (manual)

PR #1206, opened 2024-09-19
Ensure we allow `application/problem+json` as a valid `Accept` header value in v2

coderbydesign:
## Link(s) to Jira
- RHCLOUD-35321

Since we are starting to use `application/problem+json` as a response content type in v2 APIs [1], we need RBAC/Django to accept this as a valid `Accept` header value, given clients built off the API spec will be sending this.

Currently, if you send a request to `/api/rbac/v2/workspaces/<uuid>/ -XDELETE -H 'Accept: application/problem+json'` which is what a client built off the spec will do (with the `Accept` header), you'll get an error from RBAC: "Not acceptable 406".

We _do_ set the correct `content-type` to align with the spec [2], however Django complains becasue it can't find the `application/problem+json` renderer.

This implements a custom renderer in Django Rest Framework [3] on the v2 views, specifically workspaces now, specifying it in the `renderer_classes` of the view, and setting the `media_type` (used to match the `Accept` value(s) to a renderer) and `format` (response) properties accordingly.

The `DELETE` test was modified prior to this change to add `Accept: application/problem+json` to confirm it failed (it did), and now succeeds with the changes in place.

[1] [https://github.com/RedHatInsights/insights-rbac/blob/master/docs/source/specs/v2/openapi.v2.yaml]
[2] https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/api/common/exception_handler.py#L93
[3] [https://www.django-rest-framework.org/api-guide/renderers/#custom-renderers]

## Local Testing
To exploit the bug on `main`, which also presents through a generated API client, send the following request:

```bash
curl http://localhost:8000/api/rbac/v2/workspaces/<uuid>/ -XDELETE -H 'Accept: application/problem+json'
```

You'll see a 406 response. Now check this branch out and run the same curl. You should also see `application/json`, `application/problem+json` accordingly, depending on success/failure of the requests.

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/management/workspace/view.py
+++ b/rbac/management/workspace/view.py
@@ -19,13 +19,12 @@ import json
 
 from django.utils.translation import gettext as _
 from django_filters import rest_framework as filters
+from management.base_viewsets import BaseV2ViewSet
 from management.permissions import WorkspaceAccessPermission
 from management.utils import validate_uuid
-from rest_framework import mixins, serializers, viewsets
+from rest_framework import serializers
 from rest_framework.filters import OrderingFilter
-from rest_framework.settings import api_settings
 
-from api.common.renderers import ProblemJSONRenderer
 from .model import Workspace
 from .serializer import WorkspaceSerializer
 
@@ -34,14 +33,7 @@ REQUIRED_PUT_FIELDS = ["name", "description", "parent_id"]
 REQUIRED_CREATE_FIELDS = ["name"]
 
 
-class WorkspaceViewSet(
-    mixins.CreateModelMixin,
-    mixins.DestroyModelMixin,
-    mixins.RetrieveModelMixin,
-    mixins.UpdateModelMixin,
-    mixins.ListModelMixin,
-    viewsets.GenericViewSet,
-):
+class WorkspaceViewSet(BaseV2ViewSet):
     """Workspace View.
 
     A viewset that provides default `create()`, `destroy` and `retrieve()`.
@@ -52,7 +44,6 @@ class WorkspaceViewSet(
     queryset = Workspace.objects.annotate()
     lookup_field = "uuid"
     serializer_class = WorkspaceSerializer
-    renderer_classes = api_settings.DEFAULT_RENDERER_CLASSES + [ProblemJSONRenderer]
     ordering_fields = ("name",)
     ordering = ("name",)
     filter_backends = (filters.DjangoFilterBackend, OrderingFilter)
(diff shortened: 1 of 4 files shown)


rbac/management/workspace/view.py
```diff
@@ -50,6 +52,7 @@ class WorkspaceViewSet(
     queryset = Workspace.objects.annotate()
     lookup_field = "uuid"
     serializer_class = WorkspaceSerializer
+    renderer_classes = api_settings.DEFAULT_RENDERER_CLASSES + [ProblemJSONRenderer]
```
coderbydesign, line 55: This ensures we pull in the default JSON renderer and just add the custom problem renderer. We'll need to do this for each v2 view, _or_ design a better way to inherit from a v2 viewset, have middleware plug this in based on path, etc. I figured this was probably the easiest for now, and we can revisit if folks think we should refactor when we introduce more views.

coderbydesign: Look to see if we can have a v2 viewset to inherit from.

tests/management/workspace/test_view.py
```diff
@@ -410,9 +426,12 @@ def test_delete_workspace(self):
 
         url = reverse("workspace-detail", kwargs={"uuid": workspace.uuid})
         client = APIClient()
-        response = client.delete(url, None, format="json", **self.headers)
+        test_headers = self.headers.copy()
+        test_headers["HTTP_ACCEPT"] = "application/problem+json"
```
coderbydesign, line 430: Look at adding a v2 identity request which inherits from v1 to add headers.

vbelchio (approved):
LGTM
QE: Passed

tests/management/workspace/test_view.py
alechenninger: One interesting thing about this I guess is that the headers are dependent on the request. So like if we added both accept media types unconditionally, I guess we wouldn't have reproduced the bug.

I wonder if we should try and use the generated client in the tests somehow? (If so this is very likely a future reliability topic I guess)

coderbydesign: @alechenninger yeah, good point. 

that's actually something I've brought up in the past for PR check/IQE tests, around potentially building clients on PR changes and having a small suite that would test client changes based on spec changes and run through some tests.

astrozzc (approved):
LGTM

tests/management/workspace/test_view.py
coderbydesign: Added a shell ticket here: https://issues.redhat.com/browse/RHCLOUD-35416

PR #1211, opened 2024-09-25
[RHCLOUD-34787] Dual Write: Generate replication event for adding and removing roles from group

lpichler:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-34787

- this generates replication event during add/remove roles from group 
- PR also adds method to generate replication event for permissions relations of system role (used only for tests)

## Local Testing
```
POST http://localhost:8000/api/rbac/v1/groups/6277ca72-6b65-463e-bc3d-e0d47481adee/roles/
x-rh-identity: 

{"roles":["b7d2f47a-6042-433c-9d43-8873b3517a65"]}
```

```
### DELETE  ROLE to group
DELETE http://localhost:8000/api/rbac/v1/groups/6277ca72-6b65-463e-bc3d-e0d47481adee/roles/?roles=b7d2f47a-6042-433c-9d43-8873b3517a65
x-rh-identity: 

```

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/management/group/definer.py
+++ b/rbac/management/group/definer.py
@@ -17,6 +17,7 @@
 
 """Handler for system defined group."""
 import logging
+from typing import Union
 from uuid import uuid4
 
 from django.db import transaction
@@ -113,13 +114,8 @@ def clone_default_group_in_public_schema(group, tenant):
 @transaction.atomic
 def add_roles(group, roles_or_role_ids, tenant, user=None):
     """Process list of roles and add them to the group."""
-    if not isinstance(roles_or_role_ids, QuerySet):
-        # If given an iterable of UUIDs, get the corresponding objects
-        roles = Role.objects.filter(uuid__in=roles_or_role_ids)
-    else:
-        roles = roles_or_role_ids
+    roles = _roles_by_query_or_ids(roles_or_role_ids)
     group_name = group.name
-    role_names = list(roles.values_list("name", flat=True))
 
     group, created = Group.objects.get_or_create(name=group_name, tenant=tenant)
     system_policy_name = "System Policy for Group {}".format(group.uuid)
@@ -130,12 +126,12 @@ def add_roles(group, roles_or_role_ids, tenant, user=None):
     if system_policy_created:
         logger.info(f"Created new system policy for tenant {tenant.org_id}.")
 
-    system_roles = Role.objects.filter(tenant=Tenant.objects.get(tenant_name="public"), name__in=role_names)
+    system_roles = roles.filter(tenant=Tenant.objects.get(tenant_name="public"))
 
     # Custom roles are locked to prevent resources from being added/removed concurrently,
     # in the case that the Roles had _no_ resources specified to begin with.
-    # This should not be necessary for system roles
-    custom_roles = Role.objects.filter(tenant=tenant, name__in=role_names).select_for_update()
+    # This should not be necessary for system roles.
+    custom_roles = roles.filter(tenant=tenant).select_for_update()
 
     for role in [*system_roles, *custom_roles]:
         # Only Organization administrators are allowed to add the role with RBAC permission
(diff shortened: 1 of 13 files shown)


rbac/management/group/relation_api_dual_write_group_handler.py
```diff
@@ -141,9 +141,15 @@ def replicate_added_role(self, role: Role):
             # There is a risk of phantom reads here, but this is why we lock the custom Role
             bindings: Iterable[BindingMapping] = BindingMapping.objects.filter(role=role).select_for_update()
 
-            for mapping in bindings:
-                self.group_relations_to_add.append(mapping.add_group_to_bindings(str(self.group.uuid)))
-                mapping.save(force_update=True)
+            if sum(1 for _ in bindings) == 0:
+                mapping = BindingMapping.for_group_and_role(role, str(self.group.uuid), self.group.tenant.org_id)
```
lpichler, line 145: Should we create binding mappings for the role if none exist, similar to how it is done for system roles?

rbac/management/group/definer.py
```diff
@@ -144,14 +157,62 @@ def add_roles(group, roles_or_role_ids, tenant, user=None):
...
+            # Intentionally queried this way to avoid use of cached binding_mappings
+            # There is a risk of phantom reads here, but this is why we lock the custom Role
+            bindings: Iterable[BindingMapping] = BindingMapping.objects.filter(role=role).select_for_update()
+            for mapping in bindings:
+                tuples_to_add.append(mapping.add_group_to_bindings(group.uuid))
+                mapping.save(force_update=True)
 
+        # Send notifications
+        # TODO: maybe use top-level method instead so replicator can be modified for tests
+        OutboxReplicator(role).replicate(
```
lpichler: Is intent to have one replication event for each role change ?
pros:
- payload will not grow 
cons:
- probably harder failure handling when one event will fail in source connector ?

rbac/management/group/relation_api_dual_write_group_handler.py
```diff
@@ -107,3 +110,58 @@ def _replicate(self):
...
+        self._replicate()
+
+    def replicate_removed_role(self, role: Role):
+        if not self.replication_enabled():
+            return
+        # This logic can be the same for system vs custom b/c we'd never create a binding in this path
+        bindings: Iterable[BindingMapping] = role.binding_mappings.select_for_update().all()
+        for binding in bindings:
+            self.group_relations_to_remove.append(binding.remove_group_from_bindings(str(self.group.uuid)))
+            if not binding.is_empty():
```
lpichler, line 162: Is it enough condition(only empty group) to remove whole binding mapping  ? Should not we extend it and check whether permissions are empty as well ?
so together:
```
  def is_empty(self):
        return len(self.mappings["groups"]) == 0 && len(self.mappings["permissions"]) == 0 
```

rbac/management/group/relation_api_dual_write_group_handler.py
alechenninger: No, because binding mappings are per role and resource, the mappings would already exist for the role. This would not calculate/persist anything that would and should have already been calculated or persisted. If there are no existing role bindings, it means there is no access defined for the role. When access is later defined for the role, policies will be looked up then and the groups should be added. Would be good to add a test case for this in the test_dual_write test cases.

rbac/management/group/relation_api_dual_write_group_handler.py
alechenninger: If the permissions are empty the role binding should have already been removed by the role update.

However it does beg the question if we should _only_ remove the binding in role updates. We probably should actually, due to the answer I gave above. So I think this empty and removal check is an error – we should **not** remove bindings in this path for **custom roles.**

For **system roles** however, we **can** remove bindings when there are no more groups, because they will be added back when the group is added.

rbac/management/group/definer.py
```diff
@@ -144,31 +149,41 @@ def add_roles(group, roles_or_role_ids, tenant, user=None):
...
-        if not system_policy.roles.filter(pk=role.pk).exists():
-            system_policy.roles.add(role)
+        if system_policy.roles.filter(pk=role.pk).exists():
+            continue
 
-            # Send notifications
-            group_role_change_notification_handler(user, group, role, "added")
+        system_policy.roles.add(role)
 
+        dual_write_handler = RelationApiDualWriteGroupHandler(group, ReplicationEventType.ASSIGN_ROLE, [])
```
alechenninger, line 159: Since the same handler is used for group membership as well as group roles, what do you think about removing the principals from the constructor and instead passing them to the specific methods (replicate_new_principals / replicate_removed_principals)?

alechenninger (commented):
See slack for commentary on my commits and other thoughts. 

TLDR – one comment here, otherwise think we should just add some group view tests that they trigger replication.

lpichler:
@alechenninger I added view tests and the other suggestion 👍

rbac/management/group/definer.py
```diff
@@ -144,31 +149,41 @@ def add_roles(group, roles_or_role_ids, tenant, user=None):
...
 def remove_roles(group, roles_or_role_ids, tenant, user=None):
     """Process list of roles and remove them from the group."""
-    if not isinstance(roles_or_role_ids, QuerySet):
-        # If given an iterable of UUIDs, get the corresponding objects
-        roles = Role.objects.filter(uuid__in=roles_or_role_ids)
-    else:
-        roles = roles_or_role_ids
-    role_names = list(roles.values_list("name", flat=True))
-
+    roles = _roles_by_query_or_ids(roles_or_role_ids)
```
alechenninger, line 168: I didn't realize I slightly changed the query in this case from the prior code. Previously it didn't bother filtering the tenant. I assume this was because if you're removing a role, it must be because the role was allowed to be there in the first place, so it doesn't really do anything if the role happens to be a role from another tenant. It would be a no-op. I guess it doesn't matter, but just pointing this out in case anyone thinks there are any issues with this.

alechenninger:
LGTM! Great work & collab on this one

alechenninger (approved):

rbac/management/group/definer.py
alechenninger: I mispoke – the query is the same.

PR #1214, opened 2024-09-27
Workspace types

coderbydesign:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-35289

## Description of Intent of Change(s)
This adds a `type` column on the `Workspace` model in order to identify in the
code and API, which type of workspace an object is:

- `root`: top-level workspace for a tenant
- `default`: the default workspace for a tenant
- `standard`: any custom non-root/default workspace

Constraints are added to ensure a single root and single default workspace per tenant.

The `type` field is also return on workspace responses (but disallowed from create and update operations via the API):
```json
{
  "uuid": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
  "parent_id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
  "type": "root",
  "name": "Workspace A",
  "description": "Description of Workspace A",
  "created": "2024-09-27T18:22:39.738Z",
  "modified": "2024-09-27T18:22:39.738Z"
}
```

## Local Testing
How can the feature be exercised?
How can the bug be exploited and fix confirmed?
Is any special local setup required?

## Checklist
- [x] if API spec changes are required, is the spec updated?
- [ ] are there any pre/post merge actions required? if so, document here.
- [x] are theses changes covered by unit tests?
- [ ] if warranted, are documentation changes accounted for?
- [ ] does this require migration changes?
  - [ ] if yes, are they backwards compatible?
- [ ] is there known, direct impact to dependent teams/components?
  - [ ] if yes, how will this be handled?

## Secure Coding Practices Checklist Link
- https://github.com/RedHatInsights/secure-coding-checklist

## Secure Coding Practices Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [ ] General Coding Practices

PR #1215, opened 2024-09-27
Workspace types api

coderbydesign:
_**Note:** this is a PR into the feature branch for #1214 until it's merged_

## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-35291

[Add type query param to Workspace spec for list](https://github.com/coderbydesign/insights-rbac/commit/3ff019e08ae07edf73c532ed5046749e12313bd9) 

This surfaces an enum in the API spec for `GET` on `/workspaces/` list:

- "all"
- "standard"
- "root"
- "default"

Where no `?type=` value, or `?type=all` signifies returning all workspace records.

[Add support for /workspaces/?type=all|standard|default|root query param](https://github.com/coderbydesign/insights-rbac/commit/04a01cb3af5dafd5e7d300c4ba557c7ca82b241e) 

When not supplied, or `all` is supplied on the `type` query param, return all
records, otherwise validate and subsequently use the `type` param value to
filter the queryset accordingly.

## Local Testing
Set your local data up to have multiple types of workspaces, and query with `/workspaces/?type=` with the correct enum values and note the response objects matching the type you filtered by.

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/docs/source/specs/typespec/main.tsp
+++ b/docs/source/specs/typespec/main.tsp
@@ -222,7 +222,7 @@ namespace Workspaces {
         @query offset?: int64 = 0;
 
         @doc("Defaults to all when param is not supplied.")
-        @query type?: WorkspaceTypesQueryParam;
+        @query type?: WorkspaceTypesQueryParam = WorkspaceTypesQueryParam.all;
     ): WorkspaceListResponse | Problems.CommonProblems;
 
     @doc("Create workspace in tenant")
--- a/docs/source/specs/typespec/main.tsp
+++ b/docs/source/specs/typespec/main.tsp
@@ -102,6 +102,11 @@ namespace Workspaces {
         "standard"
     }
 
+    enum WorkspaceTypesQueryParam {
+        "all",
+        ...WorkspaceTypes
+    }
+
     model Workspace {
         @key uuid: UUID;
         parent_id?: UUID;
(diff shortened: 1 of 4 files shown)


astrozzc (approved):
LGTM!
Just wondering if we need the explicit "all" at all.
Maybe just in the request saying if you don't specify specific type, all types would be returned. Then we don't need the extra all and those logic

docs/source/specs/typespec/main.tsp
```diff
@@ -215,6 +220,9 @@ namespace Workspaces {
     @get op list(
         @query limit?: int64 = 10;
         @query offset?: int64 = 0;
+
+        @doc("Defaults to all when param is not supplied.")
+        @query type?: WorkspaceTypesQueryParam;
```
alechenninger, line 225: What do you think of adding the explicit default like this?
```suggestion
        @query type?: WorkspaceTypesQueryParam = WorkspaceTypesQueryParam.all;
```

alechenninger (approved):
Left one minor suggestion but LGTM!

docs/source/specs/typespec/main.tsp
coderbydesign: Well...I actually _tried_ to add this as a default the other day but must have been doing it wrong, because it wasn't compiling and I thought it was because of my custom enum. This works though, so updating! 😄

PR #1217, opened 2024-10-01
Updated format for error message when trying to add a group that alre…

EvanCasey13:
…ady exists for a tenant

## Link(s) to Jira
https://issues.redhat.com/browse/RHCLOUD-31878

Updated the Format of the message returned when trying to add a duplicate group to a tenant

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


rbac/management/utils.py
```diff
@@ -280,6 +280,12 @@ def validate_group_name(name):
         message = f"{name} is reserved, please use another name."
         raise serializers.ValidationError({key: _(message)})
 
+def check_group_exists(groupName):
+    groupTenant = Group.objects.get(name=groupName)
```
lpichler, line 284: can you get tenant from request  ?
something like (in create method)

```
tenant = self.request.tenant
```

rbac/management/utils.py
```diff
@@ -280,6 +280,12 @@ def validate_group_name(name):
...
+    if Group.objects.filter(name=groupName).exists():
```
lpichler, line 285: We need to include the tenant in this query; otherwise, it will return results for other tenants.

Alternatively, we could execute a single query that returns the group object (or None), and then you can extract the tenant object from the group object and use org_id in error message. (org_id is known for user, tenant_id is rather internal RBAC thing )

lpichler:
@EvanCasey13 can you add test for the view ? it should in tests/mgmnt/groups....  
you can runt tests by ` tox -r` (all tests + linters)

or specific test:

```
tox -- tests/management/group -k test_add_group_roles_system_policy_create_success
```

rbac/management/utils.py
alechenninger: You might want to instead catch a unique constraint violation when the create happens. This does a few things for you:

1. It fixes the query (would only error for the same tenant)
2. It catches race conditions (this check would not catch if a group was concurrently added, you will still get the constraint violation error)
3. It eliminates a second query (only does the one insert, rather than read + insert)

But, up to y'all :) 🤘🏻

PR #1219, opened 2024-10-02
Enable support for read-only mode on APIs

coderbydesign:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-35526

This adds an optional `READ_ONLY_API_MODE` environment variable consumed in settings, which defaults to `False`. It can be configured via params in the deploy template to enable/disable with config per environment.

This is used in a new `ReadOnlyApiMiddleware` middleware for all RBAC requests, which checks the request method and the configuration setting to determine whether or not to process the request.

In the event of a write method + read-only mode being enabled, we'll return a 405 with a message explaining we're in read-only mode, to be used during migration/maintenance windows for the service.

Another option would be to use 503, though since we're still accepting GET requests a 405 with explicit message may be the best approach.

## Local Testing
By default, you should not be prevented from any write API requests.
Update your `.env` to set `READ_ONLY_API_MODE=True` and note that you receive a 405 on anything but `GET` requests.

## TODO:
- Update API specs?

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


rbac/rbac/middleware.py
```diff
@@ -420,3 +420,16 @@ def process_request(self, request):  # pylint: disable=no-self-use
...
+
+
+class ReadOnlyApiMiddleware(MiddlewareMixin):  # pylint: disable=too-few-public-methods
+    """Middleware to enable read-only on APIs when configured."""
+
+    def process_request(self, request):  # pylint: disable=no-self-use
+        """Process request ReadOnlyApiMiddleware."""
+        if settings.READ_ONLY_API_MODE and request.method in ["POST", "PUT", "PATCH", "DELETE"]:
+            return HttpResponse(
+                json.dumps({"error": "This API is currently in read-only mode. Please try again later."}),
```
astrozzc, line 432: Wondering what message would pop up in customer's console, they might be confused with API.
Maybe something like the system is currently under maintenance and in read-only mode......

astrozzc (approved):
LGTM!

rbac/rbac/middleware.py
coderbydesign: 👍 I was going to ping platex to see how a 405 might be handled right now to see if they'd render their own message or need to use the API message.

rbac/rbac/middleware.py
```diff
@@ -420,3 +420,16 @@ def process_request(self, request):  # pylint: disable=no-self-use
 
         """
         setattr(request, "_dont_enforce_csrf_checks", True)
+
+
+class ReadOnlyApiMiddleware(MiddlewareMixin):  # pylint: disable=too-few-public-methods
+    """Middleware to enable read-only on APIs when configured."""
+
+    def process_request(self, request):  # pylint: disable=no-self-use
```
lpichler, line 428: As I checked docs it looks that django style for middleware was changed 
[example](https://docs.djangoproject.com/en/4.2/topics/http/middleware/#writing-your-own-middleware).

but we would need to change it for all middlewares - so maybe this will be better to do it in separate (probably tech debt) task - so I think it is OK 👍

lpichler (approved):

rbac/rbac/middleware.py
coderbydesign: Oh, interesting! Thanks for pointing that out. Yeah, I'd agree that since this pattern is currently used throughout it would be a good candidate for tech debt.

coderbydesign: Added https://issues.redhat.com/browse/RHCLOUD-35536

lpichler: Thanks!!!

PR #1228, opened 2024-10-07
Add CJI to help migration

astrozzc:
## Link(s) to Jira
-

The what, why and how.

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


alechenninger:
@astrozzc Looking good! You might want to rebase on top of dual write for role branch and change the merge base to that. Some of the signatures changed. There are definitely a lot of connections to this one and the other PR as well, so we may want to just work on the same branch there even and we can collab on that if you want. Alternatively, you could wait for that to be merged, or rebase on top of that branch (even though it is WIP, it is reasonably far along).

alechenninger:
Otherwise the PR looks good!

rbac/migration_tool/migrate.py
```diff
@@ -211,3 +212,19 @@ def migrate_data(exclude_apps: list = [], orgs: list = [], write_relationships:
...
+    relationships.append(create_relationship("tenant", str(tenant.org_id), "user", user_id, "member"))
+    group_default = tenant.group_set.filter(platform_default=True).first()
+    if not group_default:  # Means it is not custom platform_default
+        group_default = DEFAULT_GROUP
+    relationships.append(create_relationship("group", str(group_default.uuid), "user", user_id, "member"))
+    if org_admin:
+        admin_default = tenant.group_set.filter(admin_default=True).first()
+        if not admin_default:
+            admin_default = DEFAULT_ADMIN_GROUP
+        relationships.append(create_relationship("group", str(admin_default.uuid), "user", user_id, "member"))
```
lpichler, line 229: I guess namespace is missing, but maybe also there are some methods in dual write group handler exists. or maybe also this https://github.com/RedHatInsights/insights-rbac/blob/c169e44d6caa2e72b096024387a1b72f4e025abd/rbac/management/group/model.py#L56.

astrozzc: Yea, I removed these and reusing Alec's bootstrapping tenants

rbac/management/tenant_service/v2.py
```diff
@@ -103,6 +104,94 @@ def update_user(
...
+        # Ignore those already created
+        Workspace.objects.bulk_create(workspaces, ignore_conflicts=True)
+
+        mappings = TenantMapping.objects.bulk_create(mappings_to_create, ignore_conflicts=True)
+        tenant_mappings.update({mapping.tenant_id: mapping for mapping in mappings})
+        bootstrapped_tenants = []
+
+        # Get existing custom groups
+        default_group_uuids = [mapping.default_group_uuid for mapping in tenant_mappings.values()]
+        existing_custom_groups = set(Group.objects.filter(uuid__in=default_group_uuids).values_list("uuid", flat=True))
```
lpichler, line 159: What about custom default groups which already exist but they don't have created tenant mapping, should we create tenant mapping with their uuid ?

rbac/management/management/commands/utils.py
```diff
@@ -0,0 +1,99 @@
...
+def get_s3_client():
+    """Create and return an S3 client."""
+    return boto3.client(
+        "s3",
+        aws_access_key_id=os.environ.get("S3_AWS_ACCESS_KEY_ID"),
+        aws_secret_access_key=os.environ.get("S3_AWS_SECRET_ACCESS_KEY"),
+    )
+
+
+def download_tenant_user_data(env):
```
alechenninger, line 43:
```suggestion
def download_tenant_user_data():
```

rbac/management/management/commands/utils.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def process_batch(batch_data):
+    """Process a batch of tenant and principal data."""
+    users = []
+    for org_id, org_admin, principal_name, user_id in batch_data:
+        user = User()
+        user.org_id = org_id
+        user.admin = org_admin
+        user.usernanme = principal_name
```
alechenninger, line 95:
```suggestion
        user.username = principal_name
```

rbac/management/tenant_service/v2.py
```diff
@@ -103,6 +104,94 @@ def update_user(
...
+            )
+        )
+        return bootstrapped_tenants
+
+    def _get_or_bootstrap_tenants(self, org_ids: list) -> list[BootstrappedTenant]:
+        """Bootstrap list of tenants, used by update_users."""
+        # Fetch existing tenants
+        existing_tenants = {
+            tenant.org_id: tenant
+            for tenant in Tenant.objects.filter(org_id__in=org_ids).select_related("tenantmapping")
```
alechenninger, line 181: Do you think we should set `related_name` on the `tenant` field of `TenantMapping` model to `tenant_mapping` to make this fit in with other related names?

rbac/management/tenant_service/v2.py
```diff
@@ -103,6 +104,94 @@ def update_user(
...
+                ),
+            ]
+
+            # Include platform for tenant
+            relationships.append(
+                create_relationship(("rbac", "tenant"), tenant_id, ("rbac", "platform"), settings.ENV_NAME, "platform")
+            )
+
+        # Ignore those already created
+        Workspace.objects.bulk_create(workspaces, ignore_conflicts=True)
```
alechenninger, line 151: Are you ignoring conflicts so that we can re-run the import? In that case maybe we should actually use `update_conflicts`.

rbac/management/tenant_service/v2.py
alechenninger: Good call – this needs to be fixed in the single update path, too. Make sure we add a test for this case.

rbac/management/management/commands/utils.py
astrozzc: Good catch!

rbac/management/tenant_service/v2.py
astrozzc: Yea, setting it looks better. I will add a migration for it

rbac/management/tenant_service/v2.py
astrozzc: Yea, I was avoiding conflicts of the created workspaces for the bootstrapped tenants. Now I refactored the code a bit to check the existing ones to use their uuid

rbac/management/management/commands/import_tenant_user_data.py
```diff
@@ -0,0 +1,32 @@
+"""Seeds command."""
```
alechenninger, line 1:
```suggestion
"""Import user data command."""
```

alechenninger, line 1:
```suggestion
"""Import user data command."""
```

rbac/management/tenant_service/v2.py
alechenninger: Oh, this changes the column name? Bummer but I guess okay.

rbac/management/tenant_service/v2.py
```diff
@@ -114,12 +203,13 @@ def update_users(self, users: list[User]):
         Args:
             users (list): List of User objects to update
         """
-        bootstrapped_list = []
+        org_ids = []
         for user in users:
             if user.org_id is None:
                 logger.warning(f"Cannot update user without org_id. Skipping. username={user.username}")
                 continue
```
alechenninger, line 210: I guess this assumes all users are enabled. We should probably just add a guard around this similar to the org_id check. e.g. 

```suggestion
            if not user.is_active:
                continue
            if user.org_id is None:
                logger.warning(f"Cannot update user without org_id. Skipping. username={user.username}")
                continue
```

(not sure if worth warning since there's no issue really if the user is not active)

alechenninger: Although, we'd also need to check below (loop at line 227)

tests/management/tenant/test_model.py
```diff
@@ -243,7 +243,7 @@ def test_bulk_adding_updating_users(self):
 
         self.service.update_users(users)
 
-        self.assertEquals(12, self.tuples.count_tuples())
+        self.assertEquals(18, self.tuples.count_tuples())
```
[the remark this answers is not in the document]
astrozzc: Thanks for pointing that out. I was bootstrapping all tenants even if it has been bootstrapped before. After thinking about it, this is redundant, I refactored the code a bit so it would skip those already bootstrapped.

rbac/management/tenant_service/v2.py
astrozzc: Yea, the column name. Right, just a small migration👍

rbac/management/tenant_service/v2.py
astrozzc: Yea, assuming all users enabled. Added the guard

rbac/management/tenant_service/v2.py
```diff
@@ -103,6 +105,125 @@ def update_user(
...
+                partition_key="rbactodo",
+                add=relationships,
+            )
+        )
+        return bootstrapped_tenants
+
+    def _get_or_bootstrap_tenants(self, org_ids: set) -> list[BootstrappedTenant]:
+        """Bootstrap list of tenants, used by update_users."""
+        # Fetch existing tenants
+        existing_tenants = {
```
lpichler, line 179: can this be simplified like this ?
```
existing_tenants = {
    tenant.org_id: tenant
    for tenant in Tenant.objects.filter(org_id__in=org_ids)
    .select_related("tenant_mapping")
    .prefetch_related(
        Prefetch(
            "group_set",
            queryset=Group.objects.filter(platform_default=True),
            to_attr="platform_default_groups",
        ),
        Prefetch(
            "workspace_set",
            queryset=Workspace.objects.filter(type__in=[Workspace.Types.ROOT, Workspace.Types.DEFAULT]),
            to_attr="workspaces",
        )
    )
}
```

lpichler (approved):
LGTM!

rbac/management/tenant_service/v2.py
astrozzc: Oh, I missed this one. Thanks! I update the logic to consider not bootstrapped ones by the tenant mapping, so the workspaces are not used anymore. Removed

rbac/management/tenant_service/v2.py
```diff
@@ -294,13 +413,42 @@ def _bootstrap_tenant(self, tenant: Tenant) -> BootstrappedTenant:
...
-        self, tenant: Tenant, mapping: TenantMapping, default_workspace: Workspace
+        self, tenant: Tenant, mapping: TenantMapping, default_workspace_uuid: str, custom_groups=None
     ) -> List[Relationship]:
         """
         Bootstrap default access for a tenant's users and admins.
 
         Creates role bindings between the tenant's default workspace, default groups, and system policies.
         """
+
+        def _create_default_relation_tupples(
```
alechenninger, line 424:
```suggestion
        def _create_default_relation_tuples(
```

This one could maybe also be moved to a top-level method so the _bootstrap_default_access is a little less busy? It's defined as an inner method but doesn't look like utilizes variables in that method's scope for what it's worth (unless I'm missing something).

astrozzc: Moved to top level

tests/management/tenant/test_model.py
```diff
@@ -386,4 +449,17 @@ def test_bulk_adding_updating_users(self):
                     subject("rbac", "role", admin_default_policy.uuid),
                 )
             ),
+            f"Expected admin default role binding to have admin default role for tenant {org_id}",
         )
+        return tenant, mapping, root, default
+
+    @patch("management.management.commands.utils.BOOT_STRAP_SERVICE")
+    def test_retrying_bulk(self, mock_bss):
```
alechenninger, line 457: Maybe this should be moved to a test class for the import_tenant_user_data module specifically? We should probably add some other tests specifically at that layer too. At least just a sanity check.

astrozzc: Updated

alechenninger (approved):
🤘🏻🤘🏻🤘🏻🤘🏻🤘🏻

PR #1230, opened 2024-10-08
RHCLOUD-34511: Add users from middleware and cleaner job to default groups, and bootstrap Tenants as needed

alechenninger:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-34511

See Jira. Given a user, this:

- Bootstraps a Tenant for the user. Bootstrapping is the process of ensuring a Tenant exists, has its built-in Workspaces, and keys for its default groups' and default role bindings' relations.
- Adds the user_id to the Principal if one exists
- Adds the user by {domain}:{user_id} to default groups, based on whether the user is an admin.

A User can be introduced in two places:

- via user STOMP queue
- via middleware authentication

Each of these trigger this logic.

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


rbac/management/principal/cleaner.py
```diff
@@ -123,98 +121,65 @@ def clean_tenants_principals():
...
             if ref["@entity-name"] == "Customer":
-                org_id = ref["#text"]
+                user.org_id = ref["#text"]
                 break
-        return {"user_id": user_id, "is_org_admin": is_org_admin, "username": user_name, "org_id": org_id}, True
-    return bop_resp["data"][0], False
-
+            if ref["@entity-name"] == "Account":
+                user.account = ref["#text"]
+                break
```
alechenninger, line 150: Should also check the entity-type or whatever it is for these fields

rbac/management/tenant/model.py
```diff
@@ -0,0 +1,470 @@
...
+            logger.warning("No admin default role found for public tenant. Default access will not be set up.")
+
+        default_workspace_uuid = str(default_workspace.uuid)
+        default_user_role_binding_uuid = str(mapping.default_user_role_binding_uuid)
+        default_admin_role_binding_uuid = str(mapping.default_admin_role_binding_uuid)
+
+        tuples_to_add: List[Relationship] = []
+
+        # Add default role binding IFF there is no custom default access for the tenant
+        if platform_default_role and not Group.objects.filter(platform_default=True, tenant=tenant).exists():
```
alechenninger, line 364: TODO: Think about race conditions here

astrozzc (approved):
Overall LGTM!

rbac/rbac/settings.py
```diff
@@ -353,6 +353,8 @@
 V2_MIGRATION_RESOURCE_APP_EXCLUDE_LIST = ENVIRONMENT.get_value(
     "V2_MIGRATION_RESOURCE_APP_EXCLUDE_LIST", default=""
 ).split(",")
+V2_BOOTSTRAP_TENANT = ENVIRONMENT.bool("V2_BOOTSTRAP_TENANT", default=False)
```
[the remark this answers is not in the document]
alechenninger: Yeah I'll add it in this one, good call out

rbac/management/group/definer.py
```diff
@@ -78,29 +79,39 @@ def seed_group():
...
         group_flag_change_notification_handler(user, group)
     return group
 
 
-def clone_default_group_in_public_schema(group, tenant):
+def clone_default_group_in_public_schema(group, tenant) -> Optional[Group]:
     """Clone the default group for a tenant into the public schema."""
+    if settings.PRINCIPAL_CLEANUP_UPDATE_ENABLED_UMB:
+        # TODO: bootstrap the tenant to get the mapping
+        # use this for uuid instead and to remove the default role binding tuple
```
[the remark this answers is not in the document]
alechenninger: This should be fixed with [RHCLOUD-35647](https://issues.redhat.com/browse/RHCLOUD-35647)

rbac/management/group/definer.py
```diff
@@ -78,29 +79,39 @@ def seed_group():
...
+    else:
+        uuid = uuid4()
+
     public_tenant = Tenant.objects.get(tenant_name="public")
     tenant_default_policy = group.policies.get(system=True)
     group.name = "Custom default access"
     group.system = False
     group.tenant = tenant
-    group.uuid = uuid4()
+    group.uuid = uuid
```
coderbydesign, line 107: Looks like this probably isn't an issue currently, but wondering if we might want to name it something more explicit to not get in a weird situation where we're stepping on the `uuid` module?

rbac/management/group/model.py
```diff
@@ -49,6 +52,29 @@ class Group(TenantAwareModel):
     system = models.BooleanField(default=False)
     admin_default = models.BooleanField(default=False)
 
+    @staticmethod
+    def relationship_to_principal_for_group(
```
coderbydesign, line 56: Not something to do now, because I think this ship has sailed anyway, but I wonder if we should have a service object layer that handles this type of thing for the models. Instead of conflating the data model logic with relationship construction, if we had a service object/class to do some of that to interface with the migrator, it might make things more clear, keep good separation, and maybe even make testing easier? Again, not necessarily in the scope of this PR but it's a thought I've had in terms of how we're organizing things.

rbac/management/tenant/model.py
```diff
@@ -0,0 +1,526 @@
+#
```
coderbydesign, line 1: Not sold one way or the other, but it was confusing for me when I saw `/management/tenant/model.py` knowing we have `/api/models.py` with a `Tenant` model/class. Trying to think about whether:

- we rename the path here to reflect the model name, so `/management/tenant_mapping/model.py`
- we put this in `/api/models.py` or its own file, alongside `Tenant`

Most of the `management` app models we keep as first-class objects we have APIs for, where the `api` app is more cross-cutting, but I'm curious folks' thoughts here.

rbac/management/group/definer.py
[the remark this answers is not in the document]
alechenninger: cc @lpichler ^

rbac/management/group/definer.py
alechenninger: yeah great point, simple fix

rbac/management/group/model.py
alechenninger: Yeah I could definitely see that. We should talk about how we want to organize logic going forward with the V2 API – it's probably a good chance to get a fresh take and establish a repeatable pattern for some of this stuff. I'm open to moving this somewhere else in the meantime too (top level methods?). Here was at least "close at hand" for the few contexts that needed to deal with group updates currently.

coderbydesign: Yeah, agreed. As long as we're confident in our tests, we should be able to refactor. 😄

rbac/migration_tool/migrate.py
```diff
@@ -87,6 +89,7 @@ def migrate_role(
     return relationships, v2_role_bindings
 
 
+# TODO: remove this
```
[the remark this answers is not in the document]
alechenninger: I can remove these here no worries

CJI should use the V2TenantBootstrapService directly I think

rbac/management/tenant/model.py
alechenninger: Yeah I wasn't sure where to put this / what to name it. It's definitely related to Tenants but it pulls in a bunch of other concepts which are management/rbac concepts, not core api concepts (e.g. group, role). Adding this stuff there would create a circular dependency (management depends on api but api depends on management).

Renaming to "tenant_mapping" is probably better than "tenant", so if nothing else that seems like a good idea.

Another option would be to do that and move the bootstrap service stuff to its own module?

coderbydesign: Yep, exactly what I was thinking too with the overlap with groups/roles, etc., so not making total sense to move to the `Tenant`. I like the idea of renaming and potentially breaking out some of the bootstrap logic.

rbac/management/tenant_service/v1.py
```diff
@@ -0,0 +1,70 @@
...
+                groups.append(group)
+                # We have to do the removal explicitly in order to clear the cache,
+                # or the console will still show the cached number of members
+                group.principals.remove(principal)
+            principal.delete()
+            if not groups:
+                logger.info(f"Principal {user.user_id} was not under any groups.")
+            for group in groups:
+                logger.info(f"Principal {user.user_id} was in group with uuid: {group.uuid}")
+        except (Tenant.DoesNotExist, Principal.DoesNotExist):
```
alechenninger, line 69: add test for this if not already present

rbac/management/tenant/model.py
```diff
@@ -0,0 +1,327 @@
...
+from migration_tool.utils import create_relationship
+
+from api.models import Tenant, User
+
+
+class TenantMapping(models.Model):
+    """Tenant mappings to V2 domain concepts."""
+
+    tenant = models.OneToOneField(Tenant, on_delete=models.CASCADE)
+    default_group_uuid = models.UUIDField(default=uuid.uuid4, editable=False, null=False)
```
lpichler: Do we need to new uuids for default_group and default_admin_group ? Is it possible to reuse existing ones from RBAC db  for ?

rbac/management/tenant_service/v2.py
```diff
@@ -0,0 +1,392 @@
...
+
+        # NOTE: This logic is prone to write skew: the platform group for the tenant may be created concurrently.
+        # Care must be taken to prevent this.
+        # Currently, when this default group is created (in group/definer.py) we:
+        # 1. Check for the existence of a tenant mapping. If exists, use that.
+        #    No race because if it already exists, this process must've already happened.
+        # 2. If tenant mapping does not exist, create it via this same bootstrap process.
+        #    Due to unique constraint, if this happens concurrently from another input (e.g. user import),
+        #    one will rollback, serializing the group creation with user import on next retry.
+        if platform_default_role and not Group.objects.filter(platform_default=True, tenant=tenant).exists():
```
lpichler, line 308: Isn't `platform_default_role` always set ? (same for admin_default_role )

rbac/management/tenant_service/v2.py
```diff
@@ -0,0 +1,392 @@
...
+
+    def _bootstrap_default_access(
+        self, tenant: Tenant, mapping: TenantMapping, default_workspace: Workspace
+    ) -> List[Relationship]:
+        """
+        Bootstrap default access for a tenant's users and admins.
+
+        Creates role bindings between the tenant's default workspace, default groups, and system policies.
+        """
+        platform_default_role = self._get_platform_default_policy_uuid()
```
lpichler, line 283: Is intent here to return policy uuid or role uuid ? Could be name of variable be improved ?
for example `platform_default_[role|group]_uuid`
same for `admin_default_role`

lpichler:
@alechenninger Will it be hard to split this PR: one for v1 and second for v2 ? I guess it will improve review process - if not I understand it and I fine with it as well.

For confirmation does this PR address , those tickets from [RHCLOUD-34511 ](https://issues.redhat.com/browse/RHCLOUD-34511) ? :
https://issues.redhat.com/browse/RHCLOUD-34858
https://issues.redhat.com/browse/RHCLOUD-35205
https://issues.redhat.com/browse/RHCLOUD-35578
https://issues.redhat.com/browse/RHCLOUD-35639

rbac/management/tenant_mapping/model.py
```diff
@@ -0,0 +1,37 @@
...
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+
+class TenantMapping(models.Model):
+    """Tenant mappings to V2 domain concepts."""
+
+    tenant = models.OneToOneField(Tenant, on_delete=models.CASCADE)
+    default_group_uuid = models.UUIDField(default=uuid.uuid4, editable=False, null=False)
+    default_admin_group_uuid = models.UUIDField(default=uuid.uuid4, editable=False, null=False)
+    default_user_role_binding_uuid = models.UUIDField(default=uuid.uuid4, editable=False, null=False)
```
lpichler, line 36: How "user" is meant here ? Is it like non-admin ?  If so maybe it could be `default_role_binding_uuid`, as there is in RBAC default_group vs default_admin_group...

coderbydesign:
I think we'll want to update the base branch on this PR now that `dual_write_for_role` has been merged?

rbac/management/tenant_service/v2.py
alechenninger: It's not always set in tests unfortunately. So rather than update a bunch of tests, I added this warning. I can see how many tests it is maybe it's not too bad to update them.

rbac/management/tenant/model.py
alechenninger: We can't because these do not correspond with RBAC groups initially. The RBAC default groups are shared across the entire platform. These need to be per-tenant. So, we need to have a separate UUID for these, and then reuse the same UUID for custom default group so that the group membership lines up for custom default group rules.

rbac/management/tenant_mapping/model.py
alechenninger: Good call – fixed

alechenninger:
> For confirmation does this PR address , those tickets from [RHCLOUD-34511 ](https://issues.redhat.com/browse/RHCLOUD-34511) ? :
> https://issues.redhat.com/browse/RHCLOUD-34858
> https://issues.redhat.com/browse/RHCLOUD-35205
> https://issues.redhat.com/browse/RHCLOUD-35578
> https://issues.redhat.com/browse/RHCLOUD-35639

Yes it does!

> @alechenninger Will it be hard to split this PR: one for v1 and second for v2 ? I guess it will improve review process - if not I understand it and I fine with it as well.

In hindsight, this would've been better, but if we can save time with the current branch that would be appreciated.

lpichler (approved):

PR #1243, opened 2024-10-16
RHCLOUD-35742: Seeded default group-roles attempt replication in add_role but these shouldn't

alechenninger:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-35742

See Jira

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


rbac/management/group/definer.py
```diff
@@ -167,8 +167,9 @@ def add_roles(group, roles_or_role_ids, tenant, user=None):
 
         system_policy.roles.add(role)
 
-        dual_write_handler = RelationApiDualWriteGroupHandler(group, ReplicationEventType.ASSIGN_ROLE, [])
-        dual_write_handler.replicate_added_role(role)
+        if group.tenant.tenant_name != "public":
```
astrozzc, line 170: Oh, I just realize that we can make use of the system flag in this case. The system True group are all in the public tenant, then we can save a query to get the tenant

rbac/management/group/definer.py
```diff
@@ -192,8 +193,9 @@ def remove_roles(group, roles_or_role_ids, tenant, user=None):
                 policy.roles.remove(role)
                 logger.info(f"Removing role {role} from group {group.name} for tenant {tenant.org_id}.")
 
-                dual_write_handler = RelationApiDualWriteGroupHandler(group, ReplicationEventType.UNASSIGN_ROLE, [])
-                dual_write_handler.replicate_removed_role(role)
+                if group.tenant.tenant_name != "public":
```
lpichler, line 196: this check is inside  `replicate_removed_role`

rbac/management/group/definer.py
alechenninger: Sometimes tests use system groups outside the public tenant so I'm worried relying on that implicit rule is a little fragile. We want to check for the public tenant specifically, and this does that directly.

That said, both methods are already passed the tenant so we can just use that. Thanks for pointing out the extra query.

rbac/management/group/definer.py
alechenninger: Yes. It needs to be because the public tenant doesn't have a default workspace. Is there an issue?

rbac/management/group/definer.py
alechenninger: Hrm, although I'm not sure why the method queries for the group when it's already given one. But, changing that is probably more invasive than we need right now. Because it does that though (the extra query above with the `tenant`), it's looking like using the `tenant` variable is viable at least. WDYT?

astrozzc: 👍 Ah ha, the tenant is already there, lol, nice!

astrozzc (approved):

lpichler (approved):

rbac/management/group/definer.py
alechenninger: Hrm yeah the issue is this part in the constructor:

```
            self.tenant = group.tenant
            self.default_workspace = Workspace.objects.get(tenant=self.tenant, type=Workspace.Types.DEFAULT)
```

Thanks for explaining your concern. I think you're right that there is redundancy. The redundancy is probably more inside `replicate_removed_role` and other methods that check the group's tenant. I'll remove those checks and add a check inside the constructor since it wouldn't work in those cases as the design evolved.

lpichler: oh I see, thanks for explaining and I definitely agree 👍

PR #1245, opened 2024-10-17
[RHCLOUD-35640] Update custom default group logic 

lpichler:
This PR adds replication
-  for creation of custom default group logic:
   - add roles to default group 
   - remove roles from default group 
- removal of custom groups

## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-35640

The what, why and how.

## Local Testing
- add roles to default group:
```
POST http://localhost:8000/api/rbac/v1/groups/<detault_group_uuid>/roles/

{"roles":["d414435a-00fc-4dd5-82dd-c5c345ad1338"]}

```

- remove roles from default group 
```
DELETE http://localhost:8000/api/rbac/v1/groups/<custom_detault_group_uuid>/roles/?roles=d414435a-00fc-4dd5-82dd-c5c345ad1338
```
- removal of custom groups
```
DELETE http://localhost:8000/api/rbac/v1/groups/<custom_detault_group_uuid>/
```

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


rbac/management/group/definer.py
```diff
@@ -92,10 +93,14 @@ def set_system_flag_before_update(group: Group, tenant, user) -> Optional[Group]
...
 def clone_default_group_in_public_schema(group, tenant) -> Optional[Group]:
     """Clone the default group for a tenant into the public schema."""
-    if settings.PRINCIPAL_CLEANUP_UPDATE_ENABLED_UMB:
-        # TODO: bootstrap the tenant to get the mapping
-        # use this for uuid instead and to remove the default role binding tuple
-        group_uuid = uuid4()
+    if settings.PRINCIPAL_CLEANUP_UPDATE_ENABLED_UMB and settings.REPLICATION_TO_RELATION_ENABLED:
+        tenant_bootstrap_service = V2TenantBootstrapService(OutboxReplicator())
+        bootstrapped_tenant = tenant_bootstrap_service.bootstrap_tenant(tenant)
+        tenant_bootstrap_service.remove_and_replicate_default_bindings(bootstrapped_tenant, group.admin_default)
```
lpichler, line 99: @alechenninger or should this be rather consolidated in one replication event ?

rbac/management/group/definer.py
```diff
@@ -92,10 +93,14 @@ def set_system_flag_before_update(group: Group, tenant, user) -> Optional[Group]
 
 def clone_default_group_in_public_schema(group, tenant) -> Optional[Group]:
     """Clone the default group for a tenant into the public schema."""
-    if settings.PRINCIPAL_CLEANUP_UPDATE_ENABLED_UMB:
-        # TODO: bootstrap the tenant to get the mapping
-        # use this for uuid instead and to remove the default role binding tuple
-        group_uuid = uuid4()
+    if settings.PRINCIPAL_CLEANUP_UPDATE_ENABLED_UMB and settings.REPLICATION_TO_RELATION_ENABLED:
+        tenant_bootstrap_service = V2TenantBootstrapService(OutboxReplicator())
+        bootstrapped_tenant = tenant_bootstrap_service.bootstrap_tenant(tenant)
```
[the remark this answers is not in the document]
astrozzc: If after my CJI job run, I think yes. But now it is not guaranteed a tenant has those workspaces created.
However, it is ok either the tenant has or not. It is guaranteed to coexist with the tenant mapping. If the tenant mapping exist, so does the default/root ws. If not, we create them

rbac/management/group/definer.py
```diff
@@ -82,20 +84,29 @@ def seed_group() -> Tuple[Group, Group]:
...
+    relationships = []
+    if settings.REPLICATION_TO_RELATION_ENABLED:
+        tenant_bootstrap_service = V2TenantBootstrapService(OutboxReplicator())
+        bootstrapped_tenant = tenant_bootstrap_service.bootstrap_tenant(tenant)
+
+        relationships = tenant_bootstrap_service.default_bindings_from_mapping(
+            bootstrapped_tenant, group.admin_default
+        )
+        if group.admin_default:
+            group_uuid = bootstrapped_tenant.mapping.default_admin_group_uuid
```
astrozzc, line 107: We don't allow admin_default group to be customized. So this clone method won't apply to admin_default. We can simplify the logic to treat it as False

rbac/management/relation_replicator/relation_replicator.py
```diff
@@ -48,6 +48,7 @@ class ReplicationEventType(str, Enum):
     EXTERNAL_USER_UPDATE = "external_user_update"
     MIGRATE_CUSTOM_ROLE = "migrate_custom_role"
     MIGRATE_TENANT_GROUPS = "migrate_tenant_groups"
+    REMOVE_DEFAULT_BINDINGS = "remove_default_bindings"
```
astrozzc, line 51: Seems this is not used anywhere

rbac/management/group/definer.py
```diff
@@ -168,6 +178,8 @@ def add_roles(group, roles_or_role_ids, tenant, user=None):
         system_policy.roles.add(role)
 
         dual_write_handler = RelationApiDualWriteGroupHandler(group, ReplicationEventType.ASSIGN_ROLE, [])
+        if relations is not None:
+            dual_write_handler.extend_relations_to_remove(relations)
```
astrozzc, line 182: Seems the relations to remove is not complete?
We have to 
1. remove the the default_bindings_from_mapping generated for the system platform_default group;
2. add the 
default_bindings_from_mapping generated for the custom platform_default group;
3. remove all the users to the system platform_default
4. add all the users to the custom platform_default

rbac/management/relation_replicator/relation_replicator.py
lpichler: yes it is now unnecessary as I consolidated replication here, thanks!

rbac/management/group/definer.py
lpichler: @astrozzc Step is there but I am not sure about step 2-4.

My understanding was that all those bindings are using some default custom uuid from mapping.

astrozzc: Thanks for Libor's walking through offline, I have some misunderstanding of some steps. So we don't need the 3 and 4

rbac/management/group/definer.py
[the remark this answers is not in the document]
alechenninger: Right, by definition a tenant is not considered "bootstrapped" (in v2) if they don't have built in workspaces. That's part of the process. So `bootstrap_tenant` must create the builtin workspaces as well as mapping.

>         # This assumes that Tenant Mapping exist

Not sure what is meant by the comment exactly – but my interpretation of it is that it is maybe misleading. The method below _does not_ assume the TenantMapping exists. It creates it if it doesn't.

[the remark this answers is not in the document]
lpichler: This method is called only from API endpoint and in that situation TenantMapping has to be created already. 
But maybe in some edge cases it is not true ? 

I can remove the comment. My main concern was whether the tenant can be created, but the default workspace is not, or if the tenant mapping could be created, but the default workspace is not. Because if the tenant mapping is not created and the default workspace is already created, this method will fail - but it looks is not possible.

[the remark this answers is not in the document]
alechenninger: > This method is called only from API endpoint and in that situation TenantMapping has to be created already. But maybe in some edge cases it is not true ?

Okay, yeah I see the issue. We need to update the settings check above to check for V2_BOOTSTRAP_TENANT I think. I think this would only happen during migration steps as of latest doc. There is a window where a default group may attempt to be created for an existing tenant that is not yet bootstrapped.

We could keep this as is, and then I think it should never happen to your point. The user import job will take care of race conditions in that case. But, I think it is a little safer to bootstrap the tenant in that case, to ensure we use the uuid from the tenant mapping. And we know we cannot get into trouble there, because of the 1-1 unique constraint on the tenant mapping, any concurrent tx creating one (e.g. the user import) will fail.

> My main concern was whether the tenant can be created, but the default workspace is not, or if the tenant mapping could be created, but the default workspace is not. Because if the tenant mapping is not created and the default workspace is already created, this method will fail - but it looks that it is not possible.

Yes, correct – we should never have workspaces without the TenantMapping. That's why the bootstrap service stuff was created – to consolidate the logic to enforce those kinds of things.

rbac/management/group/definer.py
```diff
@@ -112,22 +118,22 @@ def clone_default_group_in_public_schema(group, tenant) -> Optional[Group]:
...
     tenant_default_policy.group = group
     tenant_default_policy.save()
     tenant_default_policy.roles.set(public_default_roles)
-    return group
+    return group, relationships
 
 
 @transaction.atomic
-def add_roles(group, roles_or_role_ids, tenant, user=None):
+def add_roles(group, roles_or_role_ids, tenant, user=None, relations=None):
```
alechenninger, line 132: Main thing I am wondering about is if we can avoid passing `relations` here all the way through.

It is a bit awkward right now. At the very least I would rename the parameter. But maybe we can have this return relations, and then have an outer method consolidate them into a single replication event. But, will take a look more tomorrow.

Sorry for the short comment – need to run now and just wanted to jot this down for you to think on :-). Catch up more tomorrow :-).

lpichler: > Main thing I am wondering about is if we can avoid passing relations here all the way through.

Yes definitely I agree! I wanted to know whether changes are ok regard to algorithm and relation generation. Anyway I will rework to avoid passing `relations`.

alechenninger: So as I was looking at this I think I have a suggestion for this, which I think would also make the logic easier to understand/validate.

First, it involves using two outbox events. One event will separate the default access, and replicate that default access with new role bindings. The other will update those role bindings to add or remove roles.

I do not think it is worth the complexity to combine those outbox events. Using two separate events does two things: (1) the additional access or removed access will be "more" eventually consistent with RBAC, by one additional event, and (2) it adds potentially redundant work. As for (1), it is already eventually consistent. Because both events are committed in one transaction, they will still be correctly ordered with respect to concurrent changes (e.g. adding/removing users from the group which is itself eventually consistent for default groups anyway). So, it actually doesn't change the semantics in any way to combine those events. To any perspective, the result is the same: access is the default, and then it changes at some point in the future, regardless of whether that happens with one replication event or two.

As for (2) (performance)... I'm just thinking it's premature optimization to think about that much otherwise. The worst case scenario is that you remove all access that was just added. So you'd start by adding all role bindings, and then you'd remove them. Realistically, if you had default access for 30 roles, that means adding ~90 tuples and then removing them. In the grand scheme of things, probably not a big deal. And if it's a problem, we can always improve it later.

So given that, here is what I would suggest:

1. Change the `RelationApiDualWriteGroupHandler` but not _quite_ as much as in this PR currently. Maybe just keep the separate replicate piece or let it take multiple roles at once. So change `replicate_added_role` to accept multiple roles, and then we just add them all together in one method call. Or separate `add_role` from `replicate_added_role` and then keep the separate `replicate()` method. (but we don't need the custom group param or other stuff)
2. In `clone_default_group_in_public_schema`, replicate an event that (a) removes default access and (b) re-adds the roles. I think this should be able to reuse the old `RelationApiDualWriteGroupHandler` code and call `replicate_added_roles` with all the roles (or whatever based on above), similar to what `add_roles` used to do. I would probably add an argument or method to the group handler for removing default access, and then remove the `default_bindings_from_mapping` method from the bootstrap service. The logic there doesn't really have to be kept in the bootstrap service for a few reasons: 
  1. The group handler already knows that default access goes to the default workspace and already looks up the default workspace.
  2. It already has the group uuid and mapping for the default role binding uuid
  3. The only missing piece then is the platform default role uuid. We could consider just not removing this part since strictly speaking all we *have* to remove is the binding relationship. Regardless, the platform default role uuid is already repeated by the seeding process, so its something that should really be exposed outside of the bootstrap service in a reusable place, anyway.
3. Then, adjust `add_roles` and `remove_roles` only to use the modified methods per 1 above. (It would be an unrelated improvement to have a single outbox event for those–I hadn't realized or forgotten that was doing multiple I think.)

One easy way to improve the performance, too, would be to mint a `ReplicationEvent` and optionally allow passing _that_ into the add/remove roles methods. Then, if present, "merge" it with the one generated by _that_ method. This would decouple it from the logic and keep the "redundant" work purely in memory confined to a single CPU.

So in summary that would look something like this (psuedo-ish code):

```python
def clone_default_group_in_public_schema(group, tenant) -> Optional[Group]:
    """Clone the default group for a tenant into the public schema."""
    bootstrapped_tenant = None
    if settings.V2_TENANT_BOOTSTRAP:
        tenant_bootstrap_service = V2TenantBootstrapService(OutboxReplicator())
        bootstrapped_tenant = tenant_bootstrap_service.bootstrap_tenant(tenant)
        group_uuid = bootstrapped_tenant.mapping.default_group_uuid
    else:
        group_uuid = uuid4()

    # ... all the same group and policy set up stuff
    dual_write_handler = RelationApiDualWriteGroupHandler(group, ReplicationEventType.CUSTOMIZE_DEFAULT_GROUP)
    dual_write_handler.replicate_added_roles(role, remove_default_access_from=bootstrapped_tenant.mapping)
    return group
```

And then add/remove_roles are mostly untouched I think (aside from any updates to dual write handler interface).

WDYT?

rbac/management/group/relation_api_dual_write_group_handler.py
```diff
@@ -123,13 +127,29 @@ def _replicate(self):
...
+    def extend_relations_to_remove(self, relations_to_remove: list[Relationship]):
+        """Extend relations to remove in replication."""
+        if not self.replication_enabled():
+            return
+        self.group_relations_to_remove.extend(relations_to_remove)
+
+    def add_system_roles_to_default_custom_group(self, custom_group: Group):
+        """Add system roles to default custom group."""
+        roles = Role.objects.filter(policies__group=custom_group)
+        system_roles = roles.filter(tenant=Tenant.objects.get(tenant_name="public"))
```
alechenninger, line 139: Curious – do you need the additional filter on tenant?

rbac/management/group/relation_api_dual_write_group_handler.py
```diff
@@ -252,7 +279,10 @@ def prepare_to_delete_group(self):
             custom_ids.append(role.id)
 
         if self.group.platform_default:
-            pass  # TODO: create default bindings,
+            bootstrap_service = V2TenantBootstrapService(replicator=NoopReplicator())
+            bootstrapped_tenant = bootstrap_service.bootstrap_tenant(self.group.tenant)
+            relations_to_add = bootstrap_service.default_bindings_from_mapping(bootstrapped_tenant)
+            self.group_relations_to_add.extend(relations_to_add)
```
alechenninger, line 285: I think this logic is technically correct / working, but I wonder if this should really be used via the bootstrap service. I guess I kind of touched on this in the other comment, but while that service does rely on it, it's a little awkward to make that the main source, for something relatively simple. I'm curious what it would look like if we queried for the mapping for a tenant instead here and created the relationships to remove inline.

We can again consider only adding/removing the binding tuple, which would simplify this, if needed.

rbac/rbac/settings.py
```diff
@@ -353,6 +353,7 @@
 V2_MIGRATION_RESOURCE_EXCLUDE_LIST = ENVIRONMENT.get_value("V2_MIGRATION_RESOURCE_EXCLUDE_LIST", default="").split(",")
 V2_BOOTSTRAP_TENANT = ENVIRONMENT.bool("V2_BOOTSTRAP_TENANT", default=False)
 V1_BOOTSTRAP_ADD_USER_ID = ENVIRONMENT.bool("V1_BOOTSTRAP_ADD_USER_ID", default=False)
+LOG_REPLICATION = ENVIRONMENT.bool("LOG_REPLICATION", default=False)
```
alechenninger, line 356: We could consider an "aggregate replicator" which passed the event to multiple implementations, and then optionally using this with the logging replicator plus another impl, as an elegant way to add logging to any replicator, but I'm not sure if it's worth it. Or if it is, maybe this can be its own small PR.

rbac/management/group/definer.py
```diff
@@ -82,20 +84,23 @@ def seed_group() -> Tuple[Group, Group]:
...
 
-def clone_default_group_in_public_schema(group, tenant) -> Optional[Group]:
+def clone_default_group_in_public_schema(group, tenant) -> Tuple[Optional[Group], Iterable[Relationship]]:
     """Clone the default group for a tenant into the public schema."""
-    if settings.PRINCIPAL_CLEANUP_UPDATE_ENABLED_UMB:
-        # TODO: bootstrap the tenant to get the mapping
-        # use this for uuid instead and to remove the default role binding tuple
-        group_uuid = uuid4()
+    relationships = []
+    if settings.REPLICATION_TO_RELATION_ENABLED:
```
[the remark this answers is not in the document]
lpichler: Thanks!

rbac/management/group/relation_api_dual_write_group_handler.py
lpichler: Intent here is to fetch only system roles which are assigned to custom default group. Let me know If you think that it is not needed.

rbac/management/group/definer.py
lpichler: Two outbox events vs. one – it seems there are arguments for both approaches. From a performance standpoint, which is more efficient: processing one or two Kafka events (considering factors like serialization, deserialization, network, I/O, etc.)? Additionally, what it is preferable if one event in the sink fails: two separate events, or a single combined event? Also init of dual write handler twice.
Arguments which are  stated also valid. Personally I am not sure on 100 % what is optimal right now, so I choose your suggestion which (if I understand it well :) ) looks pretty simple and straightforward.

There are still some piece where clarification is needed:

> 1. Default binding tuples :
currently was described as tuples:

A: Workspace (default ws id) -> Role Binding ( from tenant mapping )
B: Role Binding (role binding from tenant mapping ) -> Role (Default policy UUID)
C: Role Binding (role binding from tenant mapping ) -> Default Group UUID (from tenant mapping)

you mentoined that not all of those are needed to add(create custom default role)/remove(delete custom default group) - is there the one which needs to be remove ?
I keept A and C - is also A necessary ?


> 2. One easy way to improve the performance, too, would be to mint a ReplicationEvent and optionally allow passing that into the add/remove roles methods. Then, if present, "merge" it with the one generated by that method. This would decouple it from the logic and keep the "redundant" work purely in memory confined to a single CPU.

This point is not clear to me -  what is meant by   
- "add/remove roles methods."  is it in dual write handler or in group/definer ?
- what is meant  "redundant" work here ?

I can see "redundant" work only in init dual write, otherwise now it calculates relations in memory and replicate two times as was suggested.

(Not sure if this is minor)

rbac/management/group/relation_api_dual_write_group_handler.py
lpichler: bootstrap service was decoupled from this place.
only question to what it is  binding tuple in https://github.com/RedHatInsights/insights-rbac/pull/1245#discussion_r1822379932

rbac/rbac/settings.py
lpichler: I removed but such simple attaching to outbox replication was definitely worth to local debugging without any other changes :)

rbac/management/group/relation_api_dual_write_group_handler.py
```diff
@@ -123,39 +130,65 @@ def _replicate(self):
...
-    def replicate_added_role(self, role: Role):
-        """Replicate added role."""
+    def extend_relations_to_remove(self, relations_to_remove: list[Relationship]):
+        """Extend relations to remove in replication."""
+        if not self.replication_enabled():
+            return
+        self.group_relations_to_remove.extend(relations_to_remove)
+
+    def replicate_added_roles(self, roles: list[Role], remove_default_access_from: Optional[TenantMapping] = None):
+        """Generate group relations and binding mapping for role."""
```
alechenninger, line 140: This removes a type warning (in the definer we pass the django manager which is not a list but is iterable)
```suggestion
    def replicate_added_roles(self, roles: Iterable[Role], remove_default_access_from: Optional[TenantMapping] = None):
```

rbac/management/group/relation_api_dual_write_group_handler.py
```diff
@@ -54,6 +56,8 @@ def __init__(
         """Initialize RelationApiDualWriteGroupHandler."""
         if not self.replication_enabled():
             return
+        if group.tenant.tenant_name == "public":
+            raise DualWriteException()
```
alechenninger, line 60: Does this result in another query? Is it worth it since the workspace query will fail regardless? FWIW there is a test case that would catch if a public tenant was ever allowed through some later refactoring.

rbac/management/group/relation_api_dual_write_group_handler.py
```diff
@@ -123,39 +130,65 @@ def _replicate(self):
...
-            role, update_mapping=add_group_to_binding, create_default_mapping_for_system_role=create_default_mapping
-        )
+        if remove_default_access_from is not None:
+            default_bindings = self._create_default_relation_tuples(
+                self.default_workspace.id,
+                remove_default_access_from.default_role_binding_uuid,
+                remove_default_access_from.default_group_uuid,
+            )
+            self.group_relations_to_remove.extend(default_bindings)
+
```
alechenninger, line 175: We can simplify this like this:

```suggestion
            default_binding = create_relationship(
                ("rbac", "workspace"),
                str(self.default_workspace.id),
                ("rbac", "role_binding"),
                str(remove_default_access_from.default_role_binding_uuid),
                "binding",
            )
            self.group_relations_to_remove.append(default_binding)
```

rbac/management/group/relation_api_dual_write_group_handler.py
```diff
@@ -123,39 +130,65 @@ def _replicate(self):
...
-        self._update_mapping_for_role(
-            role, update_mapping=add_group_to_binding, create_default_mapping_for_system_role=create_default_mapping
-        )
+        if remove_default_access_from is not None:
+            default_bindings = self._create_default_relation_tuples(
+                self.default_workspace.id,
+                remove_default_access_from.default_role_binding_uuid,
+                remove_default_access_from.default_group_uuid,
+            )
+            self.group_relations_to_remove.extend(default_bindings)
```
alechenninger, line 174: To clarify what I meant about simplifying tuple maintenance, this is what I meant:

```suggestion
            default_binding = self._default_binding(mapping=remove_default_access_from)
            self.group_relations_to_remove.extend(default_binding)
```

And below I have a suggestion for that function

rbac/management/group/relation_api_dual_write_group_handler.py
```diff
@@ -123,39 +130,65 @@ def _replicate(self):
...
+        if not self.replication_enabled():
+            return
+
+        if self.group.tenant.tenant_name == "public":
+            return
         self._replicate()
 
-    def replicate_removed_role(self, role: Role):
+    def replicate_removed_roles(self, roles: list[Role]):
         """Replicate removed role."""
```
alechenninger, line 186: Not strictly needed but if we change the other one, they would be consistent
```suggestion
    def replicate_removed_roles(self, roles: Iterable[Role]):
```

rbac/management/group/relation_api_dual_write_group_handler.py
```diff
@@ -123,39 +130,65 @@ def _replicate(self):
         except Exception as e:
             raise DualWriteException(e)
 
-    def replicate_added_role(self, role: Role):
-        """Replicate added role."""
+    def extend_relations_to_remove(self, relations_to_remove: list[Relationship]):
+        """Extend relations to remove in replication."""
+        if not self.replication_enabled():
+            return
+        self.group_relations_to_remove.extend(relations_to_remove)
```
alechenninger, line 137: I think we can remove this now
```suggestion
```

rbac/management/group/relation_api_dual_write_group_handler.py
```diff
@@ -252,11 +284,60 @@ def prepare_to_delete_group(self):
...
+                ("rbac", "role_binding"),
+                str(role_binding_uuid),
+                ("rbac", "group"),
+                str(default_group_uuid),
+                "subject",
+                "member",
+            ),
+        ]
+
     def replicate_deleted_group(self):
```
alechenninger, line 341: Maybe remove and replace calls with `replicate()` now that we have that?

rbac/management/group/relation_api_dual_write_group_handler.py
```diff
@@ -123,39 +130,65 @@ def _replicate(self):
...
+        """Replicate added role."""
+        if not self.replication_enabled():
+            return
+
+        if self.group.tenant.tenant_name == "public":
+            return
         self._replicate()
 
-    def replicate_removed_role(self, role: Role):
+    def replicate_removed_roles(self, roles: list[Role]):
```
alechenninger, line 185: Not strictly needed but if we change the other method it would be consistent
```suggestion
    def replicate_removed_roles(self, roles: Iterable[Role]):
```

rbac/management/group/relation_api_dual_write_group_handler.py
```diff
@@ -123,39 +130,65 @@ def _replicate(self):
...
 
-    def replicate_added_role(self, role: Role):
-        """Replicate added role."""
+    def extend_relations_to_remove(self, relations_to_remove: list[Relationship]):
+        """Extend relations to remove in replication."""
+        if not self.replication_enabled():
+            return
+        self.group_relations_to_remove.extend(relations_to_remove)
+
+    def replicate_added_roles(self, roles: list[Role], remove_default_access_from: Optional[TenantMapping] = None):
```
alechenninger, line 139: This removes a type warning (in the definer we pass the django manager which is not a list but is iterable)
```suggestion
    def replicate_added_roles(self, roles: Iterable[Role], remove_default_access_from: Optional[TenantMapping] = None):
```

rbac/management/group/definer.py
alechenninger: A single event is definitely more efficient but kinda as I alluded to I think the cost is too great here. In general I think we should be more careful about performance, but mainly around unnecessary queries and especially around the read path. This is write path, and only a very particular write (modifying a default group for this first time).

For the tuples, we can just deal with adding and removing A. I added suggestions for this on the PR. That way we don't need any extra lookups (mainly to reduce the amount of code and logic) since we already have the mapping and the default workspace ID.

> This point is not clear to me - what is meant by
>
> "add/remove roles methods." is it in dual write handler or in group/definer ?

In the definer, so in this alteration you'd pass along the event as an argument (e.g. instead of the relationships that we had before.)

> what is meant "redundant" work here ?

Removing tuples that we just added – this could be worked out in memory rather than doing all the work in spicedb postgres to add and then remove the same tuples in a subsequent message. If it's all one we can just send one deduplicated batch to the sink.

> I can see "redundant" work only in init dual write, otherwise now it calculates relations in memory and replicate two times as was suggested.

What's redundant is that, in the worst case, you could be modifying the default group to remove all roles. In this case we'd publish two events: one event to add all tuples, and one event to remove them all. Better just not to add them in the first place. But it's not a big deal.

>
> (Not sure if this is minor)

I think it is definitely minor so I wouldn't worry about it.

rbac/management/group/relation_api_dual_write_group_handler.py
alechenninger: Also, do you want this to replicate?

Two options I think...

Option 1: Change the name for this and remove (e.g. `add_roles` and `remove_roles`) and keep the separate `replicate()` method. In which case, also remove the `replicate_deleted_group()` method since it's redundant with `replicate()`. 

or

Option 2: Keep the names, but make the methods replicate, and remove the separate `replicate()` method. (`replicate_deleted_group()` stays in this case)

rbac/management/group/definer.py
```diff
@@ -166,12 +179,11 @@ def add_roles(group, roles_or_role_ids, tenant, user=None):
...
-
-        if tenant.tenant_name != "public":
-            dual_write_handler = RelationApiDualWriteGroupHandler(group, ReplicationEventType.ASSIGN_ROLE)
-            dual_write_handler.replicate_added_role(role)
-        # Send notifications
         group_role_change_notification_handler(user, group, role, "added")
+        if dual_write_handler:
+            dual_write_handler.replicate_added_roles([role])
+    if dual_write_handler:
+        dual_write_handler.replicate()
```
alechenninger, line 186: GitHub isn't letting me make the full suggestion, but with the method that supports a list, maybe we want to do this:

```
    new_roles: list[Role] = []

    for role in [*system_roles, *custom_roles]:
        # Only Organization administrators are allowed to add the role with RBAC permission
        # higher than "read" into a group.
        for access in role.access.all():
            if (
                access.permission_application() == "rbac"
                and access.permission.verb != "read"
                and user
                and not user.admin
            ):
                key = "add-roles"
                message = (
                    "Non org admin users are not allowed to add RBAC role with higher than 'read' permission "
                    "into groups."
                )

        # Only add the role if it was not attached
        if system_policy.roles.filter(pk=role.pk).exists():
            continue

        system_policy.roles.add(role)
        group_role_change_notification_handler(user, group, role, "added")
        new_roles.append(role)
    
    if group.tenant.tenant_name != "public":
        dual_write_handler = RelationApiDualWriteGroupHandler(group, ReplicationEventType.ASSIGN_ROLE)
        dual_write_handler.replicate_added_roles(new_roles)
        dual_write_handler.replicate()
```

See other comment wrt `replicate_added_roles` name

I think this should really be either:

```
        dual_write_handler.add_roles(new_roles)
        dual_write_handler.replicate()
```

or just

```
        dual_write_handler.replicate_added_roles(new_roles)
```

rbac/management/group/definer.py
```diff
@@ -186,19 +198,25 @@ def remove_roles(group, roles_or_role_ids, tenant, user=None):
...
-                    dual_write_handler.replicate_removed_role(role)
-
+                if dual_write_handler:
+                    dual_write_handler.replicate_removed_roles([role])
                 # Send notifications
                 group_role_change_notification_handler(user, group, role, "removed")
+                removed_roles.append(role)
+
+    if dual_write_handler:
+        dual_write_handler.replicate()
```
alechenninger, line 219: Similar to above, maybe like this:

```
    removed_roles = []

    for policy in group.policies.all():
        for role in [*system_roles, *custom_roles]:
            # Only remove the role if it was attached
            if policy.roles.filter(pk=role.pk).exists():
                policy.roles.remove(role)
                logger.info(f"Removing role {role} from group {group.name} for tenant {tenant.org_id}.")
                # Send notifications
                group_role_change_notification_handler(user, group, role, "removed")
                removed_roles.append(role)

    if group.tenant.tenant_name != "public":
        dual_write_handler = RelationApiDualWriteGroupHandler(group, ReplicationEventType.UNASSIGN_ROLE)
        dual_write_handler.replicate_removed_roles(removed_roles)
        dual_write_handler.replicate()
```

(and again either just `replicate_removed_roles`, or `remove_roles` then `replicate`)

tests/management/group/test_view.py
```diff
@@ -1003,6 +999,35 @@ def test_delete_custom_default_group(self):
...
         self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
 
+        actual_call_arg = mock_method.call_args[0][0]
+
+        to_add = actual_call_arg["relations_to_add"]
+
+        tenant_mapping = TenantMapping.objects.get(tenant=self.tenant)
+
+        def assert_group_tuples(tuple_to_replicate):
+            relation_tuple = relation_api_tuple(
```
alechenninger, line 1009: I wonder if we could reuse some of the existing concepts more in tests. I find these tests a little hard to follow and update personally but it might just be me. For example, instead of new methods to create tuples like `relation_api_tuple`, can we use the `Relationship` type already used, or `create_relationship` method, or use `InMemoryTuples` and the `RelationTuple` type?

Maybe we have too many ways of doing the same thing... 🤔. It makes running into new tests sorta like starting over to learn how "those" tests do it.

tests/management/group/test_view.py
```diff
@@ -1768,6 +1795,71 @@ def test_system_flag_update_on_add(self, send_kafka_message):
...
+            to_add = actual_call_arg["relations_to_add"]
+
+            binding_mapping = BindingMapping.objects.get(
+                role=default_role, resource_type_name="workspace", resource_id=self.default_workspace.id
+            )
+
+            # New platform default role for tenant created
+            custom_default_group = Group.objects.get(platform_default=True, tenant=self.tenant)
+
+            def assert_group_tuples(tuple_to_replicate):
```
alechenninger, line 1809: Is this reused? Why define a method and then call it? Can we do the asserts directly?

tests/management/group/test_view.py
```diff
@@ -1776,8 +1868,6 @@ def test_system_flag_update_on_add(self, send_kafka_message):
             self.assertEqual(self.defGroup.tenant, self.public_tenant)
             self.assertEqual(response.status_code, status.HTTP_200_OK)
 
-            # New platform default role for tenant created
-            custom_default_group = Group.objects.get(platform_default=True, tenant=self.tenant)
             self.assertEqual(custom_default_group.name, "Custom default access")
             self.assertFalse(custom_default_group.system)
             self.assertEqual(custom_default_group.tenant, self.tenant)
```
alechenninger, line 1873: Maybe add an assert to check that the uuid is the same as the default group uuid in the TenantMapping?

tests/management/group/test_view.py
```diff
@@ -1831,8 +1921,10 @@ def test_system_flag_update_on_add(self, send_kafka_message):
             ]
             kafka_mock.assert_has_calls(notification_messages, any_order=True)
 
+    @override_settings(V2_BOOTSTRAP_TENANT=True)
+    @patch("management.relation_replicator.outbox_replicator.OutboxReplicator._save_replication_event")
     @patch("core.kafka.RBACProducer.send_kafka_message")
-    def test_system_flag_update_on_remove(self, send_kafka_message):
+    def test_system_flag_update_on_remove(self, send_kafka_message, mock_method):
```
alechenninger, line 1927: I like that you adjusted this test case to account for still one remaining role in the group. Maybe we should add back a test case for when we remove all the roles in the default group?

rbac/management/group/definer.py
```diff
@@ -127,7 +137,6 @@ def add_roles(group, roles_or_role_ids, tenant, user=None):
     """Process list of roles and add them to the group."""
```
alechenninger, line 137: I think we can remove the `@transaction.atomic` on add_roles and remove_roles now. They are both now necessarily called within outer atomic blocks so it's adding unnecessary savepoints I think.

lpichler: I removed it for `add_roles`, `remove_role`  functions.

`add_roles`, `remove_role` are not in transaction block for seed process. If I did not overlook anything It looks that whole seed process is not in transaction.

Do you agree with wrapping seed process into transaction in follow up PR? It seems beneficial.

tests/management/group/test_view.py
lpichler: Added 👍

tests/management/group/test_view.py
lpichler: yes, probably there was some evolution - now it is not needed, removed 👍

rbac/management/group/definer.py
alechenninger: I think at least where `add_roles` and `remove_roles` are used in seeding, they are in atomic blocks:

```
def seed_group() -> Tuple[Group, Group]:
    """Create or update default group."""
    with transaction.atomic():
        ...
```

rbac/management/group/relation_api_dual_write_group_handler.py
```diff
@@ -252,11 +280,32 @@ def prepare_to_delete_group(self):
...
+
+        return create_relationship(
+            ("rbac", "workspace"),
+            str(self.default_workspace.id),
+            ("rbac", "role_binding"),
+            str(mapping.default_role_binding_uuid),
+            "binding",
+        )
+
+    def _get_public_tenant(self) -> Optional[Tenant]:
```
alechenninger, line 304:
```suggestion
    def _get_public_tenant(self) -> Tenant:
```

rbac/management/group/definer.py
lpichler: oh, cool I overlooked it in that case, thanks!

rbac/management/group/relation_api_dual_write_group_handler.py
```diff
@@ -123,39 +130,55 @@ def _replicate(self):
...
-            assert role.system is True, "Expected system role. Mappings for custom roles must already be created."
-            binding = V2rolebinding(
-                str(uuid4()),
-                # Assumes same role UUID for V2 system role equivalent.
-                V2role.for_system_role(str(role.uuid)),
-                V2boundresource(("rbac", "workspace"), str(self.default_workspace.id)),
-                groups=frozenset([str(self.group.uuid)]),
+        for role in roles:
+
+            def create_default_mapping():
```
astrozzc, line 143: Can we move this function out of the loop? Each iteration of the loop might create a new function object

astrozzc (approved):
LGTM overall, up to Alec to approve it

rbac/management/group/relation_api_dual_write_group_handler.py
lpichler: you are right it is not worth as there is workspace query, I removed this 👍

rbac/management/group/relation_api_dual_write_group_handler.py
lpichler: good catch! sure It is possible 👍

tests/management/group/test_view.py
lpichler: I agree with there is place for improvements and unification. 

btw I looks there is better to [way](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_args) for call_args.

Let's consider mentioned points and I will create ticket for it 👍

rbac/management/group/relation_api_dual_write_group_handler.py
```diff
@@ -123,39 +129,55 @@ def _replicate(self):
...
                 # Assumes same role UUID for V2 system role equivalent.
-                V2role.for_system_role(str(role.uuid)),
+                V2role.for_system_role(str(system_role.uuid)),
                 V2boundresource(("rbac", "workspace"), str(self.default_workspace.id)),
                 groups=frozenset([str(self.group.uuid)]),
             )
-            mapping = BindingMapping.for_role_binding(binding, role)
+            mapping = BindingMapping.for_role_binding(binding, system_role)
             self.group_relations_to_add.extend(mapping.as_tuples())
             return mapping
```
alechenninger, line 153: I think we can move this just to be a top-level function now if you want (make it private in that case e.g. with _ prefix)

rbac/management/group/relation_api_dual_write_group_handler.py
```diff
@@ -123,39 +129,55 @@ def _replicate(self):
...
 
-        self._update_mapping_for_role(
-            role, update_mapping=add_group_to_binding, create_default_mapping_for_system_role=create_default_mapping
-        )
+        for role in roles:
+
```
alechenninger, line 156:
```suggestion
```

alechenninger (approved):
Minor suggestion but otherwise LGTM! Thanks for this one

rbac/management/group/relation_api_dual_write_group_handler.py
```diff
@@ -123,54 +129,59 @@ def _replicate(self):
...
+        for role in roles:
+            self._update_mapping_for_role_removal(role)
 
     def _update_mapping_for_role_removal(self, role: Role):
         def remove_group_from_binding(mapping: BindingMapping):
             self.group_relations_to_remove.append(mapping.remove_group_from_bindings(str(self.group.uuid)))
 
-        self._update_mapping_for_role(
-            role, update_mapping=remove_group_from_binding, create_default_mapping_for_system_role=lambda: None
-        )
```
alechenninger, line 166: Changing how this works I think may introduce a bug – if this code was called to remove a system role from a group where it wasn't assigned, it would end up getting added. I would keep the old logic where this could be controlled by the arguments to the method.

alechenninger (changes requested):

rbac/management/group/relation_api_dual_write_group_handler.py
lpichler: I see. I removed last commit. I added comment.

alechenninger (approved):
LGTM

PR #1248, opened 2024-10-18
RHCLOUD-30751 Principal Links endpoint fixes

EvanCasey13:
## Link(s) to Jira
https://issues.redhat.com/browse/RHCLOUD-30751

Fixed bug where incorrect links were being returned for the principals endpoint

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


rbac/management/principal/view.py
```diff
@@ -188,7 +187,7 @@ def get(self, request):
                 "first": f"{path}?limit={limit}&offset=0{usernames_filter}",
                 "next": f"{path}?limit={limit}&offset={offset + limit}{usernames_filter}",
                 "previous": f"{path}?limit={limit}&offset={previous_offset}{usernames_filter}",
-                "last": None,
+                "last": f"{path}?limit={limit}&offset={int(limit - int(count))}{usernames_filter}",
```
petracihalova, line 190: if limit < count then offset will be negative

maybe we can reuse the `get_paginated_response()` from [rbac/api/common/pagination.py](https://github.com/RedHatInsights/insights-rbac/blob/af853082447418d58a1cdde5e8bbdf350f8ba0ae/rbac/api/common/pagination.py#L80)

rbac/management/principal/view.py
```diff
@@ -183,12 +182,13 @@ def get(self, request):
                 count = len(data)
             else:
                 count = None
+            last_link = int(count) - int(limit) if (int(count) - int(limit)) >= 0 else 0
```
petracihalova, line 185: please consider better name for the variable, it should describe the value that is inside

rbac/management/principal/view.py
```diff
@@ -183,12 +182,13 @@ def get(self, request):
...
             response_data["meta"] = {"count": count, "limit": limit, "offset": offset}
             response_data["links"] = {
                 "first": f"{path}?limit={limit}&offset=0{usernames_filter}",
                 "next": f"{path}?limit={limit}&offset={offset + limit}{usernames_filter}",
                 "previous": f"{path}?limit={limit}&offset={previous_offset}{usernames_filter}",
-                "last": None,
+                "last": f"{path}?limit={limit}&offset={last_link}{usernames_filter}",
```
petracihalova, line 191: it is needed to fix all links here to work in same way as for other endpoints like `GET /groups/` for example `GET /principals/` for limit=10 and offset=0 (default values) the previous link should be `null`

rbac/management/principal/view.py
lpichler: @EvanCasey13 is it possible to leverage `get_paginated_response ` somehow ?
it looks there some cases which needs to covered.

rbac/management/principal/view.py
```diff
@@ -123,7 +123,8 @@ def get(self, request):
         previous_offset = 0
         if offset - limit > 0:
             previous_offset = offset - limit
-
+        else:
+            None
```
astrozzc, line 127: This seems can be removed

rbac/management/principal/view.py
```diff
@@ -183,12 +184,15 @@ def get(self, request):
                 count = len(data)
             else:
                 count = None
+            last_link_offset = int(count) - int(limit) if (int(count) - int(limit)) >= 0 else 0
```
astrozzc, line 187: Nit: we can put int(count) - int(limit) into a variable so we don't have to calculate it twice

astrozzc:
Oh, as Libor mentioned, can we reuse the get_paginated_response to achieve this to simplify it. An example usage is here: https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/access/view.py#L130-L132

PR #1254, opened 2024-10-24
Set partition key to environment name

wcmitchell:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-35627

Sets the partition key used by Debezium connectors to current
environment name. 

This will keep debezium outbox messages in a guaranteed order per
environment. It will limit our future options for parallelism, but is a
safer approach currently.

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/management/group/relation_api_dual_write_group_handler.py
+++ b/rbac/management/group/relation_api_dual_write_group_handler.py
@@ -113,9 +113,7 @@ class RelationApiDualWriteGroupHandler:
                 ReplicationEvent(
                     event_type=self.event_type,
                     info={"group_uuid": str(self.group.uuid)},
-                    # TODO: need to think about partitioning
-                    # Maybe resource id
-                    partition_key="rbactodo",
+                    partition_key="settings.ENV_NAME",
                     remove=self.group_relations_to_remove,
                     add=self.group_relations_to_add,
                 ),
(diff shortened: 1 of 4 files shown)


rbac/management/group/relation_api_dual_write_group_handler.py
```diff
@@ -113,9 +113,7 @@ def _replicate(self):
                 ReplicationEvent(
                     event_type=self.event_type,
                     info={"group_uuid": str(self.group.uuid)},
-                    # TODO: need to think about partitioning
-                    # Maybe resource id
-                    partition_key="rbactodo",
+                    partition_key="settings.ENV_NAME",
```
lpichler, line 116: @wcmitchell do you mean to have here value of settings.ENV_NAME ? If so, it will be without quotes ` partition_key=settings.ENV_NAME` or [example](https://github.com/RedHatInsights/insights-rbac/blob/93eebbf2d4c9d5119a14e6eed8b81fc7d8e866bf/rbac/management/tenant_service/v2.py#L279)

alechenninger:
Thought on this one: if we want to change later, how would we?

If we changed partition key while running, we could end up reordering replication which would be Bad™.

I guess we'd have to stop all writes, let current outbox log "empty," change the key, resume writes. So an outage.

I'm wondering if there is another way? Otherwise maybe we want to set this to a more scalable value now to avoid the outage. cc @coderbydesign

alechenninger:
See #1277

PR #1258, opened 2024-10-24
RHCLOUD-35914 - Add flag to seed existing unchanged system roles

lennysgarage:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-35914

- Adds a new flag "force-create" that will always replicate a create event for existing system roles.

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


deploy/rbac-clowdapp.yml
```diff
@@ -631,6 +631,19 @@ objects:
           volumeMounts:
           - mountPath: /tmp
             name: users-data
+      - name: seed-roles-perms-groups
+        podSpec:
+          image: quay.io/cloudservices/rbac:latest
+          command:
+            - ./rbac/manage.py seeds
+            - --force-create
```
alechenninger, line 639: You probably need to separate `seeds` part in its own element?

And does this need `--roles` ? I guess not, but maybe helps clarify our intent.

rbac/management/role/definer.py
```diff
@@ -93,6 +93,8 @@ def _make_role(data, dual_write_handler):
             role.access.all().delete()
             role_obj_change_notification_handler(role, "updated")
         else:
+            if force_create_relationships:
+                dual_write_handler.replicate_new_system_role(role)
```
lpichler, line 97: Maybe can we add here some log message also ? I am not sure if we should keep this in case: 
```
 logger.info("No change in system role %s", name)
```

deploy/rbac-clowdapp.yml
lennysgarage: I was taking reference from the job above `- ./rbac/manage.py import_tenant_user_data` which seems to be just the filename, couldn't find other references for it. It doesn't require `--roles`, but added to clarify!

rbac/management/role/definer.py
lennysgarage: I can add a log message something like `logger.info("Replicated system role %s", name)`? I guess we could also return role in the same block to skip the no change log message?

lpichler:
I think it is not issue, but what to do this task separated from current seed process ? (at least to have it in separate function) this will run only at beginning so maybe we don't need to add another if-statement in ordinary seed process.

rbac/management/role/definer.py
lpichler: yes, it sounds great

lennysgarage:
> I think it is not issue, but what to do this task separated from current seed process ? (at least to have it in separate function) this will run only at beginning so maybe we don't need to add another if-statement in ordinary seed process.

@lpichler If I'm understanding correct, you envision when using the flag that the only operations that occur is the create replication events and nothing else sent?

rbac/management/role/definer.py
```diff
@@ -116,19 +120,19 @@ def _make_role(data, dual_write_handler):
...
 
 
-def _update_or_create_roles(roles, dual_write_handler):
+def _update_or_create_roles(roles, dual_write_handler, force_create_relationships=False):
     """Update or create roles from list."""
     current_role_ids = set()
     for role_json in roles:
         try:
-            role = _make_role(role_json, dual_write_handler)
+            role = _make_role(role_json, dual_write_handler, force_create_relationships)
```
lpichler, line 128:
```
if force_create_relationships
    dual_write_handler = new()
    dual_write_handler.replicated_new_role(role)
else
    role = _make_role(role_json, dual_write_handler)
end
```

lpichler:
> > I think it is not issue, but what to do this task separated from current seed process ? (at least to have it in separate function) this will run only at beginning so maybe we don't need to add another if-statement in ordinary seed process.
> 
> @lpichler If I'm understanding correct, you envision when using the flag that the only operations that occur is the create replication events and nothing else sent, just before the rest of seeding is run?

yes, like this https://github.com/RedHatInsights/insights-rbac/pull/1258/files#r1816745831
I meant like don't put into make_role method.

rbac/management/role/definer.py
lennysgarage: Okay the problem I see with this is that we don't actually have the role here to replicate only the json representation. We would still need to get the role like we do in `_make_role`

lpichler: I see! so I am fine with as it is now, thanks!

lpichler (approved):

alechenninger (approved):

PR #1260, opened 2024-10-25
RHCLOUD-35889 parent_id to uuid as PK on model/response

coderbydesign:
## Link(s) to Jira
- [RHCLOUD-35889](https://issues.redhat.com/browse/RHCLOUD-35889)

Have clearer representation on the parent ID being a UUID value on the FK relation for workspaces

### Background for Consideration
[Update the workspace model PK to be a uuid and drop the explicit uuid column](https://github.com/RedHatInsights/insights-rbac/commit/0d95c01863ee669a7793e25e9b747cbe96659dc4) 

This would start a new convention with the v2 APIs and models in order to have
consistent, clear naming, particularly when it comes to FK references.

We currently have the `uuid` field as the self-referencial FK column on the `Workspace`
model. More details around the impetus for changing the naming around IDs can be
found in https://github.com/RedHatInsights/insights-rbac/pull/1257.

These changes offer an alternate approach, since we have no data in stage/production,
where we no longer use the `uuid` as the `lookup_field` in Django, but rather
use a `uuid` as the `id` format. The rationale for not doing this, and having an
explicit `uuid` was primarily for having sequential integers as the PK/FK relations.
However, UUID7 is a time-ordered UUID, eliminating index issues and solving the
need for having distributed ID values across our services.

We're using `uuid-utils` [1] which is a compliant implementation using Rust's
UUID library. There's also an open proposal [2,3] to add it to Python's standard
library.

This updates the model, view and serializer. In order to move the `id` from int
to uuid, we need two migrations:
- one to move the current `id` column, and the `parent` column (because of the FK ref)
  as well as making the current `uuid` column the PK
- a second to then rename the `uuid` column to `id` and add the `parent` FK ref/column back

[1] https://github.com/aminalaee/uuid-utils
[2] https://github.com/python/cpython/issues/89083
[3] https://discuss.python.org/t/add-uuid7-in-uuid-module-in-standard-library/44390

## Local Testing
Unit test coverage + query workspaces (with and without ancestry) to confirm field name change from `uuid` and `parent_uuid` to `id` and `parent_id` respectively.

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


rbac/management/tenant_service/v2.py
```diff
@@ -442,31 +441,30 @@ def _bootstrap_default_access(
...
         relationships = []
 
-        root_workspace_uuid = uuid4()
-        default_workspace_uuid = uuid4()
-
-        root = Workspace(uuid=root_workspace_uuid, tenant=tenant, type=Workspace.Types.ROOT, name="Root Workspace")
+        root = Workspace(tenant=tenant, type=Workspace.Types.ROOT, name="Root Workspace")
         default = Workspace(
-            uuid=default_workspace_uuid,
-            parent_id=root.uuid,
```
alechenninger, line 451: Nice

alechenninger (approved):
LGTM

astrozzc (approved):

PR #1270, opened 2024-10-29
[RHCLOUD-35976] Role removal - fix condition for the system role identification

petracihalova:
[RHCLOUD-35976](https://issues.redhat.com/browse/RHCLOUD-35976)

when we remove the role, we need to check that the role is not a system role (because it is not allowed to remove system roles)

in that check we should identify the system role as a role that belongs to the public tenant


astrozzc (approved):

vbelchio:
QE: Passed (regression)

PR #1273, opened 2024-10-30
RHCLOUD-31117 Added ternary operator to handle no username being pass…

EvanCasey13:
…ed when filtering Principals

## Link(s) to Jira
https://issues.redhat.com/browse/RHCLOUD-31117

Error occurs when trying to filter Principal with type as "service-account" and match criteria as "exact" or "partial" where a username is not provided in the query.

Added a ternary operator to handle this, if no username is provided it will use the requesting users username in the query
`username = user.username if not usernames else usernames[0]`

This subverts the 500 error when trying to filter without a user name, although I think another option would be to set the username to an empty string if its not provided i.e
`username = "" if not usernames else usernames[0]` 

I'm not entirely sure how this would work passing an empty string for the username I think it might return nothing as the query would be 
`/api/rbac/v1/principals/?limit=10&offset=0&match_criteria=exact&status=enabled&admin_only=false&type=service-account&usernames=""`

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


petracihalova (changes requested):
it looks like the request for user based principals without `usernames` query param the `match_criteria` param is ignored

with this change the request for service account based principals without `usernames` query param the `match_criteria` param is not ignored

it should be probably same, so we need to clarify what is expected here
+
I would appreciate some tests for this change

tests/management/principal/test_it_service.py
```diff
@@ -1570,3 +1570,53 @@ def test_merge_principals_it_service_accounts(self) -> None:
...
+        print(count)
+        self.assertEqual(
+            0,
+            count,
+            "unexpected number of service accounts fetched for the tenant",
+        )
+
+    @override_settings(IT_BYPASS_IT_CALLS=True)
+    def test_principal_filtering_without_username_match_criteria_partial_skipped(self):
+        """Test the function under test skips the expected service account when match_criteria is exact without usernames"""
```
petracihalova, line 1601: its small nitpick ... your options for the filter contains matech_criteria = partial and in the docstring is match-criteria is exact ... otherwise LGTM, good job, lets merge it 👍 ❤️

PR #1277, opened 2024-11-01
RHCLOUD-35627: Partition key

alechenninger:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-35627

Building off of #1254 , I fixed the settings lookup and added some abstraction around the key.

The abstraction may be overkill, but given how critical the partition key is to correctness, I was a little uneasy about how open ended the field value was, and how much duplication was going on between various sources of replication events. At least this will make it harder to get wrong and maybe cause a bit more thought come the time we need to scale beyond a single partition. It also gave me a place to document how we might do that in the future :-).

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


rbac/management/relation_replicator/outbox_replicator.py
```diff
@@ -30,6 +32,10 @@
 class OutboxReplicator(RelationReplicator):
     """Replicates relations via the outbox table."""
 
+    def __init__(self, log: Optional["OutboxLog"] = None):
```
astrozzc, line 35: Do you keep the definition of OutboxLog later on purpose?

astrozzc (approved):
LGTM!

rbac/management/relation_replicator/outbox_replicator.py
alechenninger: I had to check myself but yes! Thanks for asking. The OutboxLog defines the protocol for type-safety. Then it can accept either the OutboxWAL (which does the normal DB stuff) or the InMemoryLog which both implement the protocol.

PR #1282, opened 2024-11-04
[WIP][RHCLOUD-36065] Fix limit validation for GET /access/ endpoint

petracihalova:
## Link(s) to Jira
- [RHCLOUD-36065](https://issues.redhat.com/browse/RHCLOUD-36065)

`GET /access/` with invalid limit value returns 500
example:
`GET /api/rbac/v1/access/?application=&limit=aaa`

In this PR I removed the validate_limit_and_offset(params) because
1. we can use default limit and offset validation we already use for other endpoints from class `StandardResultsSetPagination`
2. the usage of  the `validate_limit_and_offset(params)` was not working because the function returns the dict object that was not captured in the function `validate_and_get_param()`

I updated the function so that `limit` and `offset` are validated consistently with our other endpoints using `StandardResultsSetPagination`: if `limit` is not a number, it defaults to 10, and if it exceeds 1000, it is capped at 1000

## Local Testing
call the /access/ endpoint with different values of `limit` query param
`GET /api/rbac/v1/access/?application=&limit=<value>`

- for non-digit value the default `limit=10` is returned -> example `limit=aaa`
- for empty value the default `limit=10` is returned -> example `limit=`
- for value 0 the default `limit=10` is returned -> example `limit=0`
- for negative number the default `limit=10` is returned -> example `limit=-10`
- for number > 1000 the max `limit=1000` is returned -> example `limit=1001`
- for integer between 0 and 1001 the given limit is returned -> example `limit=25`

### DIFF
--- a/rbac/management/access/view.py
+++ b/rbac/management/access/view.py
@@ -137,7 +137,6 @@ class AccessView(APIView):
         """Return the paginator instance associated with the view, or `None`."""
         if not hasattr(self, "_paginator"):
             self._paginator = self.pagination_class()
-            self._paginator.max_limit = None
         return self._paginator
 
     def paginate_queryset(self, queryset):
--- a/rbac/management/access/view.py
+++ b/rbac/management/access/view.py
@@ -143,8 +143,6 @@ class AccessView(APIView):
         """Return a single page of results, or `None` if pagination is disabled."""
         if self.paginator is None:
             return None
-        if "limit" not in self.request.query_params:
-            self.paginator.default_limit = len(queryset)
         return self.paginator.paginate_queryset(queryset, self.request, view=self)
 
     def get_paginated_response(self, data):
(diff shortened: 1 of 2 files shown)


rbac/management/access/view.py
```diff
@@ -138,15 +137,12 @@ def paginator(self):
...
             self._paginator = self.pagination_class()
-            self._paginator.max_limit = None
         return self._paginator
 
     def paginate_queryset(self, queryset):
         """Return a single page of results, or `None` if pagination is disabled."""
         if self.paginator is None:
             return None
-        if "limit" not in self.request.query_params:
-            self.paginator.default_limit = len(queryset)
```
astrozzc, line 149: Oh! Wait! Would this return all records by default?

petracihalova:
I will close this PR and prepare new one only with fix for invalid values for `limit` and we can discuss rest separately

PR #1250, opened 2024-10-18
Require "parent_id" field on Workspace POST

coderbydesign:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-35802

## Description of Intent of Change(s)
We should be requiring `parent_id` on `POST` requests for `Workspace` objects

## Local Testing
Verify unit tests or send a `POST` request with/without `parent_id`.

## Checklist
- [ ] if API spec changes are required, is the spec updated?
- [ ] are there any pre/post merge actions required? if so, document here.
- [x] are theses changes covered by unit tests?
- [ ] if warranted, are documentation changes accounted for?
- [ ] does this require migration changes?
  - [ ] if yes, are they backwards compatible?
- [ ] is there known, direct impact to dependent teams/components?
  - [ ] if yes, how will this be handled?

## Secure Coding Practices Checklist Link
- https://github.com/RedHatInsights/secure-coding-checklist

## Secure Coding Practices Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [ ] General Coding Practices

PR #1284, opened 2024-11-04
RHCLOUD-35890 enforce parent id in workspace create code

coderbydesign:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-35890

This ensures that before saving a workspace (from code directly, the API is covered in https://github.com/RedHatInsights/insights-rbac/pull/1250),
we'll fail and send back a validation error from the model if it's a non-root workspace
and there's no `parent_id` supplied.

## Local Testing
Create a non-root workspace object without a `parent_id`

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


astrozzc (approved):
LGTM!

rbac/management/workspace/model.py
```diff
@@ -55,6 +56,11 @@ def save(self, *args, **kwargs):
         self.full_clean()
         super().save(*args, **kwargs)
 
+    def clean(self):
+        """Validate the model."""
+        if self.type != self.Types.ROOT and self.parent_id is None:
+            raise ValidationError({"parent_id": ("This field cannot be blank.")})
```
alechenninger, line 62:
```suggestion
            raise ValidationError({"parent_id": ("This field cannot be blank for non-Root-type Workspaces.")})
```
(or something like that)

tests/management/workspace/test_model.py
```diff
@@ -37,90 +37,124 @@ def tearDown(self):
...
     def test_multiple_standard_per_tenant(self):
         """Test tenant can have multiple standard workspaces"""
         try:
             for n in ["1", "2", "3"]:
-                Workspace.objects.create(name=f"T1 Standard Workspace Number {n}", type="standard", tenant=self.tenant)
+                Workspace.objects.create(
+                    name=f"T1 Standard Workspace Number {n}",
+                    type=Workspace.Types.STANDARD,
+                    tenant=self.tenant,
+                    parent=self.tenant_1_default_workspace,
```
alechenninger, line 136: Standard workspaces should be allowed to go underneath the root workspace, too. FWIW, I don't think I see any tests that exercise that. Not strictly related to this PR but might be good to add something for that.

alechenninger (approved):
Minor suggestions but otherwise LGTM.

I like that this caught some tests which weren't necessarily the "correct" set up. It does make me wonder if we should encapsulate Workspace (or V2 models) creation with a Factory type or static method (essentially named constructor) of some kind to enforce / avoid repeating some of these conventions. I dunno what that would look like exactly but it's something perhaps worth experimenting with when we start the prototype for the remaining V2 models.

tests/management/workspace/test_model.py
coderbydesign: That's fair. I'll add a test case.

rbac/management/workspace/model.py
coderbydesign: I'm happy to make the change to be more explicit. I was going for consistency here since this emulates the validation error raised by the API on required fields. That said, the clients are different (end user vs internal caller).

lpichler (approved):

PR #1286, opened 2024-11-05
RHCLOUD-36017 Custom roles cannot be created or updated using existin…

EvanCasey13:
…g display_name for any system roles

## Link(s) to Jira
https://issues.redhat.com/browse/RHCLOUD-36017

When trying to create or update a custom role through the API you cannot use the same 'display_name' as any system role that exists. 

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/management/role/view.py
+++ b/rbac/management/role/view.py
@@ -269,6 +269,11 @@ class RoleViewSet(
         """
         self.validate_role(request)
         try:
+            if Role.objects.filter(display_name=request.data.get("name"), system=True).exists():
+                return Response(
+                    {"Error": f"The role name '{request.data.get('name')}' is reserved, please use another name"},
+                    status=status.HTTP_400_BAD_REQUEST,
+                )
             with transaction.atomic():
                 return super().create(request=request, args=args, kwargs=kwargs)
         except IntegrityError as e:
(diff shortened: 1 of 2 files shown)


alechenninger:
I wonder if we should prefix all custom roles names automatically? Or try to differentiate them some different way? (an icon in the gui? or some "role type" column that designates a role as system or custom?)

Checking for system role names when a custom role is created doesn't necessarily prevent collisions from happening – for example a customer could name a role "Workspace Auditor" and then we later create a system role with the same name. Maybe that's fine and we're okay with that?

But at least for V2 Roles and GUI, I wonder if we should think about this differently? Curious what you think @coderbydesign ?

petracihalova:
I suggest we take a step back and discuss what is expected here ... our goal is to prevent the creation of a new role with the same name as an existing custom role within the same tenant or system role ... however it should be still possible to create custom role with same name if this role belongs to different tenant .. we also need to consider the role’s name and display_name fields and how these may affect our approach ensuring we are handling both fields correctly

rbac/management/role/view.py
```diff
@@ -645,3 +646,15 @@ def delete_policies_if_no_role_attached(self, role):
...
                 policy.delete()
+
+    def restrict_role_create_update_if_system_role_exists(self, request):
+        """If a system role has the same display_name when a role is being created or updated restrict the operation."""
+        req_method = request.method
+        req_name = request.data.get("name")
+        public_tenant = Tenant.objects.get(tenant_name="public")
+        base_role_queryset = Role.objects.filter(display_name=req_name, tenant__in=[request.tenant, public_tenant])
+        req_system_role_exists = base_role_queryset.filter(system=True).exists()
+        if req_system_role_exists and req_method == "PUT":
```
vbelchio, line 657: Due to the specifications, we have PUT and PATCH methods that are able to change role information.
Is PATCH also covered here?

PR #1287, opened 2024-11-05
[RHCLOUD-35997] Add admin endpoint to delete/list workspace/tenantmapping

astrozzc:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-35997

The what, why and how.

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/api/utils.py
+++ b/rbac/api/utils.py
@@ -1,10 +1,23 @@
 """Helper utilities for api module."""
 
+import logging
 from django.db import transaction
-from management.utils import account_id_for_tenant
+from django.shortcuts import get_object_or_404
 
 from api.models import Tenant
+from management.models import BindingMapping
+from management.tenant_mapping.model import TenantMapping
+from management.workspace.model import Workspace
+from management.utils import account_id_for_tenant
 
+
+
+RESOURCE_MODEL_MAPPING = {
+    "workspace": Workspace,
+    "mapping": TenantMapping,
+    "binding": BindingMapping,
+}
+logger = logging.getLogger(__name__)
 
 def populate_tenant_account_id():
     """Populate tenant's account id's."""
@@ -13,3 +26,25 @@ def populate_tenant_account_id():
         with transaction.atomic():
             tenant.account_id = account_id_for_tenant(tenant)
             tenant.save()
+
+
+def migration_resource_deletion(resource, org_id):
+    resource_objs = RESOURCE_MODEL_MAPPING[resource].objects.all()
+    if org_id:
+        tenant = get_object_or_404(Tenant, org_id=org_id)
+        if resource == "binding":
+            resource_objs = resource_objs.filter(role__tenant=tenant)
+        else:
+            resource_objs = resource_objs.filter(tenant=tenant)
+    else:
+        if resource == "binding":
+            public_tenant = Tenant.objects.get(tenant_name="public")
+            resource_objs = resource_objs.exclude(role__tenant=public_tenant)
+    if resource == "workspace":
+        # Have to delete the ones without children first or deletion will fail
+        logger.info("Deleting workspaces without children.")
+        while(resource_objs.filter(children=None).exists()):
+            resource_objs.filter(children=None).delete()
+        logger.info("All workspaces without children removed.")
+    resource_objs.delete()
+    logger.info(f"Resources of type {resource} deleted.", status=204)
--- a/rbac/api/utils.py
+++ b/rbac/api/utils.py
@@ -4,14 +4,10 @@ import logging
 
 from django.db import transaction
 from django.shortcuts import get_object_or_404
-
-from api.models import Tenant
 from management.models import BindingMapping
 from management.tenant_mapping.model import TenantMapping
 from management.utils import account_id_for_tenant
 from management.workspace.model import Workspace
-from management.utils import account_id_for_tenant
-
 
 from api.models import Tenant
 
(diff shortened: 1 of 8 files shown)


rbac/internal/views.py
```diff
@@ -504,6 +509,52 @@ class SentryDiagnosticError(Exception):
     pass
 
 
+def migration_resources(request):
```
coderbydesign, line 512: As we chatted about, we'll probably want to kick this off in a worker process so that if the deletes exceed the http timeout, it won't be an issue.

rbac/api/utils.py
```diff
@@ -1,15 +1,49 @@
...
             tenant.account_id = account_id_for_tenant(tenant)
             tenant.save()
+
+
+def migration_resource_deletion(resource, org_id):
+    resource_objs = RESOURCE_MODEL_MAPPING[resource].objects.all()
+    if org_id:
+        tenant = get_object_or_404(Tenant, org_id=org_id)
+        if resource == "binding":
+            resource_objs = resource_objs.filter(role__tenant=tenant)
```
alechenninger, line 35: This will miss bindings for system roles because those are in the public tenant. Is this intentional?

I think we probably want removing the binding to be based off of the tenant of the resource, however we don't have a foreign key here, and we might not always know the owning Tenant of the bound resource. For now, we actually do, though, since the resource is always a Workspace. Maybe we could just start there? That is, find all workspaces for a tenant, and then delete all the binding mappings for those workspaces (using resource type namespace, resource type name, and resource id fields to filter).

In the long run, we probably won't be able to support removing bindings by org, unless we make it clear it's limited to only certain narrow filter options (like custom roles, or workspace bindings).

It also may be safer for us to just blow away everything, but it probably doesn't hurt to have something org specific in case it is needed. It just has to be used very carefully of course.

rbac/api/utils.py
```diff
@@ -1,15 +1,49 @@
...
+        else:
+            resource_objs = resource_objs.filter(tenant=tenant)
+    else:
+        if resource == "binding":
+            public_tenant = Tenant.objects.get(tenant_name="public")
+            resource_objs = resource_objs.exclude(role__tenant=public_tenant)
```
alechenninger, line 41: Why are these excluded in this case? It would mean we still leave bindings around if they have been bound with system roles. Do we want that?

rbac/internal/views.py
```diff
@@ -504,6 +505,51 @@ class SentryDiagnosticError(Exception):
...
+    elif request.method == "GET":
+        resource_objs = resource_model.objects.all()
+        if org_id:
+            tenant = get_object_or_404(Tenant, org_id=org_id)
+            if tenant.tenant_name == "public":
+                return HttpResponse("Invalid request, tenant cannot be the public one.", status=400)
+            if resource == "binding":
+                resource_objs = resource_objs.filter(role__tenant=tenant)
+            else:
+                resource_objs = resource_objs.filter(tenant=tenant)
```
alechenninger, line 546: Maybe we can factor this org_id queryset filtering to a common place for both reads and deletes?

rbac/api/utils.py
astrozzc: I had the wrong impression the bindings for system roles have to be kept. I am ignoring the org_id for bindingmapping for now

rbac/internal/views.py
astrozzc: Updated

rbac/internal/views.py
```diff
@@ -55,6 +56,10 @@
 
 logger = logging.getLogger(__name__)
 TENANTS = TenantCache()
+RESOURCE_MODEL_MAPPING = {
+    "workspace": Workspace,
+    "mapping": TenantMapping,
```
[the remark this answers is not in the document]
astrozzc: Added

rbac/api/utils.py
```diff
@@ -1,15 +1,52 @@
...
             tenant.account_id = account_id_for_tenant(tenant)
             tenant.save()
+
+
+def get_resources(resource, org_id):
+    """Get queryset by org_id."""
+    queryset = RESOURCE_MODEL_MAPPING[resource].objects.all()
+    if resource != "binding" and org_id:
+        tenant = get_object_or_404(Tenant, org_id=org_id)
+        queryset = queryset.filter(tenant=tenant)
```
alechenninger, line 37: Maybe abort or something if its the binding? Just so we have to be intentional if we want to delete everything, and someone doesn't think they're only deleting bindings for one org and end up accidentally deleting everything.

astrozzc: Updated

lpichler:
@astrozzc Does this need update to open api spec ?

rbac/api/utils.py
```diff
@@ -1,15 +1,54 @@
...
+        if resource == "binding":
+            raise ValueError("Binding cannot be filtered by org_id.")
+        tenant = get_object_or_404(Tenant, org_id=org_id)
+        queryset = queryset.filter(tenant=tenant)
+    return queryset
+
+
+def migration_resource_deletion(resource, org_id):
+    """Delete migration related resources."""
+    resource_objs = get_resources(resource, org_id)
```
lpichler, line 45: Does `resource` needs to be checked if exists in `RESOURCE_MODEL_MAPPING` ? Maybe this check should in `get_resources` function ?

rbac/internal/views.py
```diff
@@ -545,6 +546,44 @@ def list_bindings_for_role(request):
     return HttpResponse(json.dumps(result), content_type="application/json", status=200)
 
 
+def migration_resources(request):
+    """View or delete specific resources related to migration.
+
+    DELETE /_private/api/utils/migration_resources/?resource=xxx&org_id=xxx
+    GET /_private/api/utils/migration_resources/?resource=xxx&org_id=xxx&limit=1000
+    optinos of resource: workspace, mapping(tenantmapping), binding(bindingmapping)
```
lpichler, line 554: typo `optinos`

astrozzc: Oops, forgot to update this, good catch!

alechenninger (approved):
LGTM

PR #1288, opened 2024-11-05
[RHCLOUD-35855] System roles assigments in migrator

lpichler:
This generates and replicates relations for system role assignments of custom default groups and custom groups.
Also:
- guard condition to allow migration of when read only mode is enabled 
- logic for migration user-group assignments was replaced by logic in  dual write handler.

### Links
[RHCLOUD-35855](https://issues.redhat.com/browse/RHCLOUD-35855)

### DIFF
--- a/tests/migration_tool/tests_migrate.py
+++ b/tests/migration_tool/tests_migrate.py
@@ -25,8 +25,7 @@ from django.test import TestCase, override_settings
 from api.models import Tenant
 from management.models import *
 from migration_tool.migrate import migrate_data
-from management.utils import clear_pk
-from setuptools.command.easy_install import sys_executable
+from management.group.definer import seed_group, clone_default_group_in_public_schema
 
 
 class MigrateTests(TestCase):
@@ -35,12 +34,31 @@ class MigrateTests(TestCase):
     def setUp(self):
         """Set up the utils tests."""
         super().setUp()
+        # public tenant
         public_tenant = Tenant.objects.get(tenant_name="public")
-        default_group = Group.objects.create(name="default", tenant=public_tenant, platform_default=True, system=True)
+
+        # system roles
+        self.system_role_1 = Role.objects.create(
+            name="System Role 1", platform_default=True, tenant=public_tenant, system=True
+        )
+        self.system_role_2 = Role.objects.create(
+            name="System Role 2", platform_default=True, system=True, tenant=public_tenant
+        )
+        # default group
+        default_group, _ = seed_group()
+        # permissions
         # This would be skipped
         permission1 = Permission.objects.create(permission="app1:hosts:read", tenant=public_tenant)
         permission2 = Permission.objects.create(permission="inventory:hosts:write", tenant=public_tenant)
-        # Two organization
+        Access.objects.bulk_create(
+            [
+                Access(permission=permission1, role=self.system_role_2, tenant=public_tenant),
+                Access(permission=permission2, role=self.system_role_2, tenant=public_tenant),
+            ]
+        )
+
+        # two organizations
+        # tenant 1 - org_id=1234567
         self.tenant = Tenant.objects.create(org_id="1234567", tenant_name="tenant")
         self.root_workspace = Workspace.objects.create(
             type=Workspace.Types.ROOT, tenant=self.tenant, name="Root Workspace"
@@ -48,12 +66,10 @@ class MigrateTests(TestCase):
         self.default_workspace = Workspace.objects.create(
             type=Workspace.Types.DEFAULT, tenant=self.tenant, name="Default Workspace", parent=self.root_workspace
         )
-
-        another_tenant = Tenant.objects.create(org_id="7654321")
-
         # setup data for organization 1234567
         self.workspace_id_1 = "123456"
         self.workspace_id_2 = "654321"
+
         # This role will be skipped because it contains permission with skipping application
         self.role_a1 = Role.objects.create(name="role_a1", tenant=self.tenant)
         self.access_a11 = Access.objects.create(permission=permission1, role=self.role_a1, tenant=self.tenant)
@@ -88,44 +104,18 @@ class MigrateTests(TestCase):
         self.policy_a2 = Policy.objects.create(name="System Policy_a2", group=self.group_a2, tenant=self.tenant)
         self.policy_a2.roles.add(self.role_a2)
 
-        self.system_role_1 = Role.objects.create(
-            name="System Role 1", platform_default=True, tenant=public_tenant, system=True
-        )
-        self.policy_a2.roles.add(self.system_role_1)
-        self.policy_a2.save()
+        # tenant 2 - org_id=7654321
+        another_tenant = Tenant.objects.create(org_id="7654321")
 
         # setup data for another tenant 7654321
         self.role_b = Role.objects.create(name="role_b", tenant=another_tenant)
         self.access_b = Access.objects.create(permission=permission2, role=self.role_b, tenant=another_tenant)
-        self.system_role_2 = Role.objects.create(name="System Role 2", system=True, tenant=public_tenant)
-        Access.objects.bulk_create(
-            [
-                Access(permission=permission1, role=self.system_role_2, tenant=public_tenant),
-                Access(permission=permission2, role=self.system_role_2, tenant=public_tenant),
-            ]
-        )
+
+        self.policy_a2.roles.add(self.system_role_1)
+        self.policy_a2.save()
 
         # create custom default group
-        group = default_group
-        default_policy = Policy.objects.create(
-            system=True, name=f"System Policy for Group {group.uuid}", group=group, tenant=self.tenant
-        )
-        group.policies.add(default_policy)
-        tenant_default_policy = group.policies.get(system=True)
-        group.name = "Custom default access"
-        group.system = False
-        group.tenant = self.tenant
-        group.uuid = uuid4()
-        clear_pk(group)
-        clear_pk(tenant_default_policy)
-        tenant_default_policy.uuid = uuid4()
-        tenant_default_policy.name = "System Policy for Group {}".format(group.uuid)
-        tenant_default_policy.tenant = self.tenant
-        group.save()
-        tenant_default_policy.group = group
-        tenant_default_policy.save()
-        tenant_default_policy.roles.add(self.system_role_2)
-        self.custom_default_group = group
+        self.custom_default_group = clone_default_group_in_public_schema(default_group, self.tenant)
 
     @override_settings(REPLICATION_TO_RELATION_ENABLED=True, PRINCIPAL_USER_DOMAIN="redhat", READ_ONLY_API_MODE=True)
     @patch("management.relation_replicator.logging_replicator.logger")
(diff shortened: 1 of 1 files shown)


rbac/management/relation_replicator/relation_replicator.py
```diff
@@ -52,6 +52,7 @@ class ReplicationEventType(str, Enum):
     MIGRATE_CUSTOM_ROLE = "migrate_custom_role"
     MIGRATE_TENANT_GROUPS = "migrate_tenant_groups"
     CUSTOMIZE_DEFAULT_GROUP = "customize_default_group"
+    ASSIGN_SYSTEM_ROLE_IN_MIGRATOR = "assign_system_role_in_migrator"
```
alechenninger, line 55: maybe `migrate_system_role_assignment` (so all `migrate_` events have a common prefix / convention?)

rbac/migration_tool/migrate.py
```diff
@@ -152,6 +154,19 @@ def migrate_data_for_tenant(tenant: Tenant, exclude_apps: list, replicator: Rela
...
 
+    public_default_roles = Role.objects.filter(platform_default=True, tenant=Tenant.objects.get(tenant_name="public"))
+
+    with transaction.atomic():
+        for group in tenant.group_set.all():
+            dual_write_handler = RelationApiDualWriteGroupHandler(
+                group, ReplicationEventType.ASSIGN_SYSTEM_ROLE_IN_MIGRATOR
+            )
+            if group.platform_default is True:
+                dual_write_handler.generate_relations_to_add_roles(public_default_roles)
```
alechenninger, line 165: I don't think we can add all default roles, because they may have removed some of these.

But, if you just go by the policy like you do below, I think you should be okay. So I think you can just take this part out.

rbac/migration_tool/migrate.py
```diff
@@ -152,6 +154,19 @@ def migrate_data_for_tenant(tenant: Tenant, exclude_apps: list, replicator: Rela
         logger.info(f"Migration completed for role: {role.name} with UUID {role.uuid}.")
     logger.info(f"Migrated {roles.count()} roles for tenant: {tenant.org_id}")
 
+    public_default_roles = Role.objects.filter(platform_default=True, tenant=Tenant.objects.get(tenant_name="public"))
+
+    with transaction.atomic():
+        for group in tenant.group_set.all():
```
alechenninger, line 160: I wonder if we can combine this with migrate_users_for_groups in some way, which already iterates through (almost) all tenant groups.

rbac/migration_tool/migrate.py
```diff
@@ -152,6 +154,19 @@ def migrate_data_for_tenant(tenant: Tenant, exclude_apps: list, replicator: Rela
         logger.info(f"Migration completed for role: {role.name} with UUID {role.uuid}.")
     logger.info(f"Migrated {roles.count()} roles for tenant: {tenant.org_id}")
 
+    public_default_roles = Role.objects.filter(platform_default=True, tenant=Tenant.objects.get(tenant_name="public"))
+
+    with transaction.atomic():
```
alechenninger, line 159: It doesn't hurt to be all in one tx, but nothing else in the migrator is currently dealing with concurrency. It is assumed that it runs without other concurrent writes.

You mentioned the issue about the middleware possibly doing concurrent writes which I'm not sure if you've looked into more yet but deserves some thought.

rbac/migration_tool/migrate.py
```diff
@@ -94,10 +97,52 @@ def migrate_role(
     return relationships, v2_role_bindings
 
 
-def migrate_users_for_groups(tenant: Tenant) -> list[common_pb2.Relationship]:
-    """Write users relationship to groups."""
+def migrate_system_role_assignments_for_groups(groups: list[Group]) -> list[common_pb2.Relationship]:
+    """Generate system role assignments for groups."""
+    with transaction.atomic():
```
alechenninger, line 102: Is this required because of the select for update?

If not, I would probably take it out because it might imply it's needed or effective in some way that it isn't. Migration does NOT account for concurrent updates, so I am worried about implying that it does.

If it's required, maybe just add a comment to that effect (required due to select for update even though migrator does not generally account for concurrent updates and so should NOT be run outside of read-only mode.)

lpichler: yes it is and I wanted to ask what scope for transaction block should be in this case.

rbac/migration_tool/migrate.py
```diff
@@ -94,10 +97,52 @@ def migrate_role(
...
+        for group in groups:
+            if dual_write_handler is None:
+                dual_write_handler = RelationApiDualWriteGroupHandler(group, None)
+            else:
+                dual_write_handler.group = group
+            system_roles = group.roles().filter(system=True)
+            if system_roles.exists() > 0:
+                dual_write_handler.generate_relations_to_add_roles(system_roles)
+                relationships.extend(dual_write_handler.group_relations_to_add)
+    return relationships
```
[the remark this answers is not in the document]
lpichler: yes I know that dual write handler is not designed for multiple groups and I wanted to avoid to do same query for default workspace for each group. Do you think it is fine with this fact ?

rbac/migration_tool/migrate.py
alechenninger: It shouldn't really matter I guess since we shouldn't do this with concurrent writes. Not too big though, so a TX per group or per tenant is probably okay. Per group might be better but probably would require testing to confirm. (Normally there is a TX _per every single SQL operation_ so per group is not too many.)

alechenninger: We also may want to add a check in the beginning of the migrator that requires read only mode is set otherwise it doesn't run as a safety mechanism.

rbac/migration_tool/migrate.py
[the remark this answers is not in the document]
lpichler: I think this is fine, we can cache default workspace in follow up PR.

rbac/migration_tool/migrate.py
```diff
@@ -94,31 +97,66 @@ def migrate_role(
...
         ReplicationEvent(
             event_type=ReplicationEventType.MIGRATE_TENANT_GROUPS,
             info={"tenant": tenant.org_id},
             partition_key=PartitionKey.byEnvironment(),
-            add=tuples,
+            add=user_relationships,
+        )
+    )
+
+    system_role_relationships = migrate_system_role_assignments_for_groups(groups)
```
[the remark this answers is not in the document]
lpichler: sure, it is simpler, done 👍

lpichler:
@alechenninger I addressed you suggesting, can you re-review ? 🙏

rbac/migration_tool/migrate.py
```diff
@@ -94,30 +96,28 @@ def migrate_role(
...
-        user_set: Iterable[Principal] = group.principals.all()
-        for user in user_set:
-            if (relationship := group.relationship_to_principal(user)) is not None:
-                relationships.append(relationship)
-    return relationships
+def migrate_groups_for_tenant(tenant: Tenant, replicator: RelationReplicator):
+    """Generate user relationships and system role assignments for groups in a tenant."""
+    groups = tenant.group_set.all()
+    for group in groups:
+        with transaction.atomic():
```
alechenninger, line 103:
```suggestion
        # The migrator does not generally deal with concurrency control,
        # but we require an atomic block due to use of select_for_update in the dual write handler.
        with transaction.atomic():
```

rbac/migration_tool/migrate.py
```diff
@@ -94,30 +96,28 @@ def migrate_role(
...
+            dual_write_handler = RelationApiDualWriteGroupHandler(
+                group, ReplicationEventType.MIGRATE_TENANT_GROUPS, replicator=replicator
+            )
+            if not group.platform_default:
+                dual_write_handler.generate_relations_to_add_principals(group.principals.all())
+            if group.system is False and group.admin_default is False:
+                system_roles = group.roles().filter(system=True)
+                if system_roles.exists() > 0:
```
alechenninger, line 111:
```suggestion
                if system_roles.exists():
```

Another option might be

```suggestion
                if any(True for _ in system_roles)
```

This is a little less readable but probably more efficient I think. If I understand django, it wouldn't do another query, it would just start/continuing iterating through the query, using results already cached.

rbac/migration_tool/migrate.py
```diff
@@ -155,6 +155,10 @@ def migrate_data_for_tenant(tenant: Tenant, exclude_apps: list, replicator: Rela
 
 def migrate_data(exclude_apps: list = [], orgs: list = [], write_relationships: str = "False"):
     """Migrate all data for all tenants."""
+    if not settings.READ_ONLY_API_MODE:
+        logger.info("Read-only API mode is required. READ_ONLY_API_MODE must be set to true.")
```
alechenninger, line 159:
```suggestion
        logger.fatal("Read-only API mode is required. READ_ONLY_API_MODE must be set to true.")
```

alechenninger: I'd argue it's fatal because it does not function at all – if you expected it to migrate, it didn't.

tests/migration_tool/tests_migrate.py
```diff
@@ -31,7 +36,7 @@ def setUp(self):
         """Set up the utils tests."""
         super().setUp()
         public_tenant = Tenant.objects.get(tenant_name="public")
-        Group.objects.create(name="default", tenant=public_tenant, platform_default=True)
+        default_group = Group.objects.create(name="default", tenant=public_tenant, platform_default=True, system=True)
```
alechenninger, line 39: I wonder if we can use `seed_group()` instead in the setup (probably after the roles are created then)? This would ensure test setup stays consistent with the app and maybe reduce some of this code.

rbac/migration_tool/migrate.py
```diff
@@ -94,30 +96,28 @@ def migrate_role(
...
+            dual_write_handler = RelationApiDualWriteGroupHandler(
+                group, ReplicationEventType.MIGRATE_TENANT_GROUPS, replicator=replicator
+            )
+            if not group.platform_default:
+                dual_write_handler.generate_relations_to_add_principals(group.principals.all())
+            if group.system is False and group.admin_default is False:
+                system_roles = group.roles().filter(system=True)
+                if system_roles.exists() > 0:
+                    dual_write_handler.generate_relations_to_add_roles(system_roles)
+            dual_write_handler.replicate()
```
alechenninger, line 113: We can address this in a follow-up PR if you would like, but along the lines of RHCLOUD-36014, I wonder if we should avoid calling this if we know there were no system_roles added and no principals added?

We could add that logic inside `replicate()`, but then it makes you wonder if we should really warn there? And in that case it's redundant because the outbox replicator already warns. So it feels like we should avoid calling replicate if we know there's nothing to replicate, otherwise the replicate method will be "suspicious" in some way :)

rbac/migration_tool/migrate.py
lpichler: yes, I am confirming it removes one query.  Evaluation or query happens only in the condition. Thanks!

rbac/migration_tool/migrate.py
lpichler: yes, `.replicate()` should not be called when it is possible to determine that it is not needed otherwise we cannot capture unintended "empty" replication.  

I made changes here. Existence checks are moved outside the transaction block to avoid initializing the dual write handler when it is not needed. These checks are also not needed before the `generate_relations_XXX` method, as it they don't perform any work on empty arrays (system_role, principals)."

tests/migration_tool/tests_migrate.py
lpichler: I added `seed_group` and `clone_default_group_in_public_schema`

it is possible use `seed_group()`, function creates also admin default group

alechenninger (approved):
LGTM!

I have another branch where I've added another migrator test which is a little more exacting to be sure, but I'll open a PR for that separately. Thank you and nice work!

PR #1291, opened 2024-11-07
RHCLOUD-34512: Replicate cross account access to Relations

alechenninger:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-34512

See jira.

As of now this is just a draft that outlines some tests.

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


rbac/api/cross_access/util.py
```diff
@@ -60,9 +70,9 @@ def get_cross_principal_name(target_org, user_id):
     return f"{target_org}-{user_id}"
 
 
-def create_principal_with_tenant(principal_name, associate_tenant):
+def create_principal_with_tenant(principal_name, associate_tenant, user_id):
     """Create cross-account principal in tenant."""
     cross_account_principal, _ = Principal.objects.get_or_create(
-        username=principal_name, cross_account=True, tenant=associate_tenant
+        username=principal_name, cross_account=True, tenant=associate_tenant, user_id=user_id
```
lpichler, line 76: user_id is stored in cross account principal and this principal is used in dual write handler.

rbac/api/cross_access/relation_api_dual_write_cross_access_handler.py
```diff
@@ -0,0 +1,175 @@
...
+                role, update_mapping=remove_principal_from_binding, create_default_mapping_for_system_role=lambda: None
+            )
+
+    def _update_mapping_for_system_role(
+        self,
+        role: Role,
+        update_mapping: Callable[[BindingMapping], None],
+        create_default_mapping_for_system_role: Callable[[], Optional[BindingMapping]],
+    ):
+        assert role.system is True, "Expected system role."
```
alechenninger, line 146: It's probably worth this being an if check and raising an exception in this case just because it would be really bad if it somehow wasn't a system role (e.g. through some path that is missed in testing)

alechenninger:
I didn't have time to do a detailed review, but a few thoughts:

- Overall, looks really good
- I'm not sure about setting the user_id on the cross-account Principal. We would end up with multiple Principals for the same user_id. It's probably fine, but I wonder if instead we want a foreign key to the Principal for the user. It might be mostly an aesthetic thing. But for example, if we needed to fix the user_id ever (which, should never happen, so it's a stretch) we would have to find all the right Principal objects to update. Or, if we ever wanted to add a unique index on user_id (e.g. to prevent race conditions in Principal creation if we stopped storing username) we wouldn't be able to. So it seems like a hygiene thing – maybe not strictly necessary, but might be a good idea.
- I think this should work, but it would be good to add tests for when removing the last group from a system role, that it does not remove the binding if there is a cross account request. I was starting to add this test myself but got caught up in the mocking problem and could not finish in time.

alechenninger:
@lpichler I can't approve the PR since I opened it but it LGTM! Nice work

lpichler (approved):

rbac/api/cross_access/view.py
```diff
@@ -262,7 +253,13 @@ def update_status(self, car, status):
         car.status = status
         if status == "approved":
             create_cross_principal(car.user_id, target_org=car.target_org)
-
+            cross_account_roles = car.roles.all()
+            if any(True for _ in cross_account_roles):
```
coderbydesign, line 257: A bit of a nit, but I think it would be more performant (assuming I'm understanding the intent here) to do something like this with [`exists()`](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#django.db.models.query.QuerySet.exists):

```suggestion
            if cross_account_roles.exists():
```

Using `any()` _will_ return early, but it executes the query on the queryset which in this case is a select on all roles for a car, where `exists()` will limit the query to a single row.

coderbydesign (commented):
Just one nit, but otherwise makes sense!

PR #1294, opened 2024-11-08
[RHCLOUD-36013] Add internal admin endpoint to bootstrap tenant

astrozzc:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-36013

The what, why and how.

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/internal/views.py
+++ b/rbac/internal/views.py
@@ -40,6 +40,7 @@ from management.principal.proxy import (
     bop_request_status_count,
     bop_request_time_tracking,
 )
+from management.relation_replicator.outbox_replicator import OutboxReplicator
 from management.role.serializer import BindingMappingSerializer
 from management.tasks import (
     migrate_data_in_worker,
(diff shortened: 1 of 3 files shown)


lpichler (approved):

rbac/internal/views.py
```diff
@@ -498,6 +501,26 @@ def data_migration(request):
...
+    if request.method != "POST":
+        return HttpResponse('Invalid method, only "POST" is allowed.', status=405)
+    logger.info("Running bootstrap tenant.")
+
+    org_id = request.GET.get("org_id")
+    if not org_id:
+        return HttpResponse('Invalid request, must supply the "org_id" query parameter.', status=400)
+    acct = request.GET.get("acct", "")
+    bootstrap_service = V2TenantBootstrapService(OutboxReplicator())
+    bootstrap_service._get_or_bootstrap_tenant(org_id, acct)
```
alechenninger, line 520: Is it possible to use a different method here? This one is private, so it's a little surprising to be used outside of the class. It may change or be removed and not expect to break this endpoint.

Two options I can think of:

- use `bootstrap_tenant` – this requires a Tenant, but we could do nothing if the Tenant does not already exist, which might make sense. I don't think we want to fake a new org ID.
- if the tenant is not present, you could also optionally use `new_bootstrapped_tenant` – this should do mostly the same thing as get_or_bootstrap but is a public method. I see though this method misses a few things for creating the tenant – needs to set ready flag and tenant name. If we need to be able to create tenants I would fix that method up and use that one. And we also might wan tto require an explicit flag to allow creating a new Tenant from this endpoint to prevent accidents.

WDYT?

PR #1295, opened 2024-11-09
NOJIRA: Test through tuple traversal

alechenninger:
## Link(s) to Jira
- n/a

I wanted to test the migrator a bit more, without relying on implementation detail like BindingMappings. To do this I experimented with some change to InMemoryTuples to allow traversing the tuples so we can test relationships without necessarily having to first lookup every ID, which is helpful when we don't know (or care) about specific IDs (e.g. they are random), but mainly the relationships between objects.

This should be generally helpful when testing replication and tuple mapping.

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


rbac/migration_tool/in_memory_tuples.py
```diff
@@ -25,15 +25,202 @@ class RelationTuple(NamedTuple):
...
+        return {key: TupleSet(self._full_set, value) for key, value in grouped_tuples.items()}
+
+    def find_group_with_tuples(
+        self,
+        predicates: List[RelationPredicate],
+        group_by: Callable[[RelationTuple], T],
+        group_filter: Callable[[T], bool] = lambda _: True,
+        require_full_match: bool = False,
+        match_once: bool = True,
+    ) -> Tuple[dict[T, "TupleSet"], dict[T, "TupleSet"]]:
```
alechenninger, line 95: These methods are just moved from InMemoryTuples.

Basically we reorganized things a little bit so that results from InMemoryTuples return a TupleSet type which can be further filtered. The novelty of this PR is that it adds a `traverse_subjects` method in this queryset which is not a filter but a traversal.

rbac/migration_tool/in_memory_tuples.py
```diff
@@ -25,15 +25,202 @@ class RelationTuple(NamedTuple):
...
+            TupleSet: A set of matched RelationTuples.
+        """
+        # For each tuple, find all the tuples where the subject is the resource
+        # of the current tuple
+        matched: set[RelationTuple] = set()
+
+        for tuple in self:
+            # Intentionally use full set here to traverse to tuples which may not be in this set
+            traversed = self._full_set.find_tuples(
+                resource(tuple.subject_type_namespace, tuple.subject_type_name, tuple.subject_id)
```
alechenninger, line 206: The key bit here is that this uses the full set of tuples to search, not just what was filtered in this result set. This result set is just used as the base for traversal (using the `resource` predicate for each subject in this set).

PR #1297, opened 2024-11-11
[RHCLOUD-36189] Cross account request limit fix

petracihalova:
## Link(s) to Jira
- [RHCLOUD-36189](https://issues.redhat.com/browse/RHCLOUD-36189)

when an invalid value for `limit` or `offset` is used in the request to the `GET /cross-account-requests/` endpoint, the server returns a 500 error

after change the default values are returned (limit=10, offset=0)

In this PR I removed the `validate_limit_and_offset()` because
- we can use default limit and offset validation as we already use for other endpoints from class `StandardResultsSetPagination`
- the usage of the `validate_limit_and_offset()` was not working because the function returns the dict object that was not captured in the code

## Local Testing
send request to the endpoint 
`GET /api/rbac/v1/cross-account-requests/?limit=<value>&offset=<value>`
with different values for the limit and offset

### DIFF
--- a/rbac/api/cross_access/view.py
+++ b/rbac/api/cross_access/view.py
@@ -22,7 +22,7 @@ from django_filters import rest_framework as filters
 from management.filters import CommonFilters
 from management.models import Role
 from management.principal.proxy import PrincipalProxy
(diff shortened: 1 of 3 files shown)


lpichler (approved):
LGTM, good find!

PR #1313, opened 2024-11-15
Fix replication counter trigger

mjholder:
## Link(s) to Jira
-

The what, why and how.
This makes the _save_replication_event function atomic. This is required for on_commit to be called. Right now the replication function is not a transaction so it won't trigger on_commit. With the transaction.atomic decorator it makes it atomic.
There is a performance risk with this change according to the [Django docs](https://docs.djangoproject.com/en/5.1/topics/db/transactions/#tying-transactions-to-http-requests)

The reason the test for this function is passing is because each TestCase is wrapped in a transaction behind the scenes.

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/management/relation_replicator/outbox_replicator.py
+++ b/rbac/management/relation_replicator/outbox_replicator.py
@@ -72,6 +72,7 @@ class OutboxReplicator(RelationReplicator):
 
         return {"relations_to_add": add_json, "relations_to_remove": remove_json}
 
+    @transaction.atomic
     def _save_replication_event(
         self,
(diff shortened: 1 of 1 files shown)


alechenninger:
Looks like this is a different problem per slack discussion

PR #1314, opened 2024-11-16
NOJIRA: FIx makefile help text

alechenninger:
The assignment operator is optional but only recognized by newer versions of make. Apparently I have an ancient make version, perhaps what is bundled with macos. This should help devs in similar situations.


astrozzc (approved):
Works for me now when typing `make help`

lpichler (approved):

PR #1323, opened 2024-11-20
RHCLOUD-36364: Only commit to outbox from user listener while advisory lock acquired

alechenninger:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-36364

We need the listener process to commit outbox changes in order. Specifically we need to avoid the race that:

1. Process P1 receives a message for user U1 and looks up the user's current state (admin=true)
2. The user state changes (admin=false)
3. Process P2 receives a message for user U1 and looks up the user's current state (admin=false)
4. P2 commits this change
5. P1 commits the old change

In this case the state would be inconsistent until the next user update, potentially indefinitely.

This avoids this by requiring an advisory lock to commit the change:

1. P1 locks
2. P2 cannot acquire the lock
3. P1 commits

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/management/principal/cleaner.py
+++ b/rbac/management/principal/cleaner.py
@@ -21,6 +21,7 @@ import os
 import ssl
 from typing import Optional
 
+from py import log
 import xmltodict
 from django.conf import settings
 from django.db import connection, transaction
@@ -189,6 +190,7 @@ def process_umb_event(frame, umb_client: Stomp, bootstrap_service: TenantBootstr
         # This is locked per transaction to ensure another listener process does not run concurrently.
         if not _lock_listener():
             # If there is another listener, let it run and abort this one.
+            logger.info("process_umb_event: Another listener is running. Aborting.")
             return False
 
         data_dict = xmltodict.parse(frame.body)
@@ -225,12 +227,14 @@ def process_principal_events_from_umb(bootstrap_service: Optional[TenantBootstra
         if not str(e).startswith(("Already connected", "Already subscribed")):
             raise e
 
-    while UMB_CLIENT.canRead(15):  # Check if queue is empty, 15 sec timeout
-        frame = UMB_CLIENT.receiveFrame()
-        if not process_umb_event(frame, UMB_CLIENT, bootstrap_service):
-            break
-    UMB_CLIENT.disconnect()
-    logger.info("process_tenant_principal_events: Principal event processing finished.")
+    try:
+        while UMB_CLIENT.canRead(15):  # Check if queue is empty, 15 sec timeout
+            frame = UMB_CLIENT.receiveFrame()
+            if not process_umb_event(frame, UMB_CLIENT, bootstrap_service):
+                break
+    finally:
+        UMB_CLIENT.disconnect()
+        logger.info("process_tenant_principal_events: Principal event processing finished.")
 
 
 def _lock_listener() -> bool:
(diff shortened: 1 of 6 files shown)


rbac/management/principal/cleaner.py
```diff
@@ -211,8 +225,17 @@ def process_principal_events_from_umb(bootstrap_service: Optional[TenantBootstra
         if not str(e).startswith(("Already connected", "Already subscribed")):
             raise e
 
-    while UMB_CLIENT.canRead(2):  # Check if queue is empty, two sec timeout
+    while UMB_CLIENT.canRead(15):  # Check if queue is empty, 15 sec timeout
         frame = UMB_CLIENT.receiveFrame()
-        process_umb_event(frame, UMB_CLIENT, bootstrap_service)
+        if not process_umb_event(frame, UMB_CLIENT, bootstrap_service):
+            return
```
astrozzc, line 231: should we close the client before returning? UMB_CLIENT.disconnect()

astrozzc (approved):
LGTM! Just have one question

rbac/management/principal/cleaner.py
alechenninger: Yeah definitely good catch

alechenninger: Fixed and added a test for it.

rbac/management/principal/cleaner.py
```diff
@@ -211,8 +225,17 @@ def process_principal_events_from_umb(bootstrap_service: Optional[TenantBootstra
...
     UMB_CLIENT.disconnect()
     logger.info("process_tenant_principal_events: Principal event processing finished.")
+
+
+def _lock_listener() -> bool:
+    """Attempt to acquire a lock for the listener and if acquired return True, else False."""
+    with connection.cursor() as cursor:
+        cursor.execute("SELECT pg_try_advisory_xact_lock(%s);", [LOCK_ID])
+        result = cursor.fetchone()
+    return result[0]  # Returns True if lock acquired, False otherwise
```
lpichler, line 241: can result be `None` ?

rbac/management/principal/cleaner.py
```diff
@@ -211,8 +225,17 @@ def process_principal_events_from_umb(bootstrap_service: Optional[TenantBootstra
...
-        process_umb_event(frame, UMB_CLIENT, bootstrap_service)
+        if not process_umb_event(frame, UMB_CLIENT, bootstrap_service):
+            break
     UMB_CLIENT.disconnect()
     logger.info("process_tenant_principal_events: Principal event processing finished.")
+
+
+def _lock_listener() -> bool:
+    """Attempt to acquire a lock for the listener and if acquired return True, else False."""
+    with connection.cursor() as cursor:
```
lpichler, line 238: maybe we can catch exception here ?

rbac/management/principal/cleaner.py
lpichler: Do we want to log situation when lock is not acquired ?

alechenninger: Dig it! Thanks for the questions.

rbac/management/principal/cleaner.py
alechenninger: Good question! I added a finally block above which should handle disconnecting properly. Otherwise we can let the exception propogate I think.

coderbydesign (approved):

PR #1328, opened 2024-11-22
Lazily mark tenants ready + optimization

coderbydesign:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-36441

As part of the Kessel migration, we've increased our tenant count 100x and need to ensure that we:

- [ ] DO NOT set ready=True during bulk user import, to ensure tenants which haven't used console before are not initialized as ready, to exclude them from other queries.
- [x] set ready=True in middleware if we do find an existing tenant
- [ ] set ready=True in middleware if we are creating the tenant from a user-initiated request
- [x] update notifications tenant query to exclude tenants which are not ready
- [x] update seed tenant query to exclude tenants which are not ready
- [ ] set all current tenants where there are no principals to ready=False in stage DB
- [ ] test coverage

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/management/seeds.py
+++ b/rbac/management/seeds.py
@@ -73,7 +73,7 @@ def purge_cache():
     from rbac.settings import MAX_SEED_THREADS
 
     with concurrent.futures.ThreadPoolExecutor(max_workers=MAX_SEED_THREADS) as executor:
-        tenants = Tenant.objects.all()
+        tenants = Tenant.objects.filter(ready=True)
         tenant_count = tenants.count()
         for idx, tenant in enumerate(list(tenants)):
             progress = f"[{idx + 1} of {tenant_count}]."
--- a/rbac/management/notifications/notification_handlers.py
+++ b/rbac/management/notifications/notification_handlers.py
@@ -57,7 +57,7 @@ def notify_all(event_type, payload):
     """Notify all tenants."""
     # To avoid memory overloaded, use iterator:
     # https://docs.djangoproject.com/en/4.0/ref/models/querysets/#django.db.models.query.QuerySet.iterator
-    for tenant in Tenant.objects.exclude(tenant_name="public").iterator():
+    for tenant in Tenant.objects.exclude(tenant_name="public").filter(ready=True).iterator():
         # Tenant name pattern is acct12345
         notify(event_type, payload, tenant.org_id)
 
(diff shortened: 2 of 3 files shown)


rbac/management/seeds.py
```diff
@@ -73,7 +73,7 @@ def purge_cache():
     from rbac.settings import MAX_SEED_THREADS
 
     with concurrent.futures.ThreadPoolExecutor(max_workers=MAX_SEED_THREADS) as executor:
-        tenants = Tenant.objects.all()
+        tenants = Tenant.objects.filter(ready=True)
         tenant_count = tenants.count()
         for idx, tenant in enumerate(list(tenants)):
```
alechenninger, line 78: Should we iterate this query too anyway?

alechenninger:
I think we will also need to index ready on Tenant

rbac/internal/integration/views.py
```diff
@@ -50,7 +50,7 @@ def modified_only_filter(self, queryset, field, modified_only):
 class TenantViewSet(viewsets.GenericViewSet, mixins.ListModelMixin):
     """Tenant view set."""
 
-    queryset = Tenant.objects.all()
+    queryset = Tenant.objects.exclude(ready=False, tenant_name="public")
```
alechenninger, line 53: maybe leave off public tenant exclusion? probably makes the query (much?) less efficient

alechenninger: We found there is actually an index on tenant_name after all, so this should be fine

rbac/management/seeds.py
alechenninger: This will be removed in #1329

rbac/migration_tool/migrate.py
```diff
@@ -188,7 +189,7 @@ def migrate_data(exclude_apps: list = [], orgs: list = [], write_relationships:
         return
 
     count = 0
-    tenants = Tenant.objects.exclude(tenant_name="public")
```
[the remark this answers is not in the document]
lpichler: it looks that ready flag doesn't help here - it looks we need to improve query to exclude tenants which does not need to be replicated otherwise it will go over 13M tenants  with no roles, groups,.. and it will take hours.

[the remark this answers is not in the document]
lpichler: Answered in meeting, existing tenants will have set ready=True.

lpichler (approved):

PR #1366, opened 2024-12-03
RHCLOUD-36577: Restrap the boot

alechenninger:
## Link(s) to Jira
- [RHCLOUD-36577](https://issues.redhat.com/browse/RHCLOUD-36577)

In order to resync data after replication loss, allow re-bootstrapping a tenant which republishes its bootstrap outbox event.

This factors out some of the tuple construction logic so it can be reused without necessarily creating net-new Workspaces, which we don't want to do in this case.

I also added some additionally logging which could help recover from instances of replication loss.

**What is replication loss?**

Replication loss is when tuples which are meant to be sent to Relations for some reason will never be sent. For example, if the outbox is used with Debezium (what we use now), this happens if the replication slot is lost.

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/internal/views.py
+++ b/rbac/internal/views.py
@@ -521,12 +521,13 @@ def bootstrap_tenant(request):
     logger.info("Running bootstrap tenant.")
 
     org_id = request.GET.get("org_id")
+    force = bool(request.GET.get("force", "false"))
     if not org_id:
         return HttpResponse('Invalid request, must supply the "org_id" query parameter.', status=400)
     with transaction.atomic():
         tenant = get_object_or_404(Tenant, org_id=org_id)
         bootstrap_service = V2TenantBootstrapService(OutboxReplicator())
-        bootstrap_service.bootstrap_tenant(tenant)
+        bootstrap_service.bootstrap_tenant(tenant, force=force)
     return HttpResponse(f"Bootstrap tenant with org_id {org_id} finished.", status=200)
 
 
--- a/rbac/internal/views.py
+++ b/rbac/internal/views.py
@@ -513,17 +513,29 @@ def data_migration(request):
 def bootstrap_tenant(request):
     """View method for bootstrapping a tenant.
 
-    POST /_private/api/utils/bootstrap_tenant/?org_id=12345
-    org_id is required,
+    POST /_private/api/utils/bootstrap_tenant/?org_id=12345&force=false
+
+    org_id:
+        (required) The org_id of the Tenant to bootstrap.
+
+    force:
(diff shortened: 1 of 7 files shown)


rbac/internal/views.py
```diff
@@ -521,12 +521,13 @@ def bootstrap_tenant(request):
     logger.info("Running bootstrap tenant.")
 
     org_id = request.GET.get("org_id")
+    force = bool(request.GET.get("force", "false"))
```
lpichler, line 524: can we put this param  into[ internal api spec](https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/internal/specs/openapi.json) ?

rbac/management/tenant_service/v2.py
```diff
@@ -262,6 +269,27 @@ def _bootstrap_tenant(self, tenant: Tenant) -> BootstrappedTenant:
 
         return BootstrappedTenant(tenant, mapping, default_workspace=default_workspace, root_workspace=root_workspace)
 
+    def _replicate_bootstrap(self, tenant: Tenant, mapping: TenantMapping):
+        """Replicate the bootstrapping of a tenant."""
+        built_in_workspaces = Workspace.objects.filter(
+            tenant=tenant, type__in=[Workspace.Types.ROOT, Workspace.Types.DEFAULT]
```
coderbydesign, line 275: Definitely not in scope for this, but it might be nice at some point if we had a custom model manager that could return the built-ins with something like `Workspace.objects.built_ins(tenant=tenant)`. A future DRY task, maybe. 😄

astrozzc (approved):

rbac/management/tenant_service/v2.py
```diff
@@ -262,6 +269,27 @@ def _bootstrap_tenant(self, tenant: Tenant) -> BootstrappedTenant:
...
+        )
+        root = next(ws for ws in built_in_workspaces if ws.type == Workspace.Types.ROOT)
```
coderbydesign, line 277: With the above, we might also be able to have a manager add to the query to make root and default return ordering deterministic so we could just have assign them in one line, but again, not necessary here.

coderbydesign (approved):

astrozzc (approved):

PR #1374, opened 2024-12-06
RHCLOUD-36669: Use user_id in case it is not available from principal

astrozzc:
## Link(s) to Jira

- [RHCLOUD-36669](https://issues.redhat.com/browse/RHCLOUD-36669) Cannot process message with inactive user where Principal does not have user ID populated yet
- [RHCLOUD-36670](https://issues.redhat.com/browse/RHCLOUD-36670) UMB messages which fail bootstrapping block the entire queue

The what, why and how.

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/management/principal/cleaner.py
+++ b/rbac/management/principal/cleaner.py
@@ -47,10 +47,15 @@ CERT_LOC = "/opt/rbac/rbac/management/principal/umb_certs/cert.pem"
 KEY_LOC = "/opt/rbac/rbac/management/principal/umb_certs/key.pem"
 LOCK_ID = 42  # For Keith, with Love
 
-METRIC_STOMP_MESSAGE_TOTAL = "stomp_messages_total"
-umb_message_processed_count = Counter(
-    METRIC_STOMP_MESSAGE_TOTAL,
+METRIC_STOMP_MESSAGES_ACK_TOTAL = "stomp_messages_ack_total"
+METRIC_STOMP_MESSAGES_NACK_TOTAL = "stomp_messages_nack_total"
+stomp_messages_ack_total = Counter(
+    METRIC_STOMP_MESSAGES_ACK_TOTAL,
     "Number of stomp UMB messages processed",
+)
+stomp_messages_nack_total = Counter(
+    METRIC_STOMP_MESSAGES_NACK_TOTAL,
+    "Number of stomp UMB messages that failed to be processed",
 )
 
 
@@ -200,35 +205,24 @@ def process_umb_event(frame, umb_client: Stomp, bootstrap_service: TenantBootstr
             logger.info("process_umb_event: Another listener is running. Aborting.")
             return False
 
-        data_dict = xmltodict.parse(frame.body)
-        canonical_message = data_dict.get("CanonicalMessage")
-        if canonical_message:
-            try:
-                user = retrieve_user_info(canonical_message)
-            except Exception as e:  # Skip processing and leave the it to be processed later
-                logger.error("process_umb_event: Error retrieving user info: %s", str(e))
-                return True
+        try:
+            data_dict = xmltodict.parse(frame.body)
+            canonical_message = data_dict.get("CanonicalMessage")
 
+            user = retrieve_user_info(canonical_message)
             # By default, only process disabled users.
             # If the setting is enabled, process all users.
             if not user.is_active or settings.PRINCIPAL_CLEANUP_UPDATE_ENABLED_UMB:
                 # If Tenant is not already ready, don't ready it
-                try:
-                    bootstrap_service.update_user(user, ready_tenant=False)
-                except ValueError as e:
-                    logger.error("process_umb_event: Error updating user: %s", str(e))
-                    capture_exception(e)
-                    umb_client.nack(frame)
-                    umb_message_processed_count.inc()
-                    return True
-        else:
-            # Message is malformed.
-            # Ensure we dont block the entire queue by discarding it.
-            # TODO: this is not the only way a message can be malformed
-            pass
+                bootstrap_service.update_user(user, ready_tenant=False)
+            umb_client.ack(frame)
+            stomp_messages_ack_total.inc()
+        except Exception as e:
+            logger.error("process_umb_event: Error processing umb message : %s", str(e))
+            capture_exception(e)
+            umb_client.nack(frame)
+            stomp_messages_nack_total.inc()
 
-    umb_client.ack(frame)
-    umb_message_processed_count.inc()
     return True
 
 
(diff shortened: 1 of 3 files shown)


rbac/management/tenant_service/v2.py
```diff
@@ -221,7 +222,11 @@ def _disable_user_in_tenant(self, user: User):
 
             for group in principal.group.all():
                 group.principals.remove(principal)
-                tuples_to_remove.append(group.relationship_to_principal(principal))
+                tupple = group.relationship_to_principal(principal)
+                if tupple is None:
+                    error_message = f"principal {principal.username} has no user id."
+                else:
+                    tuples_to_remove.append(tupple)
```
alechenninger, line 229: Trying to think through our options here... This is really an illegal state – we should never get to the point where we're trying to remove a user that we don't have an ID for. We could...

1. Delete the Principal even if we can't remove the tuple. This would keep us in an illegal state, where the graph is inconsistent with RBAC. But at least we would have removed the user from RBAC's v1 access result.
2. Keep the Principal. This keeps the access graph consistent with RBAC, but leaves the user around even though we know it should be removed.

I'm wondering if 1 is really better? What is the impact of leaving the principal? What is the impact of deleting it?

For example, there may be some code that assumes if a Principal isn't there, then it's already been removed from the graph. Actually I think there's an example of that right here in this method:

```python
            pass
```

This would actually make it hard for us to repair the data, because we'd no longer be able to rely on a message correcting it, we'd have to manually fix the graph.

We could change this to try to remove the tuple from the graph anyway, but that has its own side effect of adding redundant work in the happy path.

I'm wondering if it's better to raise an exception early after all? What do you think?

rbac/management/principal/cleaner.py
```diff
@@ -212,7 +213,14 @@ def process_umb_event(frame, umb_client: Stomp, bootstrap_service: TenantBootstr
             # If the setting is enabled, process all users.
             if not user.is_active or settings.PRINCIPAL_CLEANUP_UPDATE_ENABLED_UMB:
                 # If Tenant is not already ready, don't ready it
-                bootstrap_service.update_user(user, ready_tenant=False)
+                try:
+                    bootstrap_service.update_user(user, ready_tenant=False)
+                except ValueError as e:
```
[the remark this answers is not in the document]
alechenninger: Especially with `nack` I think it is safer to broaden this, since the nack should trigger a limited number of retries. This depends on the policy of the broker, but it can be configured to a desirable behavior. For example in our case this should retry ~6 times before sending to a DLQ and unblocking the head of the queue.

rbac/management/principal/cleaner.py
```diff
@@ -212,7 +213,14 @@ def process_umb_event(frame, umb_client: Stomp, bootstrap_service: TenantBootstr
...
+                    logger.error("process_umb_event: Error updating user: %s", str(e))
+                    capture_exception(e)
+                    umb_client.nack(frame)
+                    umb_message_processed_count.inc()
```
alechenninger, line 222: I wonder if this may be misleading, since the message isn't actually "processed" in the "done" sense. Maybe we should add a label to the counter for this? I'm not sure what the best name. Maybe `frame="ack"` vs `frame="nack"`?

alechenninger: Another option is to use a different metric.

rbac/management/tenant_service/v2.py
```diff
@@ -221,7 +221,10 @@ def _disable_user_in_tenant(self, user: User):
 
             for group in principal.group.all():
                 group.principals.remove(principal)
-                tuples_to_remove.append(group.relationship_to_principal(principal))
+                # The user id might be None for the principal so we use user instead
+                tupple = group.relationship_to_principal(user)
+                if tupple is None:
```
alechenninger, line 226:
```suggestion
                tuple = group.relationship_to_principal(user)
                if tuple is None:
```

rbac/management/tenant_service/v2.py
```diff
@@ -221,7 +221,10 @@ def _disable_user_in_tenant(self, user: User):
...
+                    raise ValueError(f"tuple is None for user {user.username}")
```
alechenninger, line 227:
```suggestion
                    raise ValueError(f"relationship_to_principal is None for user {user.username}")
```

rbac/management/principal/cleaner.py
```diff
@@ -199,28 +205,24 @@ def process_umb_event(frame, umb_client: Stomp, bootstrap_service: TenantBootstr
...
                 # If Tenant is not already ready, don't ready it
                 bootstrap_service.update_user(user, ready_tenant=False)
-        else:
-            # Message is malformed.
-            # Ensure we dont block the entire queue by discarding it.
-            # TODO: this is not the only way a message can be malformed
-            pass
+            umb_client.ack(frame)
+            stomp_messages_ack_total.inc()
+        except Exception as e:
```
alechenninger, line 220: We may want to introduce two types of exceptions:

1. Those that are retryable
2. Those that aren't

If we think about categorizing the problems in terms of HTTP status codes, 1 would be 5xx errors, and 2 would be 4xx errors.

This is similar to how KafkaConnect deals with exceptions also.

I say this because of broker retry behavior may not always be optimized for both cases (as in our case).

For now I can think of two ways forward:

1. Keep as is and see how it works out in practice. Best case temporary failures are successfully retried and we don't have to intervene. Worst case bad messages pause the queue for a while while the broker retries, impacting UX.
2. Use ack instead of nack here, and manually retry failures. Best case bad messages never hold up the queue, however worst case we have a lot of missed messages due to transient errors.

Either way you go, we will I think want to monitor and based on experience categorize the exceptions so we can ack vs nack accordingly eventually.

alechenninger (approved):

alechenninger (approved):
Looks great!

I think we'll want to iterate on the except / nack / ack handling piece. Left a comment, but I leave which option up to the team.

rbac/management/principal/cleaner.py
alechenninger: Maybe something we could try for now would be like

```
        except OSError as e:
            # log as error & nack for retry...
        except Exception as e:
            # log as critical & ack so not retried
```

(but up to you, just doing some research on python exception hierarchy :))

astrozzc: Cool, yea, a bit leaning towards the first option because the number of missed messages might be large, would be hard to manually retry all the failures. Better late then never, lol.

Yea, as you mentioned, we can iterate on the except / nack / ack handling piece later, I created a JIRA task to https://issues.redhat.com/browse/RHCLOUD-36806

PR #1383, opened 2024-12-11
[RHCLOUD-33414] - restrict the Custom default access group renaming

EvanCasey13:
## Link(s) to Jira
https://issues.redhat.com/browse/RHCLOUD-33414

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


rbac/management/group/view.py
```diff
@@ -435,6 +455,7 @@ def update(self, request, *args, **kwargs):
             }
         """
         validate_uuid(kwargs.get("uuid"), "group uuid validation")
+        self.restrict_custom_default_group_renaming(request)
```
astrozzc, line 458: If you put this after the group = self.get_object(), you can check if the group is custom default group by group.platform_default. Which will simplify your method a lot

rbac/management/group/view.py
```diff
@@ -438,6 +458,10 @@ def update(self, request, *args, **kwargs):
         self.protect_system_groups("update")
 
         group = self.get_object()
+
+        if group.platform_default:
+            self.restrict_custom_default_group_renaming(request)
```
astrozzc, line 463: Since you get the group above, you can put it into the method. If the name/description different from group.name/description, raise validation error

rbac/management/group/view.py
```diff
@@ -225,6 +225,26 @@ def protect_system_groups(self, action, group=None):
...
+    def restrict_custom_default_group_renaming(self, request, group):
+        """Restrict users from changing the name or description of the Custom default group."""
+        method = request.method
+        name = request.data.get("name")
+        description = request.data.get("description")
+        invalid_methods = {"PUT"}
+        is_updating = method in invalid_methods
+        try:
+            custom_default_group = Group.objects.get(name="Custom default group")
+            is_custom_group = (
```
astrozzc, line 237: you can use `group.platform_default and not group.system` to determine it is custom default group. You can

rbac/management/group/view.py
```diff
@@ -225,6 +225,17 @@ def protect_system_groups(self, action, group=None):
             error = {key: [_(message)]}
             raise serializers.ValidationError(error)
 
+    def restrict_custom_default_group_renaming(self, request, group):
```
lpichler, line 228: This method does multiple  things:
- check HTTP method
- check custom group
- raise Validation error 

Is there a change to split it ? More details in comments

Can this logic be part only of update operation ? (So we don't need check on  HTTP method)

rbac/management/group/view.py
```diff
@@ -225,6 +225,17 @@ def protect_system_groups(self, action, group=None):
...
+        """Restrict users from changing the name or description of the Custom default group."""
+        method = request.method
+        invalid_methods = {"PUT"}
+        is_updating = method in invalid_methods
+        if group.platform_default and not group.system and is_updating:
+            key = "Custom default group"
```
lpichler, line 234: is to possible to keep this consistent across app ? key is `detail` - it refers to this https://datatracker.ietf.org/doc/html/rfc7807#section-3.1

tests/management/group/test_view.py
```diff
@@ -187,6 +187,11 @@ def setUp(self):
         self.defPolicy = Policy(name="defPolicy", system=True, tenant=self.public_tenant, group=self.defGroup)
         self.defPolicy.save()
 
+        self.Custom_default_group = Group(
```
lpichler, line 190: nit: I think preferred is lower case.

rbac/management/group/serializer.py
```diff
@@ -43,6 +43,21 @@ def get_roleCount(self, obj):
         """Role count for the serializer."""
         return obj.role_count()
 
+    def is_custom_default_group(self, group):
+        """Check if the group is custom default or not."""
+        if group.platform_default and not group.system:
+            return True
+        return False
```
astrozzc, line 50: You can do return group.platform_default and not group.system

astrozzc (approved):
LGTM!

rbac/management/group/serializer.py
```diff
@@ -43,6 +43,21 @@ def get_roleCount(self, obj):
...
+
+    def restrict_custom_default_group_renaming(self, group):
+        """Restrict users from changing the name or description of the Custom default group."""
+        is_custom_default = self.is_custom_default_group(group)
+        if is_custom_default:
+            key = "detail"
+            message = "Updating the name or description of Custom default group is restricted"
```
petracihalova, line 57: nitpick: it would be nice to have the name of the group in single quotes to distinguish what is name group

rbac/management/group/serializer.py
```diff
@@ -43,6 +43,21 @@ def get_roleCount(self, obj):
...
+            error = {key: [(message)]}
```
petracihalova, line 58: I believe the parentheses here are not needed
`error = {key: [message]}`

petracihalova (changes requested):
hi Evan, thank you for the change, but we need to review it a little bit, please after you try to change the `Custom default access` group and you get 400 (that is expected), check the data from database and you will see the record was changed, that is because `self.restrict_custom_default_group_renaming(group)` is called after `group = super().update(instance, validated_data)`. 

```
    def update(self, instance, validated_data):
        """Update the role object in the database."""
        group = super().update(instance, validated_data)
        self.restrict_custom_default_group_renaming(group)
        group_obj_change_notification_handler(self.context["request"].user, group, "updated")
        return group
```

I think we should use different direction here, I am open to discussion because I am not so experienced here and maybe we can use something similar as `protect_system_groups()` in the view ([here](https://github.com/RedHatInsights/insights-rbac/blob/878fbb81f0997274384e27c01d063952c64e71ba/rbac/management/group/view.py#L438)), and we should still keep in mind that it is possible to delete this group and add and remove roles from it

one option is to add new condition into `protect_system_groups()`, for example something like

```
    def protect_system_groups(self, action, group=None):
        """Deny modifications on system groups."""
        if group is None:
            group = self.get_object()
        if group.system:
            key = "group"
            message = "{} cannot be performed on system groups.".format(action.upper())
            error = {key: [_(message)]}
        if group.platform_default and action == "update":
            key = "group"
            message = "It is not allowed to change name or description of the 'Custom default access' group."
            error = {key: [_(message)]}
```
cc: @lpichler @astrozzc

astrozzc:
> hi Evan, thank you for the change, but we need to review it a little bit, please after you try to change the `Custom default access` group and you get 400 (that is expected), check the data from database and you will see the record was changed, that is because `self.restrict_custom_default_group_renaming(group)` is called after `group = super().update(instance, validated_data)`.
> 
> ```
>     def update(self, instance, validated_data):
>         """Update the role object in the database."""
>         group = super().update(instance, validated_data)
>         self.restrict_custom_default_group_renaming(group)
>         group_obj_change_notification_handler(self.context["request"].user, group, "updated")
>         return group
> ```
> 
> I think we should use different direction here, I am open to discussion because I am not so experienced here and maybe we can use something similar as `protect_system_groups()` in the view ([here](https://github.com/RedHatInsights/insights-rbac/blob/878fbb81f0997274384e27c01d063952c64e71ba/rbac/management/group/view.py#L438)), and we should still keep in mind that it is possible to delete this group and add and remove roles from it
> 
> one option is to add new condition into `protect_system_groups()`, for example something like
> 
> ```
>     def protect_system_groups(self, action, group=None):
>         """Deny modifications on system groups."""
>         if group is None:
>             group = self.get_object()
>         if group.system:
>             key = "group"
>             message = "{} cannot be performed on system groups.".format(action.upper())
>             error = {key: [_(message)]}
>             raise serializers.ValidationError(error)
>         if group.platform_default and action == "update":
>             key = "group"
>             message = "It is not allowed to change name or description of the 'Custom default access' group."
>             error = {key: [_(message)]}
>             raise serializers.ValidationError(error)
> ```
> 
> cc: @lpichler @astrozzc

astrozzc:
Yea, thanks Petra, good catch!!! And I agree putting the function inside the view seems to be appropriate.

The `protect_system_groups` has been used in several places, it is better have a separate function `restrict_custom_default_group_renaming` next to `protect_system_groups`, so that we don't run this checking of renaming logic only when necessary.

astrozzc:
Yea, thanks Petra, good catch!!! And I agree putting the function inside the view seems to be appropriate.

The `protect_system_groups` has been used in several places, it is better have a separate function `restrict_custom_default_group_renaming` next to `protect_system_groups`, so that we don't run this checking of renaming logic only when necessary.

vbelchio:
QE: Passed

tests/management/group/test_view.py
```diff
@@ -808,6 +809,16 @@ def test_update_default_group(self):
...
+    def test_update_custom_default_group(self):
+        """Test that Custom default group is protected from updates"""
+        customDefGroup = Group(name="customDefGroup", platform_default=True, system=False, tenant=self.tenant)
+        customDefGroup.save()
+        url = reverse("v1_management:group-detail", kwargs={"uuid": customDefGroup.uuid})
+        test_data = {"name": "new_name" + "_updated", "description": "new_description" + "_updated"}
+        client = APIClient()
+        response = client.put(url, test_data, format="json", **self.headers)
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
```
petracihalova, line 821: nice job 👍 I have small nitpick ... I would like to check not only status code but the text of the err message to be 100% sure that rbac returns 400 from correct condition, for example something like this
```
 expected_err_message = "Updating the name or description of 'Custom default group' is restricted"
 self.assertEqual(response.data.get("errors")[0].get("detail"), expected_err_message)
```
you can add it if you want but for now we can merge this PR

petracihalova (approved):

PR #779, opened 2023-01-04
[RHCLOUD-22827]Add logs for internal RBAC endpoint requests

abaiken:
## Link(s) to Jira
- [RHCLOUD-22827](https://issues.redhat.com/browse/RHCLOUD-22827)

## Description of Intent of Change(s)
Add logging for internal endpoints & add `is_internal` field to the log object 

## Local Testing
1. Bring up the project as you would normally (I used docker-compose)
2. Test non-internal endpoint by running `curl http://localhost:8000/api/rbac/v1/roles/ -v`
In the rbac server logs you should see something similar to the following: 
```
rbac-server_1     | [2023-01-05 14:48:48,178] INFO: {'method': 'GET', 'path': '/api/rbac/v1/roles/', 'status': 200, 'request_id': None, 'account': '10001', 'org_id': '11111', 'username': 'user_dev', 'is_admin': True, 'is_system': False, 'is_internal': False}
```
3. Test an internal endpoint by running `curl hhttp://localhost:8000/_private/api/tenant/ -v -H 'User-Type: associate'`
In the rbac server logs you should see something similar to the following: 
```
rbac-server_1     | [2023-01-05 14:58:16,676] INFO: {'method': 'GET', 'path': '/_private/api/tenant/', 'status': 200, 'request_id': None, 'account': None, 'org_id': None, 'username': 'associate_dev@bar.com', 'is_admin': True, 'is_system': False, 'is_internal': True}
```
## Checklist
- [ ] if API spec changes are required, is the spec updated?
- [ ] are there any pre/post merge actions required? if so, document here.
- [ ] are theses changes covered by unit tests?
- [ ] if warranted, are documentation changes accounted for?
- [ ] does this require migration changes?
  - [ ] if yes, are they backwards compatible?
- [ ] is there known, direct impact to dependent teams/components?
  - [ ] if yes, how will this be handled?

## Secure Coding Practices Checklist Link
- https://github.com/RedHatInsights/secure-coding-checklist

## Secure Coding Practices Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [ ] General Coding Practices

PR #1400, opened 2024-12-18
Refactor process_response metrics and logging

mjholder:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-36533

The what, why and how.
I refactored the process_response to use one block of code for both paths (internal and external endpoints). Before it would shortcut for internal endpoints but the logic for both paths were identical. So i opted for one block.

The `behalf` variable would always be false. Now it uses the same logic as the function `IdentityHeaderMiddleware.log_request` to determine if it should be `system` or `principal`. That is lines 328-339 of the same file.

## Local Testing
I ran the same tests from #779 which is where this functionality originated
## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/rbac/middleware.py
+++ b/rbac/rbac/middleware.py
@@ -386,18 +386,9 @@ class IdentityHeaderMiddleware(MiddlewareMixin):
             request (object): The request object
             response (object): The response object
         """
-        is_internal = False
-        if any([request.path.startswith(prefix) for prefix in settings.INTERNAL_API_PATH_PREFIXES]):
-            # This request is for a private API endpoint
-            is_internal = True
-            IdentityHeaderMiddleware.log_request(request, response, is_internal)
-            return response
-
-        behalf = "principal"
-        is_system = False
-
-        if is_system:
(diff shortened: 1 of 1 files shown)


rbac/rbac/middleware.py
```diff
@@ -386,18 +386,17 @@ def process_response(self, request, response):  # pylint: disable=no-self-use
...
         is_system = False
 
-        if is_system:
-            behalf = "system"
+        if hasattr(request, "user") and request.user:
+            username = request.user.username
+            if username:
+                is_system = request.user.system
+            else:
+                is_system = False
```
petracihalova, line 397: LGTM, just nitpick ... I think the else is not needed here, because the False value is set few lines above

petracihalova (approved):

PR #1418, opened 2025-01-03
[RHCLOUD-27273] Enforce correct value/operation pairings

mjholder:
BLOCKED BY: https://github.com/RedHatInsights/insights-rbac/pull/1448

## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-27273

The what, why and how.
In order to keep JSON parsing working we need to make sure the value types and operations provided by resourceDefinitions are compatible.
Specifically, if a String is provided the operation needs to be `equals`. Otherwise, if a List is provided, the operation needs to be `in`.
In order to achieve this I added some extra checks in the ResourceDefinition Serializer to make sure the pairings are correct.
## Local Testing

I added a couple tests to the unit test suite. If you want to test locally you can try to create a role with either of the following access definitions:
```
{
    "permission": "someApp:*:*",
    "resourceDefinitions": [
        {"attributeFilter": {"key": "keyA.id", "operation": "equals", "value": ["value1", "value2"]}}
    ],
}
```
OR
```
{
    "permission": "someApp:*:*",
    "resourceDefinitions": [
        {"attributeFilter": {"key": "keyA.id", "operation": "in", "value": "valueA"}}
    ],
}
```
## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


vbelchio:
@mjholder do we need to make changes in the OpenAPI specs?

mjholder:
> @mjholder do we need to make changes in the OpenAPI specs?

That's a good question. We do have the types that are acceptable listed but not the fact that a string expects `equals` and a list expects `in`.
I'm looking to see if there is a way to achieve this in OpenAPI 3

mjholder:
> > @mjholder do we need to make changes in the OpenAPI specs?
> 
> That's a good question. We do have the types that are acceptable listed but not the fact that a string expects `equals` and a list expects `in`. I'm looking to see if there is a way to achieve this in OpenAPI 3

It looks like if we move to OpenAPI 3.1 we can use conditional blocks for schema validation but not in 3.0.0.
This if from the docs on what's added in 3.1: https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#page-42

rbac/management/role/serializer.py
```diff
@@ -49,6 +49,18 @@ def validate_attributeFilter(self, value):
...
+            if type(values) is str and op == "in":
+                key = "format"
+                message = "attributeFilter operation 'in' expects a List value"
+                error = {key: [_(message)]}
+                raise serializers.ValidationError(error)
+            elif type(values) is list and op == "equals":
+                key = "format"
+                message = "attributeFilter operation 'equals' expects a String value"
+                error = {key: [_(message)]}
+                raise serializers.ValidationError(error)
```
astrozzc, line 63: LGTM!
Nit: Instead of raising this in several places, can we use `if error, raise serializers.ValidationError(error)` after all these checks?

astrozzc (approved):

rbac/management/role/serializer.py
```diff
@@ -49,6 +49,18 @@ def validate_attributeFilter(self, value):
             message = f"attributeFilter operation must be one of {ALLOWED_OPERATIONS}"
             error = {key: [_(message)]}
             raise serializers.ValidationError(error)
+        else:
+            values = value.get("value")
+            if type(values) is str and op == "in":
```
vkrizan, line 54: These checks are insufficient, as these check only for string. If for example `int` was supplied, it would fall through. Why not check the expected type for each operator?

Be aware, that previously a documentation mentioned a string of comma separated values. Have you checked current state of the database? If users would have already stored badly formatted data, they could hit these errors.

I suggest to prefer `isinstance` over a strong type check. See [Duck Typing](https://docs.python.org/3/glossary.html#term-duck-typing).

vkrizan (changes requested):

petracihalova:
fixed typo in the attribute filter validator and did small refactoring there, then I updated few tests, mostly because now with the attribute filter validation we cannot create invalid data even in tests via API request so we have to create the directly in the database, this PR should be squashed before the merge, please @astrozzc can you take a look?

astrozzc (approved):
LGTM

PR #1441, opened 2025-01-16
[RHCLOUD-37152] Add locks on Roles, CAR and groups  in migrator

lpichler:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-37152

This PR add locks on Roles, CAR and groups in migrator and it will solve concurrency control during migration and dual writes.

## Local Testing
..

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/rbac/migration_tool/migrate.py
+++ b/rbac/migration_tool/migrate.py
@@ -40,24 +40,24 @@ logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
 def migrate_groups_for_tenant(tenant: Tenant, replicator: RelationReplicator):
     """Generate user relationships and system role assignments for groups in a tenant."""
-    groups = tenant.group_set.all()
+    groups = tenant.group_set.values("pk")
     for group in groups:
-        principals: list[Principal] = []
-        system_roles: list[Role] = []
-        if not group.platform_default:
-            principals = group.principals.all()
-        if group.system is False and group.admin_default is False:
-            system_roles = group.roles().public_tenant_only()
-        if any(True for _ in system_roles) or any(True for _ in principals):
-            # The migrator deals with concurrency control.
-            # We need an atomic block because the select_for_update is used in the dual write handler,
-            # and the group must be locked to add principals to the groups.
-            # NOTE: The lock on the group is not necessary when adding system roles to the group,
-            # as the binding mappings are locked during this process to ensure concurrency control.
-            # Start of transaction for group operations
-            with transaction.atomic():
-                # Lock group
-                Group.objects.select_for_update().get(pk=group.pk)
+        # The migrator deals with concurrency control.
+        # We need an atomic block because the select_for_update is used in the dual write handler,
+        # and the group must be locked to add principals to the groups.
+        # NOTE: The lock on the group is not necessary when adding system roles to the group,
+        # as the binding mappings are locked during this process to ensure concurrency control.
+        # Start of transaction for group operations
+        with transaction.atomic():
+            # Requery the group with a lock
+            group = Group.objects.select_for_update().get(pk=group["pk"])
+            principals: list[Principal] = []
+            system_roles: list[Role] = []
+            if not group.platform_default:
+                principals = group.principals.all()
+            if group.system is False and group.admin_default is False:
+                system_roles = group.roles().public_tenant_only()
+            if any(True for _ in system_roles) or any(True for _ in principals):
                 dual_write_handler = RelationApiDualWriteGroupHandler(
                     group, ReplicationEventType.MIGRATE_TENANT_GROUPS, replicator=replicator
                 )
@@ -68,7 +68,7 @@ def migrate_groups_for_tenant(tenant: Tenant, replicator: RelationReplicator):
                 # dual_write_handler
                 dual_write_handler.generate_relations_to_add_roles(system_roles)
                 dual_write_handler.replicate()
-            # End of transaction for group operations, locks are released
+        # End of transaction for group operations, locks are released
 
 
 def migrate_roles_for_tenant(tenant, exclude_apps, replicator):
--- a/rbac/migration_tool/migrate.py
+++ b/rbac/migration_tool/migrate.py
@@ -16,86 +16,26 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.
 """
 
 import logging
-from typing import Iterable
 
 from django.conf import settings
 from django.db import transaction
-from kessel.relations.v1beta1 import common_pb2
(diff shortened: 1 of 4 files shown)


rbac/migration_tool/migrate.py
```diff
@@ -108,46 +49,44 @@ def migrate_groups_for_tenant(tenant: Tenant, replicator: RelationReplicator):
...
-            # but we require an atomic block due to use of select_for_update in the dual write handler.
+            # The migrator deals with concurrency control.
+            # We need an atomic block because the select_for_update is used in the dual write handler,
+            # and the group must be locked to add principals to the groups.
+            # NOTE: The lock on the group is not necessary when adding system roles to the group,
+            # as the binding mappings are locked during this process to ensure concurrency control.
+            # Start of transaction for group operations
             with transaction.atomic():
+                # Lock group
+                Group.objects.select_for_update().get(pk=group.pk)
```
alechenninger, line 60: I think we may want to use the group state as of this query, rather than relying on the group state prior to this. I'm wondering for example about if between the time the principals are queried, and the group is locked, principals are removed. That removal replicates first, and then the add happens which adds them back.

It's also overall maybe just easier to reason about.

To do that it is probably as simple as moving the `with transaction.atomic()` to the top of and just inside the `for` loop. It will potentially lock more groups, because it will end up locking groups that we don't have anything to replicate for, but that's probably fine.

e.g.

```
    for group in groups:
        with transaction.atomic():
            principals: list[Principal] = []
            system_roles: list[Role] = []
            if not group.platform_default:
            # ... etc
```

alechenninger: Actually sorry that example is misleading I think. I was multitasking.


```
    for group in groups:  
        with transaction.atomic():
            # Requery the group with a lock
            group = Group.objects.select_for_update().get(pk=group.pk)
            principals: list[Principal] = []  
            system_roles: list[Role] = []  
            if not group.platform_default:  
            # ... etc
```

alechenninger: (you could optimize this a little bit by getting just the group pk's in the prior `tenant.group_set` query)

alechenninger: Also, I think we may need to modify the group queryset in the view to lock the group in any principal modification, not just adds.

The reason is because in this case, the migrator assumes an "add principals" operation, but that's not actually happening. So normally when add and remove interleave, it should be fine, because they both also are modifying the database. So they may interleave, but regardless the outbox is always consistent with the database, so if a principal is there (or not), the same will be reflected in relations.

In this case, the migrator is not actually modifying the group-principal data, it's just producing an outbox message. If it also did `group.principals.add(...)` then the outbox and relations would be consistent even if this interleaved with a removal, however that is probably inappropriate because it would "undo" the removal operation. (In the case of normal dual write, _another user "undid" the operation_ so it's not as big of a deal if they interleave.)

The upside is making the lock more general makes the code a little easier to reason about I think. The downside is it can hurt performance, but I think that is not really a big concern here, since it is really unlikely that there is contention over these specific records.

rbac/migration_tool/migrate.py
```diff
@@ -108,46 +49,44 @@ def migrate_groups_for_tenant(tenant: Tenant, replicator: RelationReplicator):
...
-                partition_key=PartitionKey.byEnvironment(),
-                add=tuples,
+        with transaction.atomic():
+            role = Role.objects.select_for_update().get(pk=role.pk)
+            dual_write_handler = RelationApiDualWriteHandler(
+                role, ReplicationEventType.MIGRATE_CUSTOM_ROLE, replicator
             )
-        )
+
+            dual_write_handler.replicate_new_or_updated_role(role)
```
alechenninger, line 89: Here I think we might need two adjustments:

- We need `dual_write_handler.prepare_for_update()` before the replicate method
- We might have a similar problem to above group principal removal with role deletion here. Normally, dual write shouldn't need to lock the role on deletion, because foreign key constraints on modifications to the role will prevent inconsistency. But here is another case where the migrator is not adding the data to the DB at the same time. So I think we could fix this one by _either_ re-running `.save()` on a role serializer here, so that access FKs serialize concurrent removal, OR we have to lock the role in the role view queryset on role deletion also.

rbac/migration_tool/migrate.py
lpichler: I added those suggestions.

rbac/migration_tool/migrate.py
lpichler: - I added `prepare_for_update`(I though that removal of current relations is not needed for role )
- I think lock on role  in delete action is already here

rbac/migration_tool/migrate.py
```diff
@@ -171,29 +111,40 @@ def migrate_data_for_tenant(tenant: Tenant, exclude_apps: list, replicator: Rela
...
 
 def migrate_data(
     exclude_apps: list = [], orgs: list = [], write_relationships: str = "False", skip_roles: bool = False
 ):
     """Migrate all data for all tenants."""
-    # Only run this in maintanence mode or
-    # if we don't write relationships (testing out the migration and clean up the created bindingmappings)
-    if not settings.READ_ONLY_API_MODE and write_relationships != "False":
+    # Run this if we don't write relationships (testing out the migration and clean up the created bindingmappings)
+    if write_relationships != "False":
```
astrozzc, line 147: If we get rid of the settings.READ_ONLY_API_MODE condition, we can never set write_relationships to True?

rbac/migration_tool/migrate.py
```diff
@@ -16,140 +16,80 @@
...
 
 def migrate_roles_for_tenant(tenant, exclude_apps, replicator):
     """Migrate all roles for a given tenant."""
-    default_workspace = Workspace.objects.get(type=Workspace.Types.DEFAULT, tenant=tenant)
-
     roles = tenant.role_set.all()
     if exclude_apps:
         roles = roles.exclude(access__permission__application__in=exclude_apps)
 
     for role in roles:
```
alechenninger, line 79: If you want I guess we could also just get the Role PKs here but not a big deal.

rbac/migration_tool/migrate.py
lpichler: yes, good point, I think we don't need this condition. I am removing it .

lpichler:
I am going to merge as I added last suggestions and I got LGTM.

PR #1454, opened 2025-01-22
[RHCLOUD-37156] Make reset import tenants admin endpoint use ready=false query instead of looking up all related objects

petracihalova:
## Link(s) to Jira
- [RHCLOUD-37156](https://issues.redhat.com/browse/RHCLOUD-37156)

we have [internal endpoint](https://github.com/RedHatInsights/insights-rbac/blob/aff997f853d6eefb74be0ccbce5e7d83969414a3/rbac/internal/urls.py#L86) for reseting imported tenants [here](https://github.com/RedHatInsights/insights-rbac/blob/8262ac5ff41933432db9491e01fe6c4daa51e1a7/rbac/internal/views.py#L629)

and for tenants identification we are looking up all related objects

this PR introduces the simplification of the query = instead of checking all related objects we can use `ready=false` flag

but because it could be useful to keep the old behavior I added new query parameter `only_ready_false_flag` (with default value `true`)
- `only_ready_false_flag=true` => use query with `ready=false` flag and get all tenants with `ready=false` flag
- `only_ready_false_flag=false` => use query that is looking up all related objects and get all tenants without related objects

in the perfect world (tenant with `ready=false` does not have related objects) both queries should return same data/result

## Local Testing
you can test it locally with internal endpoint
```
GET / DELETE http://localhost:8000/_private/api/utils/reset_imported_tenants/
GET / DELETE http://localhost:8000/_private/api/utils/reset_imported_tenants/?only_ready_false_flag=true
GET / DELETE http://localhost:8000/_private/api/utils/reset_imported_tenants/?only_ready_false_flag=false
```

## TODO
- [x] update open api spec for internal endpoints
- [x] add ACs in [RHCLOUD-37156](https://issues.redhat.com/browse/RHCLOUD-37156)
- [X] unit tests


rbac/internal/views.py
```diff
@@ -637,44 +637,51 @@ def reset_imported_tenants(request: HttpRequest) -> HttpResponse:
     # Request should accept a query parameter to exclude certain tenants so we can exclude the ~129
     excluded = request.GET.getlist("exclude_id", [])
 
+    # The default query created with "ready=false" flag otherwise is used query that checks that tenant
+    # does not have records in all tables.
+    query_with_ready_flag = request.GET.get("query_with_ready_flag", "true").strip().lower() == "true"
```
astrozzc, line 642: maybe we should call it query_with_only_ready_flag since we are excluding other query conditions

petracihalova: yeah the naming here is the most difficult part ... so the new query is only with `ready=false` flag and the original query is wihout `ready` flag but with all the conditions that search for related objects

petracihalova: maybe we should call it `only_ready_false_flag` or something like that 🤔

astrozzc: sounds good 👍

petracihalova: renamed

astrozzc (approved):

petracihalova:
rebased and conflicts solved

PR #1487, opened 2025-02-03
RHCLOUD-37662: Make migrator idempotent

alechenninger:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-37662

Tests checklist:
Groups:
- [x] Given we add the same role to group dual write handler multiple times, we should see the group bound multiple times
Note: This is testing existing expected behavior to make sure that is clear 
Test:` test_adding_same_role_again_and_unassign_it_once`
- [x] Given we add the same role to group dual write handler multiple times, and then remove the role, we should still see the group bound and no tuple should be returned
Note: This is testing existing expected behavior to make sure that is clear
Test:` test_adding_same_role_again_and_unassign_it_once`
- [x] Given we add the same role to group dual write handler multiple times, and then reset the role, we should no longer see group bound, and the tuple should be returned for that group relation
Test: `test_reset_when_role_added_multiple_times` 

- [x] Given we add the same role to group dual write handler multiples times, and reset the role multiple times, we should no longer see group bound, and the tuple should be returned for that group relation
Test: `test_reset_called_multiple_times_when_role_added_multiple_times`

- [x] Same scenarios as above, but for cross account request dual write handler

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


tests/management/role/test_dual_write.py
```diff
@@ -444,6 +444,49 @@ def test_adding_same_role_again_and_unassign_it_once(self):
         )
         self.assertEquals(len(tuples), 0)
 
+    def test_group_not_bound_after_reset_when_role_added_multiple_times(self):
```
lpichler, line 447: >  From jira Given we add the same role to group dual write handler multiple times, and then reset the role, we should no longer see group bound, and the tuple should be returned for that group relation

Is correct that we should not see group bound ? 
when reset group-roles assignments and there is group-roles assignment, should be group still in mapping ?

astrozzc: Yes, it should be. Reset makes sure it only gets added once.

rbac/management/role/model.py
```diff
@@ -174,10 +174,9 @@ def unassign_group(self, group_uuid) -> Optional[Relationship]:
         """
         relationship = None
         while True:
-            previous_relationship = self.pop_group_from_bindings(group_uuid)
-            if previous_relationship is None:
+            relationship = self.pop_group_from_bindings(group_uuid)
+            if relationship is not None:
                 break
-            relationship = previous_relationship
         return relationship
```
alechenninger, line 180: I don't think this is right? We want to remove all of them, but this will only remove one, because it stops once the relationship is not None. If there is more than one, does this work?

lpichler (approved):

rbac/api/cross_access/view.py
```diff
@@ -241,7 +241,7 @@ def update_status(self, car, status):
             self._with_dual_write_handler(
                 car,
                 ReplicationEventType.APPROVE_CROSS_ACCOUNT_REQUEST,
-                lambda dual_write_handler, cross_account_roles: dual_write_handler.generate_relations_to_add_roles(
+                lambda dual_write_handler, cross_account_roles: dual_write_handler.generate_relations_reset_roles(
```
lpichler, line 244: is this intent to switch from `generate_relations_to_add_roles` to `generate_relations_reset_roles` ?

astrozzc: Good question, after thinking about it, should use the generate_relations_to_add_roles to reduce the interruption. Updated

PR #1505, opened 2025-02-11
[RHCLOUD-37288] move the logic for service account request preparation and response processing in own method

petracihalova:
as part of preparations for [RHCLOUD-37288](https://issues.redhat.com/browse/RHCLOUD-37288)
I created own static method for service account request processing


rbac/management/principal/view.py
```diff
@@ -243,3 +210,39 @@ def users_from_proxy(self, user, query_params, options, limit, offset):
...
+        token_validator = ITSSOTokenValidator()
+        user.bearer_token = token_validator.validate_token(
+            request=request, additional_scopes_to_validate=set[ScopeClaims]([ScopeClaims.SERVICE_ACCOUNTS_CLAIM])
+        )
+
+        try:
+            it_service = ITService()
+            service_accounts, sa_count = it_service.get_service_accounts(user=user, options=options)
+        except (requests.exceptions.ConnectionError, UnexpectedStatusCodeFromITError):
+            return Response(
```
astrozzc, line 236: If there is an error, I think this response will be stored in the resp, where it is called, and process normally instead of returning the response. Maybe we can try catch outside of this method service_accounts_from_it_service

petracihalova: yeah that is a good point ... it looks like in this case we should return dict with error message and then [here](https://github.com/RedHatInsights/insights-rbac/blob/a8c204fae160251baf78648b1dee0abba6c56824/rbac/management/principal/view.py#L176) is decided if the response contains 200 or something else ... it is solved the same way for user based principals when BOP is called ... I will add a test for this case

PR #1510, opened 2025-02-13
[RHCLOUD-37871] add the paginator in the principals view to handle limit and offset

petracihalova:
## Link(s) to Jira
- [RHCLOUD-37871](https://issues.redhat.com/browse/RHCLOUD-37871)

replace the current `limit` and `offset` handling with our custom paginator `StandardResultsSetPagination`

because the pagination is done on the BOP/IT service side, we dont use the paginator in the principals view, 
but the `limit` and `offset` still can be handled by the paginator as we do for other views

so now any invalid value is processed by paginator and if needed then default values are returned

**the initial motivation** for this was that with current implementation when you try to list all principals with limit = 0, then all principals are returned and links are incorrect

```GET /api/rbac/v1/principals/?limit=0```

now with paginator all invalid values are covered in same way as we do in other endpoints
for example
```
GET /api/rbac/v1/principals/?limit=0
GET /api/rbac/v1/principals/?offset=-10
GET /api/rbac/v1/principals/?limit=foo
```

## Local Testing
send request with invalid limit and offset and list the principals
```
GET /api/rbac/v1/principals/?limit=<value>&offset=<value>
GET /api/rbac/v1/principals/?type=service-account&limit=<value>&offset=<value>
```

if valid value is provided, then the same value is used in the query
if invalid value is provided, then the default values for limit and offset are used

default value for limit = 10
default value for offset = 0

valid value for the limit is non-negative integer (limit > 0) 
valid value for the offset is positive integer (offset >= 0)


Ellen-Yi-Dong (approved):
LGTM!! 
Tested the code locally and worked for me as mentioned in description!!

lpichler (approved):

vbelchio:
QE: Passed (regression)

PR #1522, opened 2025-02-19
RHCLOUD-37616 - Service account principals filtering and sorting

EvanCasey13:
## Link(s) to Jira
https://issues.redhat.com/browse/RHCLOUD-37616

Implemented filtering for service accounts on name, owner and description with test cases for filtering. Implemented sorting of service accounts by Service Account Name, Description, Client ID, Owner & Time created both ascending and descending.

## Local Testing
Running this curl commands replacing the filtering parameter with the valid fields of name, owner or description or with order_by using name, description, clientId, owner or time_created.

`curl -v "http://localhost:8000/api/rbac/v1/principals/?type=service-account&name=e6921ca0-9cd3-013c-1dc6-6aa2427b506a-name" -H "accept: application/json" | json_pp`

order_by can be reversed using '-' so in this case &order_by=-name
`curl -v "http://localhost:8000/api/rbac/v1/principals/?type=service-account&order_by=name" -H "accept: application/json" | json_pp`

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


astrozzc (approved):

petracihalova:
🤔 in relation with the open api spec we need to consider that this filtering and sorting is related only with the service account based principals but the endpoint `/principals/` is also used for the user based principals so the query parameter description could be confusing now

rbac/management/principal/view.py
```diff
@@ -131,7 +131,11 @@ def get(self, request):
             query_params, PRINCIPAL_TYPE_KEY, VALID_PRINCIPAL_TYPE_VALUE, default_value=USER_KEY, required=False
         )
         options["principal_type"] = principal_type
-
+        # Optional query parameters for service account specific filtering
+        params = ["name", "description", "clientId", "owner", "time_created", "order_by"]
```
petracihalova, line 135: 🤔 I'm sorry if I missed something but not sure about "clientId", "owner" and "time_created"? these are values for the "order_by" parameter right? and according to the comment, there should be filtering query params so it looks like the "order_by" shouldn't be there either 🤔

rbac/management/principal/it_service.py
```diff
@@ -303,11 +303,34 @@ def get_service_accounts(self, user: User, options: dict[str, Any] = {}) -> Tupl
...
+        description = options.get("description")
+        filtered_service_accounts = []
+        sa_query_passed = name or owner or description
 
         count = len(service_accounts)
         # flake8 ignore E203 = Whitespace before ':' -> false positive https://github.com/PyCQA/pycodestyle/issues/373
         service_accounts = service_accounts[offset : offset + limit]  # type: ignore # noqa: E203
-
+        # If any one service account filter parameter is provided extract & return the specific service accounts
+        if sa_query_passed:
```
petracihalova, line 317: I did few tests and it looks like the pagination is not working correctly, my local tenant contains 4 service accounts, 3 of them  belongs to ownerA and the last one to ownerB, I tried to filter by `?owner=ownerA` and I received 3 SA in the response as expected, then I set `limit=2` to check pagination and my response contains only 1 SA (2 expected) and the count under the meta shows 1 (3 is expected), I can provide more details if needed, I believe it is because the limit and offset are set before the filtering is done, I am happy to provide more details if needed

petracihalova (changes requested):

rbac/management/principal/view.py
EvanCasey13: The comment above the code is wrong, that is only to check what was provided in query params and assign to options to be used in IT service class. Just prevents assigning an option if it is not actually provided in the query parameters.

tests/management/principal/test_view.py
```diff
@@ -798,6 +798,1197 @@ def test_read_principal_users(self, mock_request):
...
+                    "owner": "jsmith",
+                    "username": "service_account-" + uuid,
+                    "time_created": 1706784741,
+                    "type": "service-account",
+                }
+            )
+
+        mock_request.return_value = mocked_values
+
+        url = f"{reverse('v1_management:principals')}?type=service-account&name=service_account_name_b6636c60"
```
petracihalova, line 836: in this test all service accounts that are returned as mocked records have the same name so then when you filter by name = `service_account_name_b6636c60` then all records are returned in the response, I would prefer test when only one (or subset) of service accounts contains the name we use in the filter so then we can check that the service accounts were actually filtered ... the same for some other tests below

rbac/management/principal/it_service.py
```diff
@@ -303,8 +303,34 @@ def get_service_accounts(self, user: User, options: dict[str, Any] = {}) -> Tupl
...
         count = len(service_accounts)
+        # If any one service account filter parameter is provided extract & return the specific service accounts
+        if sa_query_passed:
+            filtered_service_accounts = []
+            for sa in service_accounts:
+                sa_description = str(sa.get("description"))
+                if (
+                    (not name or sa.get("name") == name)
+                    and (not owner or sa.get("owner") == owner)
+                    and (not description or description in sa_description)
```
petracihalova, line 321: I tried to use the filtering by description and I noticed that the filtering is case sensitive 🤔 and I believe we should have this case insensitive ... what you think @astrozzc

tests/management/principal/test_view.py
```diff
@@ -798,6 +798,1226 @@ def test_read_principal_users(self, mock_request):
 
         cross_account_principal.delete()
 
+    @override_settings(IT_BYPASS_TOKEN_VALIDATION=True)
```
petracihalova, line 801: good job with tests 👍

rbac/management/principal/it_service.py
astrozzc: Sounds good, let's use case insensitive then

astrozzc:
> @EvanCasey13 I did few tests and the new filtering is working well, I have question what is expected if user wants to use filtering and ordering in one query? now it looks like if you send request with both options, then only filtering is applied
> 
> ```
> GET /api/rbac/v1/principals/?type=service-account&order_by=description&description=<value>
> ```
Good point, can we pull that else block out, seems it can be applied to either sa_query_passed exists or not

petracihalova (approved):
good job 👍

petracihalova:
PR tested with these queries:
Basic queries:
```
- GET /principals/?type=service-account
- GET /principals/?type=all
- GET  /principals/?type=user
- GET /principals/?type=foo
```

Filtering:
```
- GET /principals/?type=service-account&name=<value> (exact match, not case sensitive)
- GET /principals/?type=service-account&owner=<value> (exact match, not case sensitive)
- GET /principals/?type=service-account&description=<value> (partial match, not case sensitive)
```

Sorting:
```
- GET /principals/?type=service-account&order_by=name 
- GET /principals/?type=service-account&order_by=-name 
- GET /principals/?type=service-account&order_by=owner 
- GET /principals/?type=service-account&order_by=-owner 
- GET /principals/?type=service-account&order_by=description 
- GET /principals/?type=service-account&order_by=-description
- GET /principals/?type=service-account&order_by=clientId 
- GET /principals/?type=service-account&order_by=-clientId
- GET /principals/?type=service-account&order_by=time_created 
- GET /principals/?type=service-account&order_by=-time_created
```

Filtering + Sorting:
```
- GET /principals/?type=service-account&order_by=time_created&name=<value>
- GET /principals/?type=service-account&order_by=-time_created&name=<value>
- GET /principals/?type=service-account&order_by=name&owner=<value>
- GET /principals/?type=service-account&order_by=-name&owner=<value>
```

PR #1534, opened 2025-02-25
query param validation: use the default value if the query param is provided without a value

petracihalova:
in utils we have a function for query params validation `validate_and_get_key()`
one case can be that the query param is provided without a value, for example `GET /principals/?type=`
= the value is `""` (empty string) and the default value is not used

I think in that case we should return the default value

and I noticed we don't have tests for this function so I wrote few 😉


lpichler (approved):
Thanks for tests!

PR #1554, opened 2025-03-10
principal_type query param refactoring

petracihalova:
- set the constants for the `principal_type` query param 
- set the default value (`user`) for the `principal_type` query param


astrozzc (approved):
LGTM! I wonder if we could use the Types of Principal model instead of defining new ones here

petracihalova:
> LGTM! I wonder if we could use the Types of Principal model instead of defining new ones here

I like the idea 👍

lpichler (approved):

PR #1512, opened 2025-02-14
[RHCLOUD-37288] add option to list both principal types in one query to the /principals/ endpoint

petracihalova:
## Link(s) to Jira
- [RHCLOUD-37288](https://issues.redhat.com/browse/RHCLOUD-37288)
- [RHCLOUD-38309](https://issues.redhat.com/browse/RHCLOUD-38309)

## Description of Intent of Change(s)
the `/principals/` endpoint allows to list the user based principals with `type=user` query param (default option)
you can list service account based principals with `type=service-account` query param
this PR adds third option `type=all` to list both principal types in one query:
- first we list service accounts if present
- then we list user based principals
- the filtering with `usernames` query param is working too (you can combine usernames of both principal types)
- the access token is needed for the query (same as for `type=service-account`)

example of the response:
```
{
    "meta": {
        "count": 86,
        "limit": 2,
        "offset": 2
    },
    "links": {
        "first": "/api/rbac/v1/principals/?limit=2&offset=0",
        "next": "/api/rbac/v1/principals/?limit=2&offset=4",
        "previous": "/api/rbac/v1/principals/?limit=2&offset=0",
        "last": "/api/rbac/v1/principals/?limit=2&offset=84"
    },
    "data": {
        "serviceAccounts": [
            {
                "type": "service-account",
                "username": "service-account-de4c127a-f818-4ead-b5b9-1fef9b597513"
                ...
            }
        ],
        "users": [
            {
                "username": "my_user_principal",
                ...
            }
        ]
    }
}
```

## Local Testing
use tenant with few user based principals and service accounts and test queries with(out) the `type` query param
```
GET /api/rbac/v1/principals/
GET /api/rbac/v1/principals/?type=user
GET /api/rbac/v1/principals/?type=service-account
GET /api/rbac/v1/principals/?type=all
```

for example if `GET /api/rbac/v1/principals/` returns 10 user based principals and `GET /api/rbac/v1/principals/?type=service-account` returns 3 service accounts, then `GET /api/rbac/v1/principals/?type=all` should return 13 principals (both types)

try combinations of limit and offset to check that pagination is working as expected

for example if your tenant contains 3 SA (service accounts) and 4 U (user based principals) then with `limit=2` and `offset=0` you get this pagination:
```
page 1 -> 2 SA
page 2 -> 1 SA + 1 U
page 3 -> 2 U
page 4 -> 1 U
```

## Checklist
- [x] open api spec updated
- [x] unit tests
- [x] added the ACs in the ticket

PR #1567, opened 2025-03-12
[RHCLOUD-38310] add option to list both principal types in a group in one query

petracihalova:
## Link(s) to Jira
- [RHCLOUD-38310](https://issues.redhat.com/browse/RHCLOUD-38310)
- [RHCLOUD-37288](https://issues.redhat.com/browse/RHCLOUD-37288)

related with https://github.com/RedHatInsights/insights-rbac/pull/1512

the `/groups/<uuid>/principals/` endpoint allows to list the user based principals with `principal_type=user` query param (default option) and service account based principals with `principal_type=service-account` query param

this PR adds third option `principal_type=all` to list both principal types in one query:
- first we list service accounts if present
- then we list user based principals
- the access token is needed for the query (same as for `principal_type=service-account`)

example of the response:
```
{
    "meta": {
        "count": 86,
        "limit": 2,
        "offset": 2
    },
    "links": {
        "first": "/api/rbac/v1/principals/?limit=2&offset=0",
        "next": "/api/rbac/v1/principals/?limit=2&offset=4",
        "previous": "/api/rbac/v1/principals/?limit=2&offset=0",
        "last": "/api/rbac/v1/principals/?limit=2&offset=84"
    },
    "data": {
        "serviceAccounts": [
            {
                "type": "service-account",
                "username": "service-account-de4c127a-f818-4ead-b5b9-1fef9b597513"
                ...
            }
        ],
        "users": [
            {
                "username": "my_user_principal",
                ...
            }
        ]
    }
}
```

## Local Testing
create a group with few user based principals and service accounts and test queries with(out) the `principal_type` query param
```
GET /api/rbac/v1/groups/<uuid>/principals/
GET /api/rbac/v1/groups/<uuid>/principals/?principal_type=user
GET /api/rbac/v1/groups/<uuid>/principals/?principal_type=service-account
GET /api/rbac/v1/groups/<uuid>/principals/?principal_type=all
```

for example if `GET /api/rbac/v1/groups/<uuid>/principals/` returns 10 user based principals 
and `GET /api/rbac/v1/groups/<uuid>/principals/?principal_type=service-account` returns 3 service accounts, 
then `GET /api/rbac/v1/groups/<uuid>/principals/?type=all` should return 13 principals (both types)

try combinations of limit and offset to check that pagination is working as expected

for example if your tenant contains 3 SA (service accounts) and 4 U (user based principals) then with `limit=2` and `offset=0` you get this pagination:
```
page 1 -> 2 SA
page 2 -> 1 SA + 1 U
page 3 -> 2 U
page 4 -> 1 U
```

## Checklist
- [x] open api spec updated
- [x] unit tests
- [x] ACs in the ticket


astrozzc (approved):
🤔 If they want all types, seems no sense to paginate them together? I wonder what's done for the pagination in this case, is the total for each of the type or for the total of the type, if the offset for both of them or each of them.

petracihalova:
> 🤔 If they want all types, seems no sense to paginate them together? I wonder what's done for the pagination in this case, is the total for each of the type or for the total of the type, if the offset for both of them or each of them.

first we return service accounts and then user based principals and because we have this condition, we can calculate how many of both types should be return in relation with limit and offset ... one example in the PR description, let me know if I can provide more details

astrozzc:
> > 🤔 If they want all types, seems no sense to paginate them together? I wonder what's done for the pagination in this case, is the total for each of the type or for the total of the type, if the offset for both of them or each of them.
> 
> first we return service accounts and then user based principals and because we have this condition, we can calculate how many of both types should be return in relation with limit and offset ... one example in the PR description, let me know if I can provide more details

Got it, yea, makes sense

PR #1566, opened 2025-03-12
[RHCLOUD-37831] Support internal api call through psk

astrozzc:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-37831

## Description of Intent of Change(s)
Create another set of apis not available to public
Use pre-shared key to authenticate
Give the key to HBI share through vault, [in stage, we already have one]
Replicate necessary relationships, e.g., default->ungrouped
Input: org_id

## Local Testing
Unit tests

## Checklist
- [x] if API spec changes are required, is the spec updated?
- [ ] are there any pre/post merge actions required? if so, document here.
- [x] are theses changes covered by unit tests?
- [ ] if warranted, are documentation changes accounted for?
- [ ] does this require migration changes?
  - [ ] if yes, are they backwards compatible?
- [ ] is there known, direct impact to dependent teams/components?
  - [ ] if yes, how will this be handled?

## Secure Coding Practices Checklist Link
- https://github.com/RedHatInsights/secure-coding-checklist

## Secure Coding Practices Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [ ] General Coding Practices

PR #1569, opened 2025-03-13
RHCLOUD-37820:  user lookup endpoint

dagbay-rh:
## Link(s) to Jira
- main: https://issues.redhat.com/browse/RHCLOUD-37820
- turnpike dependency: https://issues.redhat.com/browse/RHCLOUD-38971

Added a new internal/private endpoint that will query RBAC for a users groups, roles, and permissions based on their username or email
```
GET /_private/api/utils/user_lookup/?username=foo&email=foo@redhat.com
```

## Local Testing
1. I set the following env vars in `.env` to test against stage bop:
```
DATABASE_NAME=postgres
DATABASE_HOST=localhost
DATABASE_PORT=15432
DATABASE_USER=postgres
DATABASE_PASSWORD=postgres
API_PATH_PREFIX=/api/rbac
DEVELOPMENT=False
V2_BOOTSTRAP_TENANT=True
DJANGO_READ_DOT_ENV_FILE=True
DJANGO_LOG_LEVEL=INFO
RBAC_LOG_LEVEL=DEBUG
PRINCIPAL_PROXY_SERVICE_PROTOCOL="https"
PRINCIPAL_PROXY_SERVICE_HOST="backoffice-proxy.apps.ext.spoke.preprod.us-west-2.aws.paas.redhat.com"
PRINCIPAL_PROXY_SERVICE_PORT=""
PRINCIPAL_PROXY_SERVICE_PATH=""
PRINCIPAL_PROXY_USER_ENV="stage"
PRINCIPAL_PROXY_CLIENT_ID="insights-rbac"
PRINCIPAL_PROXY_API_TOKEN="<token from vault>"
PRINCIPAL_PROXY_SERVICE_SSL_VERIFY="False"
```
2. activate your virtual env via `pipenv shell`
3. spin up RBAC db and service locally

#### Test user that exists in no custom groups:
1. simply hit the endpoint and query for a user that exists in stage:
```
curl --location 'http://localhost:8000/_private/api/utils/user_lookup/?username=dagbay%40redhat.com' 
--header 'x-rh-identity: <your identity header>'

{
    "username": "dagbay@redhat.com",
    "email_address": "dagbay@redhat.com",
    "groups": [
        {
            "name": "Default access",
            "description": "This group contains the roles that all users inherit by default. Adding or removing roles in this group will affect permissions for all users in your organization.",
            "uuid": "519dfedb-7a17-481e-bc62-90f1fd25ef57",
            "platform_default": true,
            "admin_default": false,
            "roles": [
                {
                    "name": "Approval Administrator Local Test",
                    "display name": "Approval Administrator Local Test",
                    "description": "An Approval administrator role that grants create, read, update and destroy permissions, just for testing locally",
                    "uuid": "abc4557b-3bda-4a34-8921-307f25da86d6",
                    "platform_default": true,
                    "admin_default": false,
                    "permissions": [
                        "approval | templates | read",
                        "approval | workflows | create",
                        "approval | workflows | read",
                        "approval | workflows | update",
                        "approval | workflows | delete",
                        "approval | requests | read",
                        "approval | actions | read",
                        "approval | actions | create"
                    ]
                },
                .........
            ]
        },
        {
            "name": "Default admin access",
            "description": "This group contains the roles that all org admin users inherit by default. Adding or removing roles in this group will affect permissions for all org admin users in your org.",
            "uuid": "ccfaed2f-1698-424d-9aa1-ee9f8c7eeb02",
            "platform_default": false,
            "admin_default": true,
            "roles":[.................]
        }
    ]
}
```

#### Test for a user that exists in more than just default groups
1. create a group:
```
curl --location 'http://localhost:8000/api/rbac/v1/groups/' \
--header 'x-rh-identity: <your identity header>' \
--header 'Content-Type: application/json' \
--data '{
    "name": "MyGroup"
}'
```
2. add an existing role to that group:
```
curl --location 'http://localhost:8000/api/rbac/v1/groups/0d4d552d-7114-4e82-a9bd-8ae20f2e9068/roles/' \
--header 'x-rh-identity: <your identity header>' \
--header 'Content-Type: application/json' \
--data '{
    "roles": [
        "<uuid of an existing role from the db>"
    ]
}'
```
3. add a principal to that group:
```
curl --location 'http://localhost:8000/api/rbac/v1/groups/<group id from previous step>/principals/' \
--header 'x-rh-identity: <your identity header>' \
--header 'Content-Type: application/json' \
--data-raw '{
    "principals": [
        {
            "username": "dagbay@redhat.com"
        }
    ]
}'
```
4. and then query for that user again to see the custom group:
```
curl --location 'http://localhost:8000/_private/api/utils/user_lookup/?username=dagbay%40redhat.com' 
--header 'x-rh-identity: <your identity header>'
{
    "username": "dagbay@redhat.com",
    "email_address": "dagbay@redhat.com",
    "groups": [
        {
            "name": "Default access",
            .............
        },
        {
            "name": "Default admin access",
            ..............
        },
        {
            "name": "MyGroup",
            "description": "",
            "uuid": "0d4d552d-7114-4e82-a9bd-8ae20f2e9068",
            "platform_default": false,
            "admin_default": false,
            "roles": [
                {
                    "name": "Cost OpenShift Viewer Local Test",
                    "display name": "Cost OpenShift Viewer Local Test",
                    "description": "A cost management role that grants read permissions on cost reports related to OpenShift sources, just for testing locally.",
                    "uuid": "df76a7ce-7745-4728-9c13-b771ebacf18a",
                    "platform_default": true,
                    "admin_default": false,
                    "permissions": [
                        "cost-management | openshift.cluster | *"
                    ]
                }
            ]
        }
    ]
}
```

## Checklist

pre merge: setup the turnpike route for this endpoint, ticket: https://issues.redhat.com/browse/RHCLOUD-38971

- [x] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [x] Database Security
- [ ] File Management
- [ ] Memory Management


rbac/api/serializers.py
```diff
@@ -56,6 +56,9 @@ def extract_header(request, header):
         Encoded(str): Base64 header
         Decoded(dict): Identity dictionary
     """
+    if header not in request.META:
```
dagbay-rh, line 59: unrelated, I just noticed the server blows up with a KeyError 500 if the identity header is not present in the request, so I wanted to catch that

rbac/internal/middleware.py
```diff
@@ -46,7 +46,7 @@ def process_request(self, request):
             return
         try:
             _, json_rh_auth = extract_header(request, self.header)
-        except (JSONDecodeError, binascii.Error):
+        except (JSONDecodeError, binascii.Error, KeyError):
```
dagbay-rh, line 49: https://github.com/RedHatInsights/insights-rbac/pull/1569/files#r1996289233

rbac/internal/utils.py
```diff
@@ -36,13 +36,20 @@ def build_internal_user(request, json_rh_auth):
     user = User()
     valid_identity_types = ["Associate", "X509"]
     try:
-        if not json_rh_auth["identity"]["type"] in valid_identity_types:
+        identity_type = json_rh_auth["identity"]["type"]
```
dagbay-rh, line 39: unrelated, I just wanted to add some logs around to help figure out why my request was failing

rbac/internal/views.py
```diff
@@ -685,12 +840,6 @@ def bootstrap_tenant(request):
     return HttpResponse(f"Bootstrapping tenants with org_ids {org_ids} were finished.", status=200)
 
 
-class SentryDiagnosticError(Exception):
```
dagbay-rh, line 688: moved this to `errors.py`

rbac/management/principal/proxy.py
```diff
@@ -282,7 +287,7 @@ def request_filtered_principals(self, principals, org_id=None, limit=None, offse
             return {"status_code": status.HTTP_200_OK, "data": []}
         filtered_principals_path = "/v1/users"
         params = self._create_params(limit, offset, options)
-        payload = {"users": principals, "include_permissions": False}
```
dagbay-rh, line 285: this was not actually honoring the `include_permissions` field in the request to bop, so I updated it to work as intended

tests/internal/test_views.py
```diff
@@ -54,30 +54,14 @@
...
-        "loggers": {
-            "management.relation_replicator.outbox_replicator": {
-                "level": "INFO",
-            },
-        },
-    },
-)
-class InternalViewsetTests(IdentityRequest):
-    """Test the internal viewset."""
+class BaseInternalViewsetTests(IdentityRequest):
```
dagbay-rh, line 57: I reworked this file a bit:

I abstracted out the setup & teardown in the existing `InternalViewsetTests` class to this new abstract base class `BaseInternalViewsetTests`

I then updated `InternalViewsetTests` to inherit from this base class, and added a new class for my tests `InternalViewsetUserLookupTests` which also inherits from this case class

This will allow us to run tests from this suite at a more granular level (ie I can specify a class which might cover a certain endpoint or feature alone), as well as help break up this massive file a bit. Ideally I think we could start to break out the individual classes into their own files, but I didn't want to rock the boat that much right now. 

Lmk what yall think. Open to suggestions if this is an unwarranted change :cowboy_hat_face:

rbac/internal/views.py
```diff
@@ -338,6 +341,155 @@ def get_org_admin(request, org_or_account):
...
+        principals = Principal.objects.filter(username__iexact=user["username"])
+    except Exception as err:
+        logger.warning(f"error querying for principal with username: '{username}' in rbac, err: {err}")
+        return handle_error(f"Internal error - failed to query rbac for user '{username}'", 500)
+
+    if len(principals) > 1:
+        logger.warning(f"multiple principles with username '{username}' exist in rbac")
+        return handle_error(f"Internal error - multiple principles with username '{username}' exist in rbac", 500)
+
+    groups = Group.platform_default_set()
```
astrozzc, line 384: Has to filter by tenant(org) to get the exact default group for that user. Would be great to have a unit test with custom default group

rbac/internal/views.py
```diff
@@ -338,6 +341,155 @@ def get_org_admin(request, org_or_account):
...
+    except Exception as err:
+        return handle_error(f"Internal error - couldn't get user from bop: {err}", 500)
+
+    result = {
+        "username": user["username"],
+        "email_address": user["email"],
+    }
+
+    try:
+        principals = Principal.objects.filter(username__iexact=user["username"])
```
astrozzc, line 375: We don't have username indexed currently, so this query might be slow. Better to get the org_id from the response and get the tenant and combine the query (we have unique index for username and tenant)

rbac/internal/views.py
```diff
@@ -338,6 +341,155 @@ def get_org_admin(request, org_or_account):
...
+        raise Exception(f"unexpected status: '{status}' returned from bop")
+
+    users = resp["data"]
+
+    if len(users) == 0:
+        raise UserNotFoundError(f"user with '{query_by}={principal}' not found in bop")
+
+    user = users[0]
+
+    if ("username" not in user) or (not user["username"]) or (user["username"].isspace()):
```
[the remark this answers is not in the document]
dagbay-rh: yeah I figure this is an extreme edge case. Just wanted to be extra defensive against potential bad data so we don't blow up with a difficult error to debug if it were to happen

rbac/internal/views.py
```diff
@@ -338,6 +341,155 @@ def get_org_admin(request, org_or_account):
...
+        raise Exception(
+            f"invalid user data for user '{query_by}={principal}': user found in bop but no username exists"
+        )
+
+    if "is_org_admin" not in user:
```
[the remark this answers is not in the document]
dagbay-rh: Gotcha, yeah that makes sense. I had a similar idea here of being defensive against a likely rare edge case

rbac/internal/middleware.py
```diff
@@ -39,14 +39,23 @@ class InternalIdentityHeaderMiddleware(MiddlewareMixin):
 
     header = RH_IDENTITY_HEADER
 
+    def __is_no_auth(self, request):
```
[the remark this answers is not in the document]
dagbay-rh: Yeah so this is a sort of weird case. We want the endpoint to be accessible behind the VPN only, but beyond that no auth is required. So in order to make that happen we are going to do this ticket in turnpike: https://issues.redhat.com/browse/RHCLOUD-37506 

so once that is done, we can make this endpoint specifically require VPN in turnpike. I am for sure open to suggestions if there is a better way to ensure this doesn't get exposed publicly tho. I could make a new "internal, vpn only" path that doesn't stem from the `_private` path since that is already exposed in turnpike. What do you think?

rbac/internal/views.py
dagbay-rh: oooo good to know ty! I'll add a test for coverage. I think it will be covered when I address your comment around getting orgId/tenant since we are getting the user from bop. I'll be sure to include test it out

rbac/internal/views.py
dagbay-rh: awesome, yeah I'll absolutely do that. Ty for the help :heart: I didn't realize I could just snag the org id from the bop user since we are querying for that already. Makes total sense

rbac/internal/middleware.py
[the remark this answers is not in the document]
astrozzc: Who is going to call this endpoint? If it is services within same cluster, we can let them call rbac through internal routes example: http://rbac-service.rbac-stage.svc.cluster.local:8000/_private/xxx

I am making a use case here for the HBI team to call us with preshared key: https://github.com/RedHatInsights/insights-rbac/pull/1566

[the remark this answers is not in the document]
dagbay-rh: It will be the customer service team, they want it to help troubleshoot issues quicker for users. So I think they want it exposed via the console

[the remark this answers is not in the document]
astrozzc: oh, yea, seems behind the VPN is a good solution for this.
@coderbydesign Keith do you have any suggestion on this?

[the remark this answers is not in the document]
coderbydesign: @dagbay-rh @astrozzc - Out of curiosity, what's the motivation for wanting/needing to not have authentication for this? For instance, most (I can't think of one that doesn't) other internal services (even behind VPN) require a valid internal SSO session. We could configure it in Turnpike to not be tied to a specific LDAP group, but still enforce they're an internal authenticated user.

dagbay-rh:
cover new cases in tests:

- [ ] org_id doesnt exist in bop
- [ ] exception when calling get_principal

rbac/management/utils.py
```diff
@@ -122,9 +122,9 @@ def get_principal(
 
 def verify_principal_with_proxy(username, request, verify_principal=True):
     """Verify username through the BOP."""
-    org_id = request.user.org_id
-    proxy = PrincipalProxy()
     if verify_principal:
+        org_id = request.user.org_id
```
dagbay-rh, line 126: bumping the check up just because the way I am using this method does not have `request.user.org_id` set

I am using `verify_principal=False` so this check doesn't run anyways

rbac/internal/views.py
dagbay-rh: ok reworked this to query for username and tenant now. Get the tenant here: https://github.com/RedHatInsights/insights-rbac/pull/1569/files#diff-fb45a5441f65035cd9b020f4b00090da2927a3b9c40c9dc56848e7129c9219ceR382

and then call this `utils.get_principal` method here: https://github.com/RedHatInsights/insights-rbac/pull/1569/files#diff-fb45a5441f65035cd9b020f4b00090da2927a3b9c40c9dc56848e7129c9219ceR390

which is doing the same query I was, but with the tenant:
https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/utils.py#L103

and this method has the added bonus of creating the principal if not found in RBAC despite the user existing in BOP

rbac/internal/middleware.py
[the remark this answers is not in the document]
dagbay-rh: following up after slack convo (I'll link the convo in the ticket):

1) we don't need the vpn requirement
2) we want to enforce the route is only available to internal users in turnpike
3) we want standard auth enforced in rbac

so for this PR:
- [ ] remove the no auth workaround, we want auth enforced
- [ ] create turnpike MR for this route, and require user to be internal to red hat (will have to ask what role that is)
turnpike routes for internal rbac here (for stage): https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/insights/turnpike/deploy.yml#L407-412

astrozzc (approved):
LGTM!

PR #1583, opened 2025-03-21
Workspace hierarchy depth limits

coderbydesign:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-38965

We'll need to be able to configure limits per deployment to dynamically update the hierarchy limits imposed in RBAC as we roll out the workspace experience.

We should default this to be limited to only creating sub-workspaces of the default, meaning the max hierarchy would look like: Root -> Default -> Sub1|Sub2|Sub3

We should also ensure that this also allows for configuration to enforce only creating sub-workspaces at the Default level, meaning you could not create a sibling to the Default workspace and create sub-workspaces for it, when disabled.

## Local Testing
Configure the flags (`WORKSPACE_HIERARCHY_DEPTH_LIMIT` and  `WORKSPACE_RESTRICT_DEFAULT_PEERS `) in this PR and create workspaces at differing levels to exercise various use-cases.

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


rbac/management/workspace/view.py
```diff
@@ -140,7 +142,26 @@ def validate_workspace(self, request, action="create"):
             self.validate_required_fields(request, REQUIRED_PUT_FIELDS)
         if parent_id:
             validate_uuid(parent_id)
-            if not Workspace.objects.filter(id=parent_id, tenant=tenant).exists():
+            try:
+                if self._exceeds_depth_limit(parent_id, tenant):
+                    message = f"Workspaces may only nest {settings.WORKSPACE_HIERARCHY_DEPTH_LIMIT} levels deep."
+                    error = {"workspace": [message]}
+                    raise serializers.ValidationError(error)
+                if self._violates_peer_restrictions(parent_id, tenant):
```
astrozzc, line 150: @Ellen-Yi-Dong Ellen is making a pr to populate the parent id with default workspace if it is none. But that's done in the model save method.
Wondering if it is better to put that before this method or put this into the workspace model verification.

Ellen-Yi-Dong: Oh! I had placed it within the model save method because I had trouble saving the updated parent id (but that could have also been me just not understanding/failing to find the thing that saved the information to the database) 
I can try to test to see again whether or not that is the same case? 
(And maybe test the workspace model verification as well!)

alechenninger: Spoke with Keith about this in collab – Keith is going to propose a refactor which should address both issues and more

rbac/management/workspace/view.py
```diff
@@ -140,7 +142,17 @@ def validate_workspace(self, request, action="create"):
...
+                if self._exceeds_depth_limit(parent_id, tenant):
+                    message = f"Workspaces may only nest {settings.WORKSPACE_HIERARCHY_DEPTH_LIMIT} levels deep."
+                    error = {"workspace": [message]}
+                    raise serializers.ValidationError(error)
+            except ObjectDoesNotExist:
                 message = f"Parent workspace '{parent_id}' doesn't exist in tenant"
                 error = {"workspace": [message]}
                 raise serializers.ValidationError(error)
+
+    def _exceeds_depth_limit(self, parent_id, tenant):
```
lpichler, line 164: In order to remove one query from `Workspace#ancestors` , it is possible to put this method (or some variant) on workspace model and call just existing `_ancestry_queryset` to get length. (without   `Workspace.objects.filter(id__in=ancestor_ids)` which is in `Workspace#ancestors`)

rbac/management/workspace/view.py
coderbydesign: All, I've opened a draft PR. It's not fully formed, but I wanted to get it out for initial feedback since I'll be out next week: https://github.com/RedHatInsights/insights-rbac/pull/1613

rbac/management/workspace/view.py
coderbydesign: @lpichler I've updated some of the ancestor/descendant logic and queries in #1596 - what would you think about doing something similar to this: https://github.com/RedHatInsights/insights-rbac/pull/1596/files#diff-c71184a4354243afa33f717d763b1170dc7d65fa60f738c90fefc60e4f7f4befR70 where we could have something for ancestry that would just return the ancestor leave ids (and itself), in which case we wouldn't need to add the additional count. 

In #1596 I note that at first I abstracted the CTE out to be reused in the `descendants` method, but it would add an additional filter query. I'd like feedback on that, which we might be able to also use here if it makes sense and we DRY it up a bit.

lpichler: Yes, I agree with the approach in #1596 and (we can do discuss it in  #1596 and use it here)

rbac/management/workspace/service.py
```diff
@@ -78,6 +87,30 @@ def destroy(self, instance: Workspace) -> None:
...
+        """Determine if peer restrictions are violated."""
+        target_root_workspace = Workspace.objects.root(tenant=tenant)
+        if settings.WORKSPACE_RESTRICT_DEFAULT_PEERS and str(target_root_workspace.id) == str(target_parent_id):
+            return True
+        return False
+
+    def _exceeds_depth_limit(self, target_parent_id: uuid.UUID, tenant: Tenant) -> bool:
+        """Determine if depth limit is exceeded."""
+        target_parent_workspace = Workspace.objects.get(id=target_parent_id, tenant=tenant)
+        max_depth_for_workspace = len(target_parent_workspace.ancestors()) + 1
```
astrozzc, line 115: Should we use example len(current_workspace.desendants()) instead of 1. But currently it is valid, maybe in the future if there are already depths for the workspace to be moved.

astrozzc (approved):
LGTM!

rbac/management/workspace/service.py
coderbydesign: Per our chat, we can follow up on this since we want to get this in prior to migration. Thanks, @astrozzc!

lpichler (approved):

deploy/rbac-clowdapp.yml
```diff
@@ -1419,3 +1435,9 @@ parameters:
 - name: REMOVE_NULL_VALUE
   description: Remove null value from the list of workspace id
   value: 'False'
+- name: WORKSPACE_HIERARCHY_DEPTH_LIMIT
+  description: The limit of workspace hierarchy depth
+  value: '2'
+- name: WORKSPACE_RESTRICT_DEFAULT_PEERS
+  description: The limit of workspace hierarchy depth
```
petracihalova, line 1442: @coderbydesign the description is same for both variables 🤔

rbac/management/workspace/service.py
```diff
@@ -78,6 +87,30 @@ def destroy(self, instance: Workspace) -> None:
...
+            raise serializers.ValidationError(error)
+
     def _parent_id_attr_update(self, attr: str, value: str, instance: Workspace) -> bool:
         """Determine if the attribute being updated is parent_id."""
         return attr == "parent_id" and instance.parent_id != value
+
+    def _violates_peer_restrictions(self, target_parent_id: uuid.UUID, tenant: Tenant) -> bool:
+        """Determine if peer restrictions are violated."""
+        target_root_workspace = Workspace.objects.root(tenant=tenant)
+        if settings.WORKSPACE_RESTRICT_DEFAULT_PEERS and str(target_root_workspace.id) == str(target_parent_id):
```
petracihalova, line 108: what about ungrouped workspaces? it is expected to allow to create sub-workspaces?

lpichler: @petracihalova it is not. Only hosts could be in ungrouped. Do we allow to created sub workspaces ?

petracihalova: [RHCLOUD-40242](https://issues.redhat.com/browse/RHCLOUD-40242)

PR #1594, opened 2025-03-28
[RHCLOUD-37866]  - internal endpoint relations api resource lookup

EvanCasey13:
## Link(s) to Jira
https://issues.redhat.com/browse/RHCLOUD-37866

Internal endpoint to hit relations api locally for lookup resource. 

## Local Testing
Need to have relations api running locally

and using this curl command to hit relations
```
curl -X POST \
  'http://localhost:8000/_private/api/relations/lookup_resource/' \
  -H 'Content-Type: application/json' \
  -H "x-rh-identity: eyJpZGVudGl0eSI6IHsidHlwZSI6ICJBc3NvY2lhdGUiLCAiYXNzb2NpYXRlIjogeyJlbWFpbCI6ICJ1c2VyQGV4YW1wbGUuY29tIn19fQ==" \
  -H "User-Type: associate" \
  -d '{
        "resource_type": {
            "name": "group", 
            "namespace": "rbac"
        },
        "relation": "member",
        "subject": {
            "subject": {
                "type": {
                    "namespace": "rbac",
                    "name": "principal"
                },
                "id": "bob"
            }
        }
      }'
```

## ENV FILE setup
You need to include this env variables in .env for the internal endpoint to work
```
CLIENT_ID=''
CLIENT_SECRET=''
```

Client ID and secret taken from vault (I can show where)

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


rbac/internal/utils.py
```diff
@@ -30,6 +33,14 @@
 
 logger = logging.getLogger(__name__)
 
+HOST = os.getenv("HOST")
```
lpichler, line 36: it is needed to put var in setting and clower yaml, so we can use it in ephemeral, stage and prod
(for now you can just set defaults for example CLIENT_SECRET and ID set just empty string it is different, we need to get it from vault  -  we can do it later)
check how it is done in this pr https://github.com/RedHatInsights/insights-rbac/pull/1583/files

rbac/internal/views.py
```diff
@@ -73,11 +80,44 @@
...
+url = os.getenv("URL")
+grant_type = os.getenv("GRANT_TYPE")
+relations_api_server = os.getenv("relation_api_gRPC_server")
 
 logger = logging.getLogger(__name__)
 TENANTS = TenantCache()
 PROXY = PrincipalProxy()
 
+# Create redis client
+redis_client = redis.StrictRedis(host="localhost", port=6379, decode_responses=True)
```
lpichler, line 97: this runs when server but we can use existing login for it, I think it needs just import related redis classes, please take a look how redis cache is used here. look on TenantCache class ,..

rbac/internal/views.py
```diff
@@ -73,11 +80,44 @@
...
+conn = None
+# Create http client
+if HOST is not None:
+    conn = http.client.HTTPSConnection(HOST)
+
+if conn is not None:
+    token = get_jwt_from_redis(redis_client, conn, grant_type, client_id, client_secret, scopes, url)
```
lpichler, line 104: this should be done on each request

rbac/internal/views.py
```diff
@@ -1325,3 +1365,32 @@ def retrieve_ungrouped_workspace(request):
         return HttpResponse(json.dumps(data), content_type="application/json", status=201)
     except Exception as e:
         return HttpResponse(str(e), status=500)
+
+
+def lookup_resource(request):
+    """GET resource from relations api."""
+    # Request parameters for resource lookup on relations api
+    resource_type_name = "group"
```
lpichler, line 1373: those needs to be parameters in request

rbac/internal/views.py
```diff
@@ -1325,3 +1365,32 @@ def retrieve_ungrouped_workspace(request):
         return HttpResponse(json.dumps(data), content_type="application/json", status=201)
     except Exception as e:
         return HttpResponse(str(e), status=500)
+
+
+def lookup_resource(request):
```
lpichler, line 1370: maybe it will be good to put some into try catch block and proper error message.

rbac/internal/utils.py
```diff
@@ -91,3 +102,40 @@ def delete_bindings(bindings):
...
+    token = json_data["access_token"]
+    return token
+
+
+def get_jwt_from_redis(redis_client, conn, grant_type, client_id, client_secret, scopes, url):
+    """Retrieve jwt token from redis or generate from Redhat SSO if not exists in redis."""
+    # Test the connection
+    try:
+        redis_client.ping()
+        print("Connected to Redis!")
```
lpichler, line 128: it is possible to use logger.info instread of prints

lpichler (changes requested):
Thanks!, looking good as first step and it is working, I did some suggestions, let me know if you have any  questions

rbac/internal/utils.py
```diff
@@ -32,6 +36,15 @@
 
 logger = logging.getLogger(__name__)
 
+HOST = settings.REDHAT_STAGE_SSO
```
lpichler, line 39: is this duplicatable with what is view.py ? I see it there as well

EvanCasey13: Whoops yep, that was from testing I think

rbac/internal/views.py
```diff
@@ -77,9 +84,40 @@
...
+
 logger = logging.getLogger(__name__)
 TENANTS = TenantCache()
 PROXY = PrincipalProxy()
+JWT = JWTCache()
+
+
+conn = None
+# Create http client
+if HOST is not None:
```
lpichler, line 104: can you put it in the little class and memoize it ?

look on ITService 

and I call constructor  with
```
http.client.HTTPSConnection(HOST)
```
only when it is needed  ? (when API is called )

rbac/internal/utils.py
```diff
@@ -95,6 +98,50 @@ def delete_bindings(bindings):
     return info
 
 
+def get_jwt_token(conn, grant_type, client_id, client_secret, scopes, url):
```
lpichler, line 101: this method can be in sso service class.

SSO service class has takes to get token - so maybe it can rename to JWTTokenProvider, so something like this. In that case,  all methods which are for to get token should be here. (calling RH sso,..)

rbac/internal/views.py
```diff
@@ -77,9 +84,57 @@
...
+sso_service = SSOservice()
+token = get_jwt_from_redis(
+    sso_service.get_conn(), settings.TOKEN_GRANT_TYPE, client_id, client_secret, settings.SCOPE, settings.OPENID_URL
+)
+
+
+@contextmanager
+def create_client_channel(addr):
+    """Create secure channel for grpc requests."""
+    # Call credential object will be invoked for every single RPC
```
lpichler, line 130: here could called method from your classes:

something like this (pseudo code):

jwt_cache = JWTCache()
...
jwt_token_provider = JWTTokenProvider()

jwt_token_manager = JwtTokenManager(jwt_cache, jwt_token_provider)
token = jwt_token_manager.get_token() # this will be new token or from redis

rbac/internal/views.py
```diff
@@ -77,9 +84,57 @@
 )
 from api.utils import RESOURCE_MODEL_MAPPING, get_resources
 
+# environment variables
+client_id = os.getenv("CLIENT_ID")
```
lpichler, line 88: Those variable could be obtained also in settings.py and the you can variable in code,  it is fine that it will be here - you can use `ENVIRONMENT.get_value` here - we are doing some for other secret variables.

deploy/rbac-clowdapp.yml
```diff
@@ -434,6 +434,25 @@ objects:
             value: ${CLOWDER_ENABLED}
           - name: CW_NULL_WORKAROUND
             value: ${CW_NULL_WORKAROUND}
+            #Relations API variables
+          - name: RBAC_CLIENT_ID
```
lpichler, line 438: Those variable need to be defined also at bottom in this file - please check others vars - there can be put particular value and here it will just referenced by `${RBAC_CLIENT_ID}`

this way allow to override it for different environment like stage/production and ephemeral - we can just state it in app-interface.

rbac/internal/views.py
```diff
@@ -77,9 +84,57 @@
...
+    def get_conn(self):
+        """Get connection to sso stage."""
+        if settings.REDHAT_STAGE_SSO is not None:
+            self.conn = http.client.HTTPSConnection(settings.REDHAT_STAGE_SSO)
+        return self.conn
+
+
+sso_service = SSOservice()
+token = get_jwt_from_redis(
+    sso_service.get_conn(), settings.TOKEN_GRANT_TYPE, client_id, client_secret, settings.SCOPE, settings.OPENID_URL
```
lpichler, line 123: it is not needed to pass setting variable as parameters- it is possible to use them in code directly.

rbac/internal/JWT.py
```diff
@@ -0,0 +1,115 @@
+#
```
lpichler, line 1: name of this this can be something like jwt_utils (ideally not capitals )

rbac/internal/JWT.py
```diff
@@ -0,0 +1,115 @@
...
+        token = json_data["access_token"]
+        return token
+
+
+class JWTManager:
+    """Class to handle management of JWT tokens."""
+
+    def __init__(self, jwt_provider, jwt_cache):
+        """Establish connection to JWT cache and provider."""
+        self.JWT_Cache = jwt_cache
```
lpichler, line 90: we can have consitent case
self.JWT_Cache -> self.jwt_cache
self.JWT_Provider -> self.jwt_provider

rbac/internal/views.py
```diff
@@ -1447,6 +1471,48 @@ def retrieve_ungrouped_workspace(request):
...
+                        type=common_pb2.ObjectType(namespace=resource_type_namespace, name=resource_subject_name),
+                        id=resource_subject_id,
+                    ),
+                ),
+            )
+        responses = stub.LookupResources(request)
+
+        if responses:
+            for r in responses:
+                return HttpResponse(r.resource)
```
lpichler, line 1506: it would be great if we would have response in json format(you can useJsonResponse)

PR #1558, opened 2025-03-11
RHCLOUD-38648 - Internal principal endpoint slice error

EvanCasey13:
## Link(s) to Jira
https://issues.redhat.com/browse/RHCLOUD-38648

## Description of Intent of Change(s)
Reoccurring error of `TypeError unhashable type: 'slice'` within internal principal endpoint , issue was around the structure being returned by principal proxy that could not be sliced to be used in function.

## Old response
```
{
    "data": {
        "userCount": len(userList),  
        "users": [{'username': 'user1'}, 
                  {'username': 'user2'}, 
                  {'username': 'user3'}] 
    },
    "status_code": 200
}
```

## New response
```
{
    "data": [
        {"username": "test1"},
        {"username": "test2"}
    ],
    "status_code": 200,
    "userCount": len(userList)
}
```

## Local Testing
How can the feature be exercised?
How can the bug be exploited and fix confirmed?
Rerun tests using tox

To run specific tests related to change:

## Internal principal endpoint
`tox -- tests/internal/integration -k test_principals_for_group_offset_limit`

`tox -- tests/internal/integration -k test_principals_for_group_pagination_exists`

## Principals
`tox -- tests/management/principal -k test_read_principal_all_username_only`

Is any special local setup required?

## Checklist
- [ ] if API spec changes are required, is the spec updated?
- [ ] are there any pre/post merge actions required? if so, document here.
- [x] are theses changes covered by unit tests?
- [ ] if warranted, are documentation changes accounted for?
- [ ] does this require migration changes?
  - [ ] if yes, are they backwards compatible?
- [ ] is there known, direct impact to dependent teams/components?
  - [ ] if yes, how will this be handled?

## Secure Coding Practices Checklist Link
- https://github.com/RedHatInsights/secure-coding-checklist

## Secure Coding Practices Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [ ] General Coding Practices

PR #1595, opened 2025-03-28
[RHCLOUD-39098] fix for list principals usernames_only

petracihalova:
## Link(s) to Jira
- [RHCLOUD-39098](https://issues.redhat.com/browse/RHCLOUD-39098)

1. fix the principal filter queryset to include the tenant (before we get all principals from DB = cca 780,000 in production)
2. fix the principal filter queryset to exclude cross account principals
3. add condition to filter principals by specified set of usernames = used for listing principals in the group (accidentally removed here https://github.com/RedHatInsights/insights-rbac/pull/1558)

the missing tenant (1) has been present for long time (> 2 years) but due to the additional condition (3) and the fact that the majority of requests was to list usernames only within a group, the RBAC service was still able to process them, however without the filter condition (3), we experienced a massive increase in memory usage during the production release

with this change we will be able to proceed with the production release without the memory usage issue, additionally we expect latency improvement in the `/groups/uuid/principals/` endpoint too - because of (1)

## Local testing
affected queries:
- `GET /principals/?username_only=true`
- `GET /groups/<uuid>/principals/?username_only=true`

in both cases, the BOP calls is expected to be skipped, and records should be returned from the RBAC database

since we have a cleaning job in place, the response should not differ significantly from queries without the `username_only=true` parameter

we need to ensure these queries are working as expected:
- `GET /principals/` returns all user based principals within the tenant from BOP
- `GET /principals/?username_only=true` returns all user based principals within the tenant, but only usernames from RBAC DB
- `GET /groups/<uuid>/principals/` returns the user based principals that belongs to the group (all principals belong to the correct tenant) 
- `GET /groups/<uuid>/principals/?username_only=true` returns the user based principals that belongs to the group, but only usernames from RBAC DB (all principals belong to the correct tenant)

responses do not contain cross account principals

everything should work for service accounts queries too (with username_only=true we skip the IT call as we do for user based principals but without the cleaning job the service accounts are not removed from db = we need some solution asap)
- `GET /principals/?type=service-account`
- `GET /principals/?type=service-account&username_only=true`
- `GET /groups/<uuid>/principals/?principal_type=service-account`
- `GET /groups/<uuid>/principals/?principal_type=service-account&username_only=true`

everything should work for all principals queries too (this is working now only for listing all principals within the tenant, for listing principals in a group will be added here https://github.com/RedHatInsights/insights-rbac/pull/1567)
- `GET /principals/?type=all`
- `GET /principals/?type=all&username_only=true`

we need to ensure the `GET /principals/` query is working with `usernames` query parameter, for example:
- `GET /principals/?usernames=<username1>,<username2>` 
- `GET /principals/?username_only=true&usernames=<username1>,<username2>` 

we need to ensure the `GET /groups/<uuid>/principals/` query is working with `principal_username` query parameter, for example:
- `GET /groups/<uuid>/principals/?principal_username=<value>`
- `GET /groups/<uuid>/principals/?username_only=true&principal_username=<value>`

### DIFF
--- a/rbac/management/principal/proxy.py
+++ b/rbac/management/principal/proxy.py
@@ -161,7 +161,7 @@ class PrincipalProxy:  # pylint: disable=too-few-public-methods
         """Send request to proxy service."""
         metrics_method = method.__name__.upper()
         if params and params.get("username_only") == "true":
-            principals = Principal.objects.filter(type="user", tenant__org_id=org_id)
+            principals = Principal.objects.filter(type="user", tenant__org_id=org_id, cross_account=False)
             offset = params.get("offset", 0)
             limit = params.get("limit", StandardResultsSetPagination.default_limit)
             userList = [dict(username=principal.username) for principal in principals]
@@ -171,7 +171,7 @@ class PrincipalProxy:  # pylint: disable=too-few-public-methods
         if settings.BYPASS_BOP_VERIFICATION:
             to_return = []
             if data is None:
-                for principal in Principal.objects.filter(type="user", tenant__org_id=org_id):
+                for principal in Principal.objects.filter(type="user", tenant__org_id=org_id, cross_account=False):
                     to_return.append(
                         dict(
                             username=principal.username,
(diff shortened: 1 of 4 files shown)


petracihalova:
only rebased, no other change

lpichler (approved):

PR #1596, opened 2025-03-28
Support v1 /access/ workspace hierarchy

coderbydesign:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-39095

[Add configuration for dynamically setting the workspace permission app/key](https://github.com/RedHatInsights/insights-rbac/commit/2009253ba33ab90b00dc2405363a1915cb846cde) 

Currently, v1 access checks for workspaces are checked against `inventory:groups:read|write`
permissions, with resource definitions of value `group.id` as the attribute filter key.

In order to make this flexible if we want to update this to use RBAC-specific
permissions, we can do that.

----

[Add application property to the ResourceDefinition model](https://github.com/RedHatInsights/insights-rbac/commit/7e76a1bde25a91cae8ef1cc8db05cfc9e3f3ae41) 

We'll need to check the `application` value of the associated permission for the
access object of a resource definition. In order to adhere to the law of demiter,
this creates a property on a `ResourceDefinition` so you can call `instance.application`
directly, and safely.

----

[Add descendants method to Workspace model](https://github.com/RedHatInsights/insights-rbac/commit/a1926d11f2b06fec936292f1de18087f5690d2af) 

In order to efficiently query descendant workspace IDs for a given workspace instance,
this follows the same pattern we've used for `anscestors` of workspaces, using a
CTE to collect all IDs to then filter in a single query.

----

[Update the ResourceDefinition serializer to return descendant workspace IDs](https://github.com/RedHatInsights/insights-rbac/commit/94092e47f4dcbe57e1b4f6a7ff3158bc47f735b1) 

We override `to_representation` here to append descendant workspace IDs to the
JSON output conditionally.

If the `ResourceDefinition` being rendered is not for a workspace permission, and
does not have an ID key for the attribute filter, nothing changes.

In the case where it does, we'll collect descendant workspace IDs and append them
to the filter list to be returned to the client.

----

**Performance considerations:** I used the [Silk profiler](https://github.com/jazzband/django-silk) locally, and there looks to be around an 11ms hit with 3 additional queries being performed to get the permission, and then two queries for the workspace being iterated on, and another for all descendants. All other relations are being prefetched. We may be able to optimize this more, or do additional testing with scaled nested workspace IDs. 

The other option is to explore calling Kessel directly to query these IDs.
Screenshot 2025-03-28 at 3 34 26 PM

## Local Testing
- Create a hierarchy of workspaces locally
- Create an `inventory:groups:read` role which grants granular access via a resourceDefinition to one of the parent workspaces ONLY, similar to: `curl http://localhost:8000/api/rbac/v1/roles/ -XPOST -H 'Content-Type: application/json' -d '{"name": "Workspace Hierarchy", "access": [{"permission": "inventory:groups:read", "resourceDefinitions": [{"attributeFilter": {"key": "group.id", "operation": "in", "value": ["<parent-uuid>"]}}]}]}'`
- Assign this new role to a group you have access to
- Query `/access/?application=inventory` and check for that granular permission result
- View that you only see that single workspace in your resource definition list, **not** the descendants
- Switch to this branch and run the same access query
- Note that you now see the descendant IDs as well

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

## Summary by Sourcery

Add support for workspace hierarchy in v1 access checks by dynamically configuring workspace permission application and attribute filtering

New Features:
- Implement dynamic workspace permission configuration for access checks
- Add ability to retrieve descendant workspace IDs in resource definitions

Enhancements:
- Extend ResourceDefinition model to support application-specific workspace permissions
- Add descendants method to Workspace model for efficient hierarchical queries

Deployment:
- Add new environment variables for configuring workspace application name and attribute filter


rbac/management/role/serializer.py
```diff
@@ -57,6 +58,28 @@ class Meta:
...
 
+    def to_representation(self, instance):
+        """Representation of ResourceDefinitions."""
+        data = super().to_representation(instance)
+        if self._is_workspace_filter(instance):
+            data.get("attributeFilter").update({"value": self._workspace_descendant_ids(instance)})
+        return data
+
+    def _is_workspace_filter(self, instance):
+        is_inventory_permission = instance.application == settings.WORKSPACE_APPLICATION_NAME
```
coderbydesign, line 69: Should we also enforce the resource type here?

rbac/management/role/serializer.py
```diff
@@ -57,6 +58,28 @@ class Meta:
...
+        is_inventory_group_filter = instance.attributeFilter.get("key") == settings.WORKSPACE_ATTRIBUTE_FILTER
+        return is_inventory_permission and is_inventory_group_filter
```
coderbydesign, line 71: Should we also enforce the resource type here?

rbac/rbac/settings.py
```diff
@@ -499,3 +499,7 @@
 V2_APIS_ENABLED = ENVIRONMENT.bool("V2_APIS_ENABLED", default=False)
 V2_READ_ONLY_API_MODE = ENVIRONMENT.bool("V2_READ_ONLY_API_MODE", default=False)
 READ_ONLY_API_MODE = ENVIRONMENT.get_value("READ_ONLY_API_MODE", default=False)
+
+# Workspace settings
+WORKSPACE_APPLICATION_NAME = ENVIRONMENT.get_value("WORKSPACE_APPLICATION_NAME", default="inventory")
+WORKSPACE_ATTRIBUTE_FILTER = ENVIRONMENT.get_value("WORKSPACE_ATTRIBUTE_FILTER", default="group.id")
```
coderbydesign, line 505: Are these valuable potentially, or overkill?

theute:
@sourcery-ai summary

theute:
@sourcery-ai review

rbac/management/workspace/model.py
```diff
@@ -93,3 +93,28 @@ def _ancestry_queryset(self):
         """,
             [self.id],
         )
+
+    def descendants(self):
```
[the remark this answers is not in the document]
alechenninger: Maybe we don't want the other dependency. However, one thing I do like about this is I think it makes it into one less query. Right now it does one query (the CTE) to get the IDs, and then another to get the workspace models for those. With this it looks like the library makes it easy to optimize that.

That said, we could probably work out that optimization ourselves.

I think if we just made the select use * instead of those fields, we could parse the results directly as Workspace model objects without the second `id__in` query?

rbac/management/workspace/model.py
```diff
@@ -93,3 +93,28 @@ def _ancestry_queryset(self):
...
+        """Return a list of descendants for a Workspace instance."""
+        descendant_ids = [d.id for d in self._descendant_queryset() if d.id != self.id]
+        descendants = Workspace.objects.filter(id__in=descendant_ids)
+        return descendants
+
+    def _descendant_queryset(self):
```
alechenninger, line 103: Starting a separate thread to avoid the AI spam heh.

So what's interesting is that as of now we don't actually need the full models in the queries.

I wonder if you could simply do this?

```suggestion
        descendants = Workspace.objects.filter(id__in=self._descendant_queryset())
        return descendants

    def _descendant_queryset(self):
```

This would make it just part of the same SQL query by lazily evaluating the queryset, I think.
It would include your _own_ id, which is not what this does currently, but if we need to exclude that, maybe we could remove that in the CTE:

```
            WITH RECURSIVE descendants AS
              (SELECT id,
                      parent_id
               FROM management_workspace
               WHERE id = %s
               UNION SELECT w.id,
                                w.parent_id
               FROM management_workspace w
               JOIN descendants d ON w.parent_id = d.id)
            SELECT id
            FROM descendants
            WHERE id <> %s
```

(note additional where on the end)

rbac/management/role/serializer.py
```diff
@@ -57,6 +58,28 @@ class Meta:
...
+    def _workspace_descendant_ids(self, instance):
+        workspace_ids = instance.attributeFilter.get("value")
+        workspaces = Workspace.objects.filter(id__in=workspace_ids).only("id")
+        all_descendant_ids = set()
+        for workspace in workspaces:
+            descendant_queryset = workspace.descendants().values_list("id", flat=True)
+            descendant_ids = list(map(str, descendant_queryset))
+            all_descendant_ids.update(descendant_ids)
+        return list(set(workspace_ids) | set(all_descendant_ids))
+
```
alechenninger, line 82: Just capturing here what we discussed in collab.

We could have a bit of an N * M thing going on here, where every N permission ends up querying descendants for every M workspace in its resource definition.

So we could use a request-level cache, using `self.context.request` (see `get_serializer_context` in DRF's `GenericAPIView` which adds the request as context). We check that, if a miss, query, and add it to the cache.

You could populate the cache just with the "parent" workspace that is queried, although this does miss some opportunities for cache reuse in case a later resource definition refers to a descendant of a prior parent.

e.g. let's say we have a hierarchy like...

```
/1
/1/2
/1/2/3
/1/4
```

If 1 is queried, we store 2, 3, and 4 for 1.

But if 2 is queried, we technically already looked up the hierarchy for 2–it had 3. But we wouldn't know from how we've structured the cache. So we'd end up running another query and getting back 3.

If we stored a tree structure in memory, we could remove the need to query for any descendant again. We look up 2, and see we already know it has 3. But that is a little more involved I guess.

alechenninger: This might work:

On cache miss...

1. Evaluate the queryset, getting parent_id and id for each workspace
2. Add each ws to a dictionary by ID
3. For each workspace, lookup its parent and add itself to a "children" array on the ws object (or another object, w/e)

Then to get the descendands...

4. Get the workspace by ID (in memory by this point). Have a method somewhere which can lookup descendants in memory by traversing the cache structure – starting with itself, recursively iterate over its children workspaces to get their IDs.

alechenninger: It *may* be worth putting this in reddis so we can reuse cross request (at the cost of eventual consistency) but I would see how things look in practice first.

One quirk to keep in mind: if workspaces are moving concurrently to this, and this is not done within a single repeatable read transaction (I think we're still using read committed), a workspace could appear in multiple hierarchies. Probably a rare and benign enough case that it's not a huge deal.

rbac/rbac/settings.py
alechenninger: IMO it's good. These are technically externally configured values so I think it makes sense here.

rbac/management/workspace/model.py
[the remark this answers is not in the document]
coderbydesign: @alechenninger the only issue with doing what you suggested by querying `*` and then using that, is that a `RawQuerySet` is iterable, but not an actual `QuerySet`, so you can't call filter/exclude/first/last/etc. on the results. Since we're providing the results of this as a method response vs something used internally, I'd prefer we still had an actual `QuerySet` to pass around. That said, I think we can use `RawSQL` to select `id` values and pass that into the `filter` for a single hit.

rbac/management/role/serializer.py
coderbydesign: That brings up a good point though, specifically if we use redis, in how we'd need to bust the cache for a set of the hierarchy when workspaces are moved/deleted

coderbydesign: I'm trying a different approach here altogether. We'll see if it shakes out, but basically using another CTE to accept an array of values, and just pass that directly from the resource definition to do it all in one query. It almost seems too simple, so I may be missing something. I've got existing tests passing, but will add more cases and push something up.

rbac/management/managers.py
```diff
@@ -66,3 +66,24 @@ def standard(self, tenant=None, tenant_id=None):
         """Delegate call to the WorkspaceQuerySet."""
         tenant_id = self._get_tenant_id(tenant=tenant, tenant_id=tenant_id)
         return self.get_queryset().standard(tenant_id)
+
+    def tree_ids_from_roots(self, ids):
```
coderbydesign, line 70:
```suggestion
    def descendant_ids_with_parents(self, ids):
```

rbac/management/managers.py
```diff
@@ -66,3 +66,24 @@ def standard(self, tenant=None, tenant_id=None):
...
+
+    def descendant_ids_with_parents(self, ids):
+        """Return the descendant and root workspace IDs based on roots supplied."""
+        with connection.cursor() as cursor:
+            sql = """
+                WITH RECURSIVE descendants AS
+                    (SELECT id,
+                            parent_id
+                    FROM management_workspace
+                    WHERE id = ANY(%s::uuid[])
```
lpichler, line 78: Should we include only standard workspaces ?

lpichler:
@coderbydesign when I created role and I was on this branch:

```

POST http://localhost:8000/api/rbac/v1/roles/
x-rh-identity: eyJlbnRpdGxlbWVudHMiOnsiYW5zaWJsZSI6eyJpc19lbnRpdGxlZCI6dHJ1ZX0sImNvc3RfbWFuYWdlbWVudCI6eyJpc19lbnRpdGxlZCI6dHJ1ZX0sImh5YnJpZCI6eyJpc19lbnRpdGxlZCI6dHJ1ZX0sImluc2lnaHRzIjp7ImlzX2VudGl0bGVkIjp0cnVlfSwibWlncmF0aW9ucyI6eyJpc19lbnRpdGxlZCI6dHJ1ZX0sIm9wZW5zaGlmdCI6eyJpc19lbnRpdGxlZCI6dHJ1ZX0sInNtYXJ0X21hbmFnZW1lbnQiOnsiaXNfZW50aXRsZWQiOnRydWV9fSwiaWRlbnRpdHkiOnsidHlwZSI6IlVzZXIiLCJhdXRoX3R5cGUiOiJiYXNpYy1hdXRoIiwib3JnX2lkIjoiMTM0MzE3MTIiLCJhY2NvdW50X251bWJlciI6IjYzODc4OTYiLCJ1c2VyIjp7InVzZXJuYW1lIjoicmhuLWVuZ2luZWVyaW5nLXJibGFuY28iLCJlbWFpbCI6InJibGFuY29AcmVkaGF0LmNvbSIsImZpcnN0X25hbWUiOiJSb21hbiIsImxhc3RfbmFtZSI6IlRlc3RpbmciLCJpc19hY3RpdmUiOnRydWUsImlzX29yZ19hZG1pbiI6dHJ1ZSwiaXNfaW50ZXJuYWwiOmZhbHNlLCJsb2NhbGUiOiJlbi1VUyIsInVzZXJfaWQiOiI1MzUyNzU3MiJ9LCJpbnRlcm5hbCI6eyJvcmdfaWQiOiIxMzQzMTcxMiIsImF1dGhfdGltZSI6MTY1MjExODUzMH19fQ==

{
  "name": "Workspace Hierarchy New 2",
  "access": [
    {"permission": "inventory:groups:read",
      "resourceDefinitions": [
        {"attributeFilter":
            {"key": "group.id", "operation": "in", "value": ["01956b44-f24c-76d1-8d64-19fce41e35ce", "01956b44-f24c-76d1-8d64-19e965dd2fe6"]}
        }
      ]
    }
  ]
}
```

is already lists hierarchy:
```
POST http://localhost:8000/api/rbac/v1/roles/

{
  "uuid": "7b45a76c-d035-4c15-b1ac-7ad2662156be",
  "name": "Workspace Hierarchy New 2",
  "display_name": "Workspace Hierarchy New 2",
  "description": null,
  "access": [
    {
      "resourceDefinitions": [
        {
          "attributeFilter": {
            "key": "group.id",
            "value": [
              "0196810f-3088-7811-a109-a2856b9ee3c0",
              "01956b44-f24c-76d1-8d64-19fce41e35ce",
              "0195b3fa-081d-7d43-a224-59b254e54452",
              "01968110-1cbb-71f2-9dbc-4447cc58e9a3",
              "0196812d-b369-73f2-a652-553dc15877d5",
              "01956b44-f24c-76d1-8d64-19e965dd2fe6",
              "01968110-8068-74b2-8174-316d4c0395c8"
            ],
            "operation": "in"
          }
        }
      ],
      "permission": "inventory:groups:read"
    }
  ],
  "applications": [
    "inventory"
  ],
  "system": false,
  "platform_default": false,
  "admin_default": false,
  "created": "2025-04-29T10:58:52.788194Z",
  "modified": "2025-04-29T10:58:52.788207Z",
  "external_role_id": null,
  "external_tenant": null
}

```

Is this expected ? I think it makes sense but It might confuse clients, maybe this evaluation should happen only in access endpoint ?

lpichler:
@coderbydesign 
Validation of ws uuid in value - there are several cases where I am curious how we will validate it, maybe we should validate it whether workspace uuid existing inside of users tenant, if so when we will to enable this validation, should it after HBI migration(I guess now it would block creation of role with inventory groups) ?  
we can make it configurable and keep current behaviour under env. variable but what would be behaviour of hierarchy query with invalid uuids ? How it will behave with existing data with non-existing workspaces ? I happy to hear your insights, eventually we can discuss in some collab meetings.

- When I used a workspace UUID with an invalid format in the attributeFilter during role creation, it returned a 500 error. 
- When I used a workspace UUID with a valid format, but belonging to a different tenant, in the attributeFilter during role creation, the role was created with that value. 
- When I used a workspace UUID with a valid format for a non-existent workspace in the attributeFilter during role creation, the role was created, and the value was saved in the resourceDefinitions table, but the API response returned empty array:

```
"access": [
    {
      "resourceDefinitions": [
        {
          "attributeFilter": {
            "key": "group.id",
            "value": [],
            "operation": "in"
          }
```

rbac/management/workspace/model.py
```diff
@@ -103,6 +98,27 @@ def _ancestry_queryset(self):
                JOIN ancestors a ON w.id = a.parent_id)
             SELECT id
             FROM ancestors
+            WHERE id != %s
+        """,
+        )
+        return Workspace.objects.filter(id__in=RawSQL(sql, [self.id, self.id]))
+
+    def descendants(self):
```
lpichler, line 106: Is this used anywhere or will it be used later ?

rbac/management/role/serializer.py
```diff
@@ -57,6 +58,20 @@ class Meta:
...
 
+    def to_representation(self, instance):
+        """Representation of ResourceDefinitions."""
+        data = super().to_representation(instance)
+        if self._is_workspace_filter(instance):
+            tree_ids = Workspace.objects.descendant_ids_with_parents(instance.attributeFilter.get("value"))
+            data.get("attributeFilter").update({"value": tree_ids})
+        return data
+
+    def _is_workspace_filter(self, instance):
```
astrozzc, line 69: Seems the instance is resource definition? maybe instance.access.permission.application
and instance.access.permission.resource_type?

coderbydesign: @astrozzc yeah if you notice in the [role model](https://github.com/RedHatInsights/insights-rbac/pull/1596/files#diff-36e87b5178c904f3090bb466061416ce6ff8fcb98dc15c8a69f407a347312a7fR111-R122) I've added those properties to the resource definition to delegate to, assuming that's what you're referring to?

rbac/management/workspace/model.py
coderbydesign: It's not. I had initially added it and was parsing out the id values prior to the CTE, and left it for parity with ancestors, but I think we could just remove it for now as well, and reference this if/when it's needed?

rbac/management/managers.py
coderbydesign: We'll definitely want/need root and default. I haven't considered ungrouped yet too much, but curious what you're thinking the issue might be there?

coderbydesign:
> @coderbydesign when I created role and I was on this branch:
> 
> ```
> 
> POST http://localhost:8000/api/rbac/v1/roles/
> x-rh-identity: eyJlbnRpdGxlbWVudHMiOnsiYW5zaWJsZSI6eyJpc19lbnRpdGxlZCI6dHJ1ZX0sImNvc3RfbWFuYWdlbWVudCI6eyJpc19lbnRpdGxlZCI6dHJ1ZX0sImh5YnJpZCI6eyJpc19lbnRpdGxlZCI6dHJ1ZX0sImluc2lnaHRzIjp7ImlzX2VudGl0bGVkIjp0cnVlfSwibWlncmF0aW9ucyI6eyJpc19lbnRpdGxlZCI6dHJ1ZX0sIm9wZW5zaGlmdCI6eyJpc19lbnRpdGxlZCI6dHJ1ZX0sInNtYXJ0X21hbmFnZW1lbnQiOnsiaXNfZW50aXRsZWQiOnRydWV9fSwiaWRlbnRpdHkiOnsidHlwZSI6IlVzZXIiLCJhdXRoX3R5cGUiOiJiYXNpYy1hdXRoIiwib3JnX2lkIjoiMTM0MzE3MTIiLCJhY2NvdW50X251bWJlciI6IjYzODc4OTYiLCJ1c2VyIjp7InVzZXJuYW1lIjoicmhuLWVuZ2luZWVyaW5nLXJibGFuY28iLCJlbWFpbCI6InJibGFuY29AcmVkaGF0LmNvbSIsImZpcnN0X25hbWUiOiJSb21hbiIsImxhc3RfbmFtZSI6IlRlc3RpbmciLCJpc19hY3RpdmUiOnRydWUsImlzX29yZ19hZG1pbiI6dHJ1ZSwiaXNfaW50ZXJuYWwiOmZhbHNlLCJsb2NhbGUiOiJlbi1VUyIsInVzZXJfaWQiOiI1MzUyNzU3MiJ9LCJpbnRlcm5hbCI6eyJvcmdfaWQiOiIxMzQzMTcxMiIsImF1dGhfdGltZSI6MTY1MjExODUzMH19fQ==
> Content-Type: application/json
> 
> {
>   "name": "Workspace Hierarchy New 2",
>   "access": [
>     {"permission": "inventory:groups:read",
>       "resourceDefinitions": [
>         {"attributeFilter":
>             {"key": "group.id", "operation": "in", "value": ["01956b44-f24c-76d1-8d64-19fce41e35ce", "01956b44-f24c-76d1-8d64-19e965dd2fe6"]}
>         }
>       ]
>     }
>   ]
> }
> ```
> 
> is already lists hierarchy:
> 
> ```
> POST http://localhost:8000/api/rbac/v1/roles/
> 
> {
>   "uuid": "7b45a76c-d035-4c15-b1ac-7ad2662156be",
>   "name": "Workspace Hierarchy New 2",
>   "display_name": "Workspace Hierarchy New 2",
>   "description": null,
>   "access": [
>     {
>       "resourceDefinitions": [
>         {
>           "attributeFilter": {
>             "key": "group.id",
>             "value": [
>               "0196810f-3088-7811-a109-a2856b9ee3c0",
>               "01956b44-f24c-76d1-8d64-19fce41e35ce",
>               "0195b3fa-081d-7d43-a224-59b254e54452",
>               "01968110-1cbb-71f2-9dbc-4447cc58e9a3",
>               "0196812d-b369-73f2-a652-553dc15877d5",
>               "01956b44-f24c-76d1-8d64-19e965dd2fe6",
>               "01968110-8068-74b2-8174-316d4c0395c8"
>             ],
>             "operation": "in"
>           }
>         }
>       ],
>       "permission": "inventory:groups:read"
>     }
>   ],
>   "applications": [
>     "inventory"
>   ],
>   "system": false,
>   "platform_default": false,
>   "admin_default": false,
>   "created": "2025-04-29T10:58:52.788194Z",
>   "modified": "2025-04-29T10:58:52.788207Z",
>   "external_role_id": null,
>   "external_tenant": null
> }
> ```
> 
> Is this expected ? I think it makes sense but It might confuse clients, maybe this evaluation should happen only in access endpoint ?

Ah, because it's running through the serializer for this as well, right? 🤔 Trying to think of a downside or side-effect of this. I honestly don't think they're used by the UI at the moment to render anything (since they're generic IDs) so I don't think it would be an issue, and is functionally accurate. But I may be missing something. Good call-out.

coderbydesign:
> @coderbydesign Validation of ws uuid in value - there are several cases where I am curious how we will validate it, maybe we should validate it whether workspace uuid existing inside of users tenant, if so when we will to enable this validation, should it after HBI migration(I guess now it would block creation of role with inventory groups) ? we can make it configurable and keep current behaviour under env. variable but what would be behaviour of hierarchy query with invalid uuids ? How it will behave with existing data with non-existing workspaces ? I happy to hear your insights, eventually we can discuss in some collab meetings.
> 
> * When I used a workspace UUID with an invalid format in the attributeFilter during role creation, it returned a 500 error.
> * When I used a workspace UUID with a valid format, but belonging to a different tenant, in the attributeFilter during role creation, the role was created with that value.
> * When I used a workspace UUID with a valid format for a non-existent workspace in the attributeFilter during role creation, the role was created, and the value was saved in the resourceDefinitions table, but the API response returned empty array:
> 
> ```
> "access": [
>     {
>       "resourceDefinitions": [
>         {
>           "attributeFilter": {
>             "key": "group.id",
>             "value": [],
>             "operation": "in"
>           }
> ```

@lpichler yeah I'd like to chat about this one to make sure I'm understanding the concern correctly.

rbac/management/role/serializer.py
astrozzc: Oh, right! LGTM then

rbac/management/managers.py
coderbydesign: Based on conversation last week, we're going to leave this to include all.

coderbydesign:
@lpichler I've addressed your feedback in subsequent commits. I also addressed performance issues in https://github.com/RedHatInsights/insights-rbac/pull/1596/commits/62ec6b0f48a042be7efb03031a2bcba5b1d5dddb which removed n+1 queries in the serializer:

**Profiling `to_representation` on Resource Definitions before prefetching:**
Screenshot 2025-05-14 at 9 50 11 AM

**Profiling `to_representation` on Resource Definitions after prefetching:**
Screenshot 2025-05-14 at 9 50 16 AM

lpichler (approved):
👏 👏 👏 LGTM

rbac/management/role/serializer.py
```diff
@@ -70,6 +77,33 @@ class Meta:
...
+
+    def _original_vals_and_descendant_ids(self, instance):
+        attr_filter_list = value_to_list(instance.attributeFilter.get("value"))
+        uuids = [val for val in attr_filter_list if is_valid_uuid(val)]
+        non_uuids = [val for val in attr_filter_list if not is_valid_uuid(val)]
+        ids_with_parents = Workspace.objects.descendant_ids_with_parents(uuids, instance.tenant_id)
+        return list(set(non_uuids + ids_with_parents))
+
+    def _should_add_hierarchy(self, instance):
+        hierarchy_enabled = settings.WORKSPACE_HIERARCHY_ENABLED is True
```
astrozzc, line 97: Do we enable this anytime after the migration?

astrozzc (approved):
LGTM!

PR #1632, opened 2025-04-11
Remove relations when permission gets deleted

astrozzc:
## Link(s) to Jira
-

The what, why and how.

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


rbac/management/role/relation_api_dual_write_handler.py
```diff
@@ -76,18 +76,30 @@ class SeedingRelationApiDualWriteHandler(BaseRelationApiDualWriteHandler):
...
         """Generate & store role's current relations."""
         if not self.replication_enabled():
             return
+        if not role:
+            role = self.role
+        assert isinstance(role, Role)
         self._current_role_relations = self._generate_relations_for_role(role)
 
-    def replicate_update_system_role(self, role: Role):
+    def replicate_update_system_role(self, role: Optional[Role] = None):
```
alechenninger, line 95: Taking a step back – looks like we have 3 ways now to assign the role. Are these all equivalent though? What is the correct way to use this class? It's not clear to me here and I suspect may be confusing for others.

rbac/internal/views.py
```diff
@@ -695,7 +696,7 @@ def permission_removal(request):
         with transaction.atomic():
             try:
                 logger.warning(f"Deleting permission '{permission}'. Requested by '{request.user.username}'")
-                permission_obj.delete()
+                delete_permission(permission_obj)
```
alechenninger, line 699: When is the admin endpoint called vs when are permissions removed during seeding?

I'm wondering if the logic for handling replication of permission removal should be done in the seeding handler or at least that module.

rbac/management/role/relation_api_dual_write_handler.py
```diff
@@ -350,3 +361,25 @@ def _generate_relations_and_mappings_for_role(self):
                 f"Failed to generate relations and mappings for role {self.role.name}, UUID :{self.role.uuid}: {e}"
             )
             raise DualWriteException(e)
+
+
+def delete_permission(permission: Permission):
```
alechenninger, line 366: I would probably move this to the seeding handler. It seems like more of a seeding concern. Yes it can involve custom roles, but that is as a result of a system-level "seeded" concept (permission) changing.

rbac/management/role/relation_api_dual_write_handler.py
astrozzc: Right, it is better to be consistent, updated the seeding handler to be consistent

lpichler (approved):

PR #1649, opened 2025-04-17
[RHCLOUD-27273] Update resource definitions internal endpoint

petracihalova:
## Link(s) to Jira
- [RHCLOUD-27273](https://issues.redhat.com/browse/RHCLOUD-27273)

example of incorrect resource definitions
```
{"attributeFilter": {"key": "key1_id", "operation": "equal", "value": ["value1", "value2"]}},
{"attributeFilter": {"key": "key2_id", "operation": "in", "value": "value1, value2"}},
{"attributeFilter": {"key": "key3_id", "operation": "in", "value": "string"}},
```
**before the change**
- `GET /api/utils/resource_definitions/` returned count of incorrect resource definitions
- `PATCH /api/utils/resource_definitions/` fixed all incorrect resource definitions
  - if the `value` is list/array and the `operation` is "equal", then the `operation` is changed into "in"
  - if the `value` is string and the `operation` is "in", then the `operation` is changed into "equal"

**after the change**
- `GET /api/utils/resource_definitions/` returns count of incorrect resource definitions
  - with query parameter `detail=true` the endpoint returns list of resource definition objects -> you can see the resource definitions before the change and double check there is no unexpected data

- `PATCH /api/utils/resource_definitions/` fixes all incorrect resource definitions
  - if the `value` is list/array, the `operation` should be "in"
  - if the `value` is string, the `operation` should be "equal"
  - if the `value` is string as a list of items separated by comma, then the `value` should be converted into list

additionally for the PATCH endpoint you can use `id` query param to fix only 1 resource definitions -> the id can be identified with GET request with `detail=true` query param

## Local Testing
create few incorrect resource definitions in your RBAC local db and send GET request without/with the `detail` query param
```
GET /api/utils/resource_definitions/
GET /api/utils/resource_definitions/?detail=true
```

then try to fix the resource definitions with PATCH request
```
PATCH /api/utils/resource_definitions/
PATCH /api/utils/resource_definitions/?id=<resource_definition_id>
```

## Checklist
- [x] api spec update
- [x] unit tests
- [x] confluence documentation

## Summary by Sourcery

Update the internal endpoint for resource definitions to improve handling of attribute filter operations and values

New Features:
- Add support for retrieving detailed resource definition information using 'detail=true' query parameter
- Implement ability to fix a specific resource definition by ID using 'id' query parameter

Enhancements:
- Improve resource definition validation by correcting operation and value types
- Add more flexible handling of string and list values in attribute filters

Tests:
- Add comprehensive test cases for new resource definition endpoint functionality
- Create tests for retrieving and patching resource definitions with different scenarios

### DIFF
--- a/rbac/internal/specs/openapi.json
+++ b/rbac/internal/specs/openapi.json
@@ -2210,6 +2210,17 @@
         "summary": "Get incorrect resource definitions.",
         "description": "Get resource definitions with incorrect attribute filters. Attribute filters with lists must use 'in' operation. Those with a single string must use 'equal'.",
         "operationId": "GetIncorrectResourceDefinitions",
+        "parameters": [
+          {
+            "name": "detail",
+            "in": "query",
+            "description": "Optional flag. If true, returns a list of resource definition objects. If false or omitted, returns only the count.",
+            "required": false,
+            "schema":{
+              "type": "boolean"
+            }
+          }
+        ],
         "responses": {
           "200": {
             "description": "The number of resource definitions that would be corrected."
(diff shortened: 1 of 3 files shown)


astrozzc (approved):

vbelchio:
QE: Passed

PR #1613, opened 2025-04-03
V2 validation and error handling

coderbydesign:
## Description of Intent of Change(s)
The what, why and how.

## Local Testing
How can the feature be exercised?
How can the bug be exploited and fix confirmed?
Is any special local setup required?

## Checklist
- [ ] if API spec changes are required, is the spec updated?
- [ ] are there any pre/post merge actions required? if so, document here.
- [ ] are theses changes covered by unit tests?
- [ ] if warranted, are documentation changes accounted for?
- [ ] does this require migration changes?
  - [ ] if yes, are they backwards compatible?
- [ ] is there known, direct impact to dependent teams/components?
  - [ ] if yes, how will this be handled?

## Secure Coding Practices Checklist Link
- https://github.com/RedHatInsights/secure-coding-checklist

## Secure Coding Practices Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [ ] General Coding Practices

## Summary by Sourcery

Refactor workspace validation and error handling to improve code structure and introduce a more robust validation approach

New Features:
- Add a new ProblemJsonResponse class to standardize error response format
- Create a WorkspaceValidationService to centralize and standardize workspace validation logic

Bug Fixes:
- Improve validation for workspace creation, update, and deletion
- Ensure more consistent and comprehensive validation across workspace operations

Enhancements:
- Introduce a centralized WorkspaceValidationService to handle all workspace-related validations
- Simplify view and serializer logic by moving validation logic to a dedicated service
- Implement a Problem JSON response format for more consistent error handling

PR #1658, opened 2025-04-25
V2 Validation Refactor

coderbydesign:
This supplants #1613 to reduce the scope of changes to just a validation refactor. We'll address exception handling and API responses in subsequent tickets/PRs.

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

## Summary by Sourcery

Refactor workspace validation logic and introduce a dedicated serializer for PATCH requests.

Enhancements:
- Introduce `WorkspacePatchSerializer` to handle partial updates, making `name` and `parent_id` optional for this operation.
- Remove custom validation logic from `WorkspaceViewSet`.
- Make `name` and `parent_id` required fields in the base `WorkspaceSerializer` for create and full update operations.

## Summary by Sourcery

Refactor workspace validation logic by introducing a dedicated service layer and simplifying serializer and viewset validation.

New Features:
- Introduced WorkspaceService to centralize workspace management logic

Enhancements:
- Simplified validation logic in WorkspaceViewSet
- Moved complex validation and business logic to WorkspaceService
- Made name and parent_id fields more consistently required

Chores:
- Removed custom validation methods from view
- Consolidated workspace creation and update logic


rbac/management/workspace/serializer.py
```diff
@@ -49,28 +48,45 @@ class Meta:
...
-                workspace, ReplicationEventType.CREATE_WORKSPACE
-            )
-            dual_write_handler.replicate_new_workspace()
-        return workspace
+        if type != Workspace.Types.STANDARD:
+            raise ValidationError({"type": [f"Only workspace type '{Workspace.Types.STANDARD}' is allowed."]})
+
+        if parent_id and tenant:
+            if not Workspace.objects.filter(id=parent_id, tenant=tenant).exists():
+                raise ValidationError({"parent_id": (f"Parent workspace '{parent_id}' does not exist in tenant.")})
```
alechenninger, line 74: My comment on the call about the DB constraint for this did not take into account that this is also searching for a workspace under the same tenant as the request.

I have doubts that that validation should happen here, though. It seems like authorization logic to me: saying that the request is restricted to the request's user's tenant. I suggest keeping permissions rules to the permissions layer. It may be nice to have a "fallback" / defense in depth in a way, but when we use Kessel for authorization, this kind of rule would be restricted to the graph. (And, for example, we will have legitimate use cases for potentially allowing this.)

So we may want to move to the DB-only validation we discussed, though in the grand scheme of things it is not urgent and could be logged in a small tech debt ticket too.

rbac/management/workspace/model.py
```diff
@@ -81,6 +82,8 @@ def clean(self):
                 raise ValidationError({"workspace": f"{self.type} workspaces must have a parent workspace."})
         elif self.type == self.Types.DEFAULT and self.parent.type != self.Types.ROOT:
             raise ValidationError({"default_parent": "Default workspace must have a root parent."})
+        elif self.id == self.parent_id:
+            raise serializers.ValidationError({"parent_id": ("The parent_id and id values must not be the same.")})
```
coderbydesign, line 86: This moves the validation to where it should exist, but does not update the error raised. We'll need to refactor our exception handling to account for model validation errors. This preserves existing behavior, and we should follow-up on refactoring v2 exception handling and problem+json responses.

tests/management/workspace/test_view.py
```diff
@@ -148,6 +150,7 @@ def test_create_workspace_assign_parent_id(self):
         self.assertEquals(data.get("type"), "standard")
         self.assertEqual(response.get("content-type"), "application/json")
 
+    @unittest.skip('v2response_error_from_errors causes this to fail generically with "This field is required"')
```
coderbydesign, line 153: I think we may need to either explicitly raise this still to have a valid error message, leave it like this, or update the exception handler. If we choose anything but the latter, we could address this when we address exception handling.

tests/management/workspace/test_view.py
```diff
@@ -486,7 +487,7 @@ def test_update_workspace_unauthorized(self):
         self.assertEqual(status_code, 403)
         self.assertEqual(response.get("content-type"), "application/problem+json")
 
-    def test_edit_workspace_not_allowed_type(self):
+    def test_edit_workspace_disregard_type(self):
```
coderbydesign, line 490: Proposing we update the intent of this test. As long as we don't publish `type` as a supported field in the spec, we'll just let the serializer (with `read_only=True`) handle this. It won't raise an exception if a value is supplied on create or update, it disregards it, like any other field that's not supported.

tests/management/workspace/test_view.py
```diff
@@ -129,6 +130,7 @@ def test_create_workspace_assign_parent_id(self):
         workspace = {
             "name": "New Workspace",
             "description": "Workspace",
+            "type": "foo",  # should be ignored
```
coderbydesign, line 133: Probably superfluous now with the updated test below.

rbac/management/workspace/view.py
```diff
@@ -103,77 +89,20 @@ def list(self, request, *args, **kwargs):
...
-            self.validate_required_fields(request, REQUIRED_CREATE_FIELDS)
-        elif action == "put":
-            self.validate_required_fields(request, REQUIRED_PUT_FIELDS)
-        if parent_id:
-            validate_uuid(parent_id)
-            if not Workspace.objects.filter(id=parent_id, tenant=tenant).exists():
-                message = f"Parent workspace '{parent_id}' doesn't exist in tenant"
-                error = {"workspace": [message]}
-                raise serializers.ValidationError(error)
+        return super().update(request, *args, **kwargs)
```
coderbydesign, line 108: This allows for unpacking `kwargs` correctly, meaning we can rely on `partial_update=True` being passed to the serializer, meaning we can get rid of the patch serializer.

rbac/management/workspace/serializer.py
```diff
@@ -49,28 +48,32 @@ class Meta:
...
+
+    def validate(self, attrs):
+        from django.core.exceptions import ValidationError
+
+        """Validate on POST, PUT and PATCH."""
+        request = self.context.get("request")
+        parent_id = attrs.get("parent_id")
+        tenant = request.tenant
+
+        if parent_id and tenant:
```
coderbydesign, line 68: Should we leave this for now and add a ticket as @alechenninger mentioned to push this composite validation into the model?

rbac/management/workspace/service.py
```diff
@@ -0,0 +1,69 @@
...
+        self._replicator = replicator
+
+    def create(self, validated_data, tenant) -> Workspace:
+        with transaction.atomic():
+            try:
+                workspace = Workspace.objects.create(**validated_data, tenant=tenant)
+                dual_write_handler = RelationApiDualWriteWorkspacepHandler(
+                    workspace, ReplicationEventType.CREATE_WORKSPACE
+                )
+                dual_write_handler.replicate_new_workspace()
```
alechenninger, line 39: Per discussion, tenant shouldn't be taken directly – it only informs the path where parent_id is not provided.

Example edit:

```python
    def create(self, validated_data, requester_tenant) -> Workspace:
        with transaction.atomic():
            try:
                parent_id = validated_data.get("parent_id")
                if parent_id is None:
                    default = Workspace.objects.default_for_tenant(requester_tenant)
                    parent_id = default.id
                parent = Workspace.objects.get(id=parent_id)
                workspace = Workspace.objects.create(**validated_data, tenant=parent.tenant)
                dual_write_handler = RelationApiDualWriteWorkspacepHandler(
                    workspace, ReplicationEventType.CREATE_WORKSPACE
                )
                dual_write_handler.replicate_new_workspace()
```

rbac/management/workspace/serializer.py
alechenninger: Per discussion, we can get rid of this in favor of 3 things (one will be TODO/new issue):

1. Meta-authz will check a user isn't trying to add a workspace to a parent they can't
2. The service should always pick the right tenant based on the parent workspace (using the user's org's default ws if no parent explicitly provided)
3. (TODO for later) Add a DB constraint that ensures a workspace's parent's tenant is always the same value (or null, to handle the root)

tests/management/workspace/test_view.py
coderbydesign: Per conversation with Alec, we'll just update the assertions here and revisit when addressing the json+problem exception handling.

tests/management/workspace/test_view.py
coderbydesign: @astrozzc

rbac/management/workspace/view.py
```diff
@@ -52,6 +45,11 @@ class WorkspaceViewSet(BaseV2ViewSet):
     ordering = ("name",)
     filter_backends = (filters.DjangoFilterBackend, OrderingFilter)
 
+    def __init__(self, **kwargs):
+        """Init viewset."""
+        super().__init__(**kwargs)
+        self._service = WorkspaceService()
```
astrozzc, line 51: When we initialize it we haven't put any replicator in it. Would it cause issue

tests/management/workspace/test_view.py
astrozzc: Good idea

rbac/management/workspace/view.py
coderbydesign: Updated per conversation to remove from the constructor for now.

lpichler (approved):

tests/management/workspace/test_service.py
```diff
@@ -0,0 +1,103 @@
...
+
+class WorkspaceServiceUpdateTests(WorkspaceServiceTestBase):
+    """Tests for the update method"""
+
+    def test_update_success(self):
+        """Test the update method succeeds"""
+        validated_data = {"name": "Bar Name", "description": "Bar Desc"}
+        updated_instance = self.service.update(self.standard_workspace, validated_data)
+        self.assertEqual(updated_instance.name, validated_data["name"])
+        self.assertEqual(updated_instance.description, validated_data["description"])
```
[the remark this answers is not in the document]
coderbydesign: This will be handled in the TODO ticket.

PR #1669, opened 2025-04-28
add github action that will check the rbac image is created in the quay repository

petracihalova:
github action that will check the quay image is created after PR merge within 30 minutes otherwise the notification will be send in the team's Slack channel


astrozzc (approved):
Nice! Just curious, how it is set to check within 30 mins?
Or it will keep checking until 30 mins maximum?

petracihalova:
> Nice! Just curious, how it is set to check within 30 mins? Or it will keep checking until 30 mins maximum?

it is by default 30 minutes, first it waits 10 minutes and then it makes 20 round of tried per 1 minute ... everything configurable ... https://github.com/petracihalova/check-quay-image

PR #1670, opened 2025-04-28
Fix parent missing when importing groups data

astrozzc:
The parent is missing when bulk importing workspaces

## Summary by Sourcery

Assign the parent workspace during batch import.

Bug Fixes:
- Ensure the parent workspace is correctly assigned when batch importing workspaces using `batch_import_workspace` utility function.

Tests:
- Add assertions to verify the parent workspace is set correctly after batch import.

### DIFF
--- a/rbac/management/management/commands/utils.py
+++ b/rbac/management/management/commands/utils.py
@@ -179,7 +179,7 @@ def batch_import_workspace(records):
             is_ungrouped = record["ungrouped"].lower() == "true"
             parent = parent_workspace_dict[record["org_id"]]
             if is_ungrouped:
-                ws_name = "Ungrouped Hosts"
+                ws_name = Workspace.SpecialNames.UNGROUPED_HOSTS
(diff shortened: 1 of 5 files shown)


rbac/management/management/commands/utils.py
```diff
@@ -178,19 +178,25 @@ def batch_import_workspace(records):
         for record in records:
             is_ungrouped = record["ungrouped"].lower() == "true"
             parent = parent_workspace_dict[record["org_id"]]
-            workspace_type = Workspace.Types.UNGROUPED_HOSTS if is_ungrouped else Workspace.Types.STANDARD
+            if is_ungrouped:
+                ws_name = "Ungrouped Hosts"
```
vbelchio, line 182: Maybe we can fix the "Ungrouped Hosts" workspace name in a variable?

astrozzc: Good idea

PR #1695, opened 2025-05-07
[RHCLOUD-39821] update open api specification for resourceDefinition object

petracihalova:
[RHCLOUD-39821](https://issues.redhat.com/browse/RHCLOUD-39821)

related with https://github.com/RedHatInsights/insights-rbac/pull/1418#issuecomment-2569611683

update the attribute filter schema to use conditional blocks
- if `operation = equal` then `value = type string`
- if `operation = in` then `value = type array / list`

## Summary by Sourcery

Documentation:
- Modify OpenAPI specification to add conditional validation for attribute filter values based on operation type

### DIFF
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -66,7 +66,9 @@
             "description": "Parameter for ordering resource by database id so that latest actions appear first",
             "schema": {
               "type": "string",
-              "enum": ["id"]
+              "enum": [
+                "id"
+              ]
             }
           }
         ],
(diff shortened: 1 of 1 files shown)


vbelchio (approved):
QE: Passed with some reservations.

Currently, the iqe_rbac_plugin is only compatible with OpenAPI Generator 4.3.1 and the changes are not accepted by this version. But, also there are no specific tests considering the changes in the plugin.
Using the latest version of OpenAPI Generator 7.13.0, the changes are accepted and working properly.
Concluding, LGTM.

PR #1727, opened 2025-05-22
[RHCLOUD-40233] add service account into a group - fix the status code for not existing uuid

petracihalova:
[RHCLOUD-40233](https://issues.redhat.com/browse/RHCLOUD-40233)

query bellow returns now 400

this PR will change it to return 404

```
POST /api/rbac/v1/groups/<group_id>/principals/
{
    "principals": [
        {
            "clientId": "66a9986a-a265-4a11-88e8-d853a0a557f4",
            "type": "service-account"
        }
    ]
}
```
where client id is valid but not existing uuid

## Summary by Sourcery

Return 404 Not Found when attempting to add a non-existent service account to a group and align error response formatting and tests accordingly.

Bug Fixes:
- Change ServiceAccountNotFoundError handler to return HTTP 404 instead of 400.
- Standardize forbidden error responses to use string status codes.

Enhancements:
- Correct error payload source field from "group" to "groups" in service-account errors.

Tests:
- Add test to verify adding a non-existent service account returns 404.

### DIFF
--- a/rbac/management/group/view.py
+++ b/rbac/management/group/view.py
@@ -833,15 +833,15 @@ class GroupViewSet(
                             ]
                         },
                     )
-                except ServiceAccountNotFoundError as sanfe:
+                except ServiceAccountNotFoundError as err:
                     return Response(
-                        status=status.HTTP_400_BAD_REQUEST,
+                        status=status.HTTP_404_NOT_FOUND,
                         data={
                             "errors": [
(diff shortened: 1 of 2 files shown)


Ellen-Yi-Dong (approved):
LGTM!

vbelchio:
QE: Passed (manual)

PR #1744, opened 2025-06-04
update the redis package to the version 6.0.0

petracihalova:
https://pypi.org/project/redis/

## Summary by Sourcery

Enhancements:
- Bump Redis package from 5.3.0 to 6.0.0 in Pipfile and update Pipfile.lock accordingly


petracihalova:
> Would this major version bump cause any issue? Seems we are fine, but better to keep an eye on the stage cache usage

yes agree, the [latest](https://pypi.org/project/redis/) redis version is 6.2.0 but I would like to do small steps

PR #1745, opened 2025-06-04
[RHCLOUD-40411] fix unhandled ValidationError when updating workspace with duplicate name (PUT/PATCH)

petracihalova:
## Link(s) to Jira
- [RHCLOUD-40411](https://issues.redhat.com/browse/RHCLOUD-40411)

previously, if a workspace name was updated to one that already exists under the same parent, the `update()` method raised an unhandled `ValidationError`, resulting in a` 500 Internal Server Error response`

```
PUT /api/rbac/v2/workspaces/{{workspace_id}}/
PATCH /api/rbac/v2/workspaces/{{workspace_id}}/

{
  "name": <existing_name>
}
```

although a db constraint is already in place to prevent duplicate names, the error was not properly caught

this PR introduces
- handling of `ValidationError` during `update()` (mirroring the existing behavior in `create()`)
- wrapping database operations in `transaction.atomic()` to ensure data consistency and enable rollback on failure

**note**: it is still valid and allowed to update a workspace name to one that already exists within the same tenant as long as the parent workspace is different -> the uniqueness constraint applies per parent, not globally within the tenant

## Local Testing

- create 2 workspaces with different names under same parent workspace
- try to update the workspace 1 with the name of workspace 2

example of expected response after change
```
{
    "status": 400,
    "detail": "A workspace with the name 'Workspace 2' already exists under same parent.",
    "instance": "/api/rbac/v2/workspaces/01970c97-89fe-7ab2-81a2-2f2757886381/"
}
```

## Summary by Sourcery

Catch and handle duplicate-name ValidationError during workspace updates to return a 400 response, wrap update operations in a transaction, and add tests and tox adjustments.

Bug Fixes:
- Return HTTP 400 with a descriptive message when updating a workspace to a duplicate name under the same parent
- Handle ValidationError in the update() method to prevent 500 errors on name conflicts

Enhancements:
- Wrap workspace update and dual-write operations in transaction.atomic() for rollback on failure

Build:
- Restrict tox test suite to workspace management tests

Tests:
- Add API tests for PUT and PATCH endpoints to verify 400 error on duplicate workspace names
- Add service-layer test to ensure ValidationError is raised for name uniqueness violations

### DIFF
--- a/rbac/internal/views.py
+++ b/rbac/internal/views.py
@@ -95,15 +95,7 @@ jwt_manager = JWTManager(jwt_provider, jwt_cache)
 @contextmanager
 def create_client_channel(addr):
     """Create secure channel for grpc requests."""
-    # Call credential object will be invoked for every single RPC
-    token = jwt_manager.get_jwt_from_redis()
-
-    call_credentials = grpc.access_token_call_credentials(token)
-
-    combined_credentials = grpc.composite_channel_credentials(grpc.local_channel_credentials(), call_credentials)
-
-    secure_channel = grpc.secure_channel(addr, combined_credentials)
-
+    secure_channel = grpc.insecure_channel(addr)
     yield secure_channel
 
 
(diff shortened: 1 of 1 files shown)


petracihalova:
> I wonder if we can put these into decorator. to reuse the kind of structure to put things into transaction and error handling

❤️ awesome idea I like it

PR #1753, opened 2025-06-04
[RHCLOUD-35765] Raise error when nonexistent role is added to group

Ellen-Yi-Dong:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-35765

The what, why and how.

When a non-existent role is added to a group, a 200 status code is returned instead of a 404 status error code. 

## Local Testing

You can test locally by using this curl command: 
```
curl --location 'localhost:8000/api/rbac/v1/groups/<random-group-uuid>/roles/' \
--header 'Content-Type: application/json' \
--header 'Content-Type: application/json' \
--header 'x-rh-identity: <x-rh-identity>' \
--data '{
    "roles": [
"<random-uuid-created>"
    ]
}
'
```

And it should return an error saying: 
```
{
    "errors": [
        {
            "detail": "This role is nonexistent/nonvalid and cannot be added to the group",
            "source": "detail",
            "status": "404"
        }
    ]
}
```

I also updated the unit tests to reflect the new change to have 404 errors. 

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

## Summary by Sourcery

Fail fast when invalid roles are added to a group and log skipped duplicates

Bug Fixes:
- Raise Http404 when no matching roles are found for addition to a group

Enhancements:
- Log an info message when attempting to re-add an already assigned role to a group


rbac/management/group/definer.py
```diff
@@ -234,7 +240,11 @@ def update_group_roles(group, roleset, tenant):
 def _roles_by_query_or_ids(roles_or_role_ids: Union[QuerySet[Role], list[str]]) -> QuerySet[Role]:
     if not isinstance(roles_or_role_ids, QuerySet):
         # If given an iterable of UUIDs, get the corresponding objects
-        return Role.objects.filter(uuid__in=roles_or_role_ids)
+        filtered_roles = Role.objects.filter(uuid__in=roles_or_role_ids)
```
astrozzc, line 243: This is within if statement, the else case is missing.

astrozzc (commented):
So only when all of the roles are not available, error will be thrown?

rbac/management/group/definer.py
Ellen-Yi-Dong: > So only when all of the roles are not available, error will be thrown?
Yesh! Only when the role is invalid/non-existent will an error be thrown! If there is a group of roles that includes both existent and non-existent, there will be no error thrown, because the non-existent role will not be added to the group.

Ellen-Yi-Dong: > This is within if statement, the else case is missing.


[2025-06-17 18:44:39,473] ERROR: Error encountered during group seeding This role is nonexistent/nonvalid and cannot be added to the group.

^^^When I place a similar statement within the else case, there are errors right when it starts with the seeding. I was thinking that since when a role is being added to a group, only the UUID is valid through the body of data.

astrozzc: Could you investigate why this error will be thrown? Because it is complaining that roles added to a group all do not exist, I wonder how could it happen for a seeding process

astrozzc: So this sounds like what we need, right? throwing 404 if the list is empty. Maybe we should add it and update the tests?

Ellen-Yi-Dong: That sounds good to me, and I can add it! 
I was a bit unsure of whether the behavior should be normalized for the seeding so that's why I was hesitant to add it :sweat

astrozzc: Ah, I see. Yea, I wonder if the seeding would encounter this issue because it is supposed to add roles that's all exist. Maybe better to throw an error in this case so we know it is unexpected instead of silently skip it 🤔

astrozzc (approved):

rbac/management/group/definer.py
Ellen-Yi-Dong: Do you think maybe throwing a warning would suffice? 
I think with throwing an error, for the unit tests, the issue also comes up when the roles are trying to be "set up". The tests start to fail whenever the tests are being set up for the tests relating to the cross_access/test_util.py

vbelchio:
@Ellen-Yi-Dong 
Maybe we need to check if the Role exists for the specific tenant.
I tried to add a Role from tenant A to a group from tenant B, and I got 200, but the role was not added to the group. This case needs to result in a 404 as well once the Role does not exist for the tenant B.

vbelchio:
QE: Passed

PR #1773, opened 2025-06-12
[RHCLOUD-40423] Skip workspace events for DEFAULT workspace

petracihalova:
## Link(s) to Jira
- [RHCLOUD-40423](https://issues.redhat.com/browse/RHCLOUD-40423)

it is not need to sent workspace events for default workspace when default workspace is updated, because default workspace doesn't exist in HBI side

## Local Testing
try to update name of standard workspace 
```
PUT /api/rbac/v2/workspaces/{{workspace_id}}/ 
{
  "name": "New name"
}
```
and check 
- the workspace is updated
- the workspace event is created 

then try to update name of default workspace and check
- the workspace is updated
- the workspace event is not created


lpichler (approved):

lpichler:
Tested it works well 👍

PR #1804, opened 2025-06-27
Ensure `username` is set on system-user, generated from JWT

coderbydesign:
In the middleware, in order to set the `request.user = user` we first check to ensure there's a `username` and `user.org_id` on the `user` object [1].

This currently causes JWT system users to be created without a `username` since it's based on `preferred_username` [2] which does not exist on the client service account tokens. This ultimately means `request.user` is not the custom user object we've created, but an anonymous user [3] which does not have the fields required to fulfill requests [4] resulting in 500s.

This adds the client/user id value as the username in the builder logic.

[1] https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/rbac/middleware.py#L304-L305
[2] https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/authorization/token_validator.py#L63
[3] https://docs.djangoproject.com/en/4.2/ref/contrib/auth/
[4] https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/permissions/workspace_access.py#L28

## Local Testing

1. Set the following in your env, where `client-id` is from the SSO client you'll be generating a JWT for:

```bash
DEVELOPMENT=False
IT_SERVICE_HOST=sso.stage.redhat.com
SYSTEM_USERS={"<client-id>": {"admin": True, "is_service_account": True, "allow_any_org": True}}
```
2. Generate a JWT for your client

3. Prior to this change, the following will fail, but should work on this branch:
```
curl http://localhost:8000/api/rbac/v2/workspaces/ -H "Authorization: Bearer ${TOKEN}" -H 'x-rh-rbac-org-id: <local-org-id>'
```

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

## Summary by Sourcery

Assign username for JWT-derived system users and add a unit test for the new behavior

Bug Fixes:
- Assign user.username to user_id when building system users from JWT tokens to prevent anonymous user errors

Tests:
- Add a unit test for build_system_user_from_token to verify username, org ID, account, and service account flags


rbac/management/utils.py
```diff
@@ -103,6 +103,7 @@ def build_system_user_from_token(request, token_validator: TokenValidator) -> Op
         system_users: dict[str, SystemUserConfig] = settings.SYSTEM_USERS
         if user and user.user_id in system_users:
             system_user = system_users[user.user_id]
+            user.username = user.user_id
```
alechenninger, line 106: Should we use the preferred_username claim?

alechenninger: Oh hrm I'm guessing service account tokens don't have this claim set?

alechenninger: We might want to only set username if it is not already set then.

alechenninger: I swear I read that! 😅 

Yeah it looks like third party org service accounts do have this claim, but not centrally managed / first party ones.

coderbydesign: Got it, ok! Thanks for the context. Updated with a condition for that.

astrozzc (approved):
LGTM

PR #1805, opened 2025-06-27
[RHCLOUD-40741] Update audit logs to have UUID

Ellen-Yi-Dong:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-40741

The what, why and how.
Similarly mentioned within the JIRA ticket, update the audit log to use uuid. Each of the audit logs should have a unique UUID. 

After the new migrations, the new database should appear with these new headers. 
Screenshot From 2025-06-27 15-05-14

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

## Summary by Sourcery

Add a unique UUID field to each audit log entry and populate existing records with generated UUIDs via new database migrations.

New Features:
- Introduce a UUID field on the AuditLog model with a unique constraint.

Enhancements:
- Provide migrations to add the UUID field, backfill existing audit logs with generated UUIDs, and enforce uniqueness on the new field.


rbac/management/migrations/0070_populate_new_uuid_field_auditlog.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 4.2.23 on 2025-06-27 21:31
+import uuid
+from django.db import migrations
+
+
+def populate_new_uuid_field(apps, schema_editor):
```
lpichler, line 6: do we need it as there is used "default=uuid4" ?

Ellen-Yi-Dong:
> @Ellen-Yi-Dong we also need to store uuids into this column from resources (groups, user,.. ) for audit log records. Do you plan to do it in this PR ?

Hello @lpichler! 
I was thinking at first to have two different PRs, but I think that I can add it to this PR!

rbac/management/migrations/0070_populate_new_uuid_field_auditlog.py
Ellen-Yi-Dong: I think we do, because even though we have it as default, that's just adding the field, and I'm thinking of populating the previous entries with the new UUID field and applying a new UUID to each audit log!

astrozzc (commented):
Could you merge the migrations? You can delete the automatically made up ones and run makemigrations again

rbac/management/migrations/0069_auditlog_uuid.py
```diff
@@ -0,0 +1,76 @@
...
+def populate_new_uuid_field(apps, schema_editor):
+    AuditLog = apps.get_model("management", "auditlog")
+    for obj in AuditLog.objects.all():
+        obj.uuid = uuid.uuid4()
+        obj.save(update_fields=["uuid"])
+
+
+def populate_resource_uuid_field(apps, schema_editor):
+    AuditLog = apps.get_model("management", "auditlog")
+    for obj in AuditLog.objects.all():
```
astrozzc, line 20: I am worried about the performance here. How many audit logs records do we have now? The iterator might hang and migration might fail. Maybe we can ignore the previous records, just leave them as they are, seems not that critical? @lpichler 

BTW, seems you forgot to call obj.save(), the update won't be saved

PR #1956, opened 2025-09-09
RHCLOUD-35481 prevent custom role permissions for resources outside of the users org

dagbay-rh:
## Link(s) to Jira
- https://issues.redhat.com/browse/RHCLOUD-35481

Add logic to the create/update roles endpoint and migrator to validate that a user cannot add permissions to a role in a workspace that is outside of their org

tests were written with assistance from claude
## Summary by Sourcery

Prevent custom role permissions from being applied to workspaces outside the user’s organization

New Features:
- Add utility functions to detect workspace resources and extract workspace IDs from resource definitions

Bug Fixes:
- Skip migrating v1 permissions targeting workspaces outside the user’s org
- Reject role validation requests that add permissions to workspaces outside the user’s org

Enhancements:
- Refactor workspace filter check in serializers to use the new utility function

## Summary by Sourcery

Enforce tenant-based scoping on workspace permissions by adding validation in the API and migration tool, refactor common workspace detection into utilities, and add comprehensive tests

New Features:
- Prevent roles from including workspace permissions outside the user’s organization in RoleViewSet
- Skip migrating v1 workspace permissions that target workspaces outside the role’s tenant in the migration tool

Enhancements:
- Introduce utility functions to identify workspace resource definitions and extract workspace IDs
- Refactor AccessSerializer to leverage the new workspace detection utility

Tests:
- Add extensive unit tests for workspace validation covering same-tenant and cross-tenant scenarios, missing filters, non-workspace resources, and utility function usage


rbac/management/role/view.py
```diff
@@ -639,6 +641,20 @@ def validate_role(self, request):
                         error = {key: [_(message)]}
                         raise serializers.ValidationError(error)
 
+                # validate permission not being added to workspace out of users org for v1 (RHCLOUD-35481)
+                resourceDefinitions = perm.get("resourceDefinitions")
+                for resourceDefinition in resourceDefinitions:
+                    attributeFilter = resourceDefinition.get("attributeFilter")
+                    if is_resource_a_workspace(app, resource_type, attributeFilter):
+                        workspace_id = get_workspace_id_from_resource_definition(attributeFilter)
```
astrozzc, line 649: If the value is a list, we have to iterate through all the values instead of getting the first one

rbac/migration_tool/sharedSystemRolesReplicatedRoleBindings.py
```diff
@@ -118,6 +122,17 @@ def v1_role_to_v2_bindings(
                     # Skip empty values
                     continue
 
+            # validate permission was not added to workspace out of users org for v1 (RHCLOUD-35481)
+            if is_resource_a_workspace(v1_perm.application, v1_perm.resource_type, attri_filter):
+                workspace_id = get_workspace_id_from_resource_definition(attri_filter)
+                is_same_tenant = Workspace.objects.filter(id=workspace_id, tenant=v1_role.tenant)
```
astrozzc, line 128: .exists() 😉

rbac/management/role/view.py
```diff
@@ -639,6 +641,20 @@ def validate_role(self, request):
...
+                        is_same_tenant = Workspace.objects.filter(id=workspace_id, tenant=request.tenant).exists()
```
coderbydesign, line 650: Yeah, I wonder if it's worth just having this always return a list, and doing a check like:
```suggestion
                        workspace_ids = get_workspace_ids_from_resource_definition(attributeFilter)
                        is_same_tenant = Workspace.objects.filter(id__in=workspace_ids, tenant=request.tenant).exists()
```

Where then `get_workspace_ids_from_resource_definition` then always returns a list regardless of what's passed in?

rbac/management/role/view.py
```diff
@@ -639,6 +641,20 @@ def validate_role(self, request):
                         error = {key: [_(message)]}
                         raise serializers.ValidationError(error)
 
+                # validate permission not being added to workspace out of users org for v1 (RHCLOUD-35481)
+                resourceDefinitions = perm.get("resourceDefinitions")
```
coderbydesign, line 645: I think we do explicitly require `resourceDefinitions` on the payload, but would it be worth adding a default here so the `for` loop below is guarded?
```suggestion
                resourceDefinitions = perm.get("resourceDefinitions", [])
```

rbac/management/role/view.py
dagbay-rh: thats the change I made :)

rbac/migration_tool/sharedSystemRolesReplicatedRoleBindings.py
dagbay-rh: - [ ] add missing exists() check

rbac/management/role/view.py
dagbay-rh: - [ ] check if value is list

rbac/management/role/view.py
dagbay-rh: - [ ] add default for list

lpichler:
@dagbay-rh @astrozzc did we check that existing data don't have already UUID which are out of orgs ?

astrozzc:
> @dagbay-rh @astrozzc did we check that existing data don't have already UUID which are out of orgs ?

Good question! I thought I asked about it but I missed it😅 
We can run a query for it in both stage and prod @dagbay-rh

PR #2692, opened 2026-04-01
fix(roles): block system role name reuse

Hyperkid123:
RHCLOUD-37254

Prevents creating or updating custom roles with names (or display names) that match existing system/seeded roles, using case-insensitive comparison.

**Problem:** The API allowed users to create custom roles with names identical to system roles (e.g. "Vulnerability viewer"), causing ambiguity and potential security risks. The UI already blocked this, but the API did not.

**Changes:**

- **V1 serializer** (`serializer.py`): Added `_validate_name_not_system_role()` helper that checks both `name` and `display_name` against system roles (where `system=True` in the public tenant). Called from `RoleSerializer.validate()` and `RolePatchSerializer.validate()`.
- **V2 serializer** (`v2_serializer.py`): Extended `validate_name()` in `RoleV2RequestSerializer` to check against seeded/platform role names in the public tenant.
- Both validations skip the check when the name hasn't changed on update.
- Returns 400 with a clear error message: `"Role name '<name>' conflicts with an existing system role."`

**Tests added (6):**
- V1: create with system name, create with system display_name, update to system name, patch to system name
- V2: create with seeded name, update to seeded name

Related: [RHCLOUD-36017](https://redhat.atlassian.net/browse/RHCLOUD-36017)

[RHCLOUD-36017]: https://redhat.atlassian.net/browse/RHCLOUD-36017?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Disallow creating or updating custom roles whose names or display names conflict with existing system/seeded roles using case-insensitive comparison.

Bug Fixes:
- Prevent custom roles in both v1 and v2 APIs from reusing names or display names of existing system/seeded roles, eliminating ambiguity and potential security issues.

Tests:
- Add v1 tests covering create, update, and patch operations failing when role name or display_name matches a system role.
- Add v2 tests verifying create and update operations fail when the role name matches a seeded role name in the public tenant.

### DIFF
--- a/rbac/management/role/serializer.py
+++ b/rbac/management/role/serializer.py
@@ -259,6 +259,7 @@ class RoleSerializer(serializers.ModelSerializer):
             message = "System roles may not be updated."
             error = {key: [_(message)]}
             raise serializers.ValidationError(error)
+        _validate_name_not_system_role(data.get("name"), data.get("display_name"), self.instance)
         return super().validate(data)
 
 
@@ -418,6 +419,7 @@ class RolePatchSerializer(RoleSerializer):
             message = "System roles may not be updated."
             error = {key: [_(message)]}
             raise serializers.ValidationError(error)
+        _validate_name_not_system_role(data.get("name"), data.get("display_name"), self.instance)
         return super().validate(data)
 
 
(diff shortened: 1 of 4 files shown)


Hyperkid123 (commented):
Thanks for the review! Addressed the i18n feedback:

- **String formatting**: Replaced f-strings with `%`-formatting in both V1 and V2 validation error messages so they're properly extractable by gettext tools.
- **V2 gettext**: Added `gettext` import and wrapped the V2 error message in `_()` for consistent i18n behavior.
- **Public tenant lookup**: Kept `Tenant.objects.get(tenant_name="public")` as-is — this is the established pattern used in 70+ places across the codebase. Introducing a helper here would be inconsistent with the rest of the code.

rbac/management/role/serializer.py
```diff
@@ -418,6 +419,7 @@ def validate(self, data):
             message = "System roles may not be updated."
             error = {key: [_(message)]}
             raise serializers.ValidationError(error)
+        _validate_name_not_system_role(data.get("name"), data.get("display_name"), self.instance)
```
lpichler: RolePatchSerializer.validate() calls _validate_name_not_system_role() and then calls super().validate(data), which is RoleSerializer.validate(), which also calls _validate_name_not_system_role(). The validation runs twice for PATCH operations. This is harmless (idempotent) but wastes a DB query.

Hyperkid123: @platex-rehor-bot are you sure the Pr is correct? The commit you referenced is not form this branch.

```
This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository.
```

karelhala:
@Hyperkid123 Good catch! The commit was pushed to the fork but not to the upstream branch — the PR was tracking the branch on `RedHatInsights/insights-rbac`, not the fork. I've now pushed it to the correct branch. The PR should show all 5 commits including 409ba8f9 which removes the duplicate `validate()` override from `RolePatchSerializer`.

Hyperkid123:
@platex-rehor-bot can you validate that the changes you made are in this PR? I think the change in identity messed this up. If not, feel free to re-create the PR on your fork, please. And update relevant references.

PR #2698, opened 2026-04-02
[RHCLOUD-46260] fix scope change issue after seeding

Ellen-Yi-Dong:
## Link(s) to Jira
- https://redhat.atlassian.net/browse/RHCLOUD-46260

The what, why and how.

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

## Summary by Sourcery

Ensure system role bindings are updated when a role’s scope changes during seeding.

Bug Fixes:
- Update existing system role bindings to the correct resource (tenant vs workspace) when a role’s scope changes between seeding runs.

Enhancements:
- Detect scope changes for existing seeded V2 roles and trigger binding migration using existing migration utilities.

Tests:
- Add integration-style test verifying that changing a system role from DEFAULT to TENANT scope during reseeding updates existing bindings from workspace to tenant.


rbac/management/role/definer.py
```diff
@@ -48,6 +48,110 @@
...
+    # Calculate scope from existing permissions
+    if resource_service:
+        permission_strings = list(existing_v2_role.permissions.values_list("permission", flat=True))
+        if permission_strings:
+            try:
+                return resource_service.highest_scope_for_permissions(permission_strings)
+            except Exception as e:
+                logger.warning(f"Failed to calculate scope from permissions for role {existing_v2_role.uuid}: {e}")
+
+    return None
```
randomnetcat: Unfortunately, this is not always sufficient; see `admin_platform_parent_scope_for_seeded_system_role`.

rbac/management/role/definer.py
```diff
@@ -48,6 +48,110 @@
...
+            if result > 0:
+                migrated_count += 1
+                logger.info(
+                    f"Migrated bindings for group {group.name} (uuid={group.uuid}) " f"with system role {v1_role.name}"
+                )
+        except Exception as e:
+            logger.error(
+                f"Failed to migrate bindings for group {group.uuid} with role {v1_role.name}: {e}",
+                exc_info=True,
+            )
```
randomnetcat, line 296: This needs to (a) lock each group before migrating it and (b) in the same transaction as that, check that the tenant is a V1 tenant (see `lock_tenant_version` and/or `assert_v1_write_allowed`).

It's likely we should just do that all in the public function to prevent misuse. Let me put up a quick PR for that.

tests/management/role/test_seeding_binding_scope_update.py
```diff
@@ -0,0 +1,325 @@
...
+    def test_binding_scope_updates_when_role_scope_changes_during_seeding(self, mock_replicate, mock_migrate_bindings):
+        """Test that scope change detection and migration work during seeding."""
+        # Redirect replicator
+        mock_replicate.side_effect = self.replicator.replicate
+
+        # Seed roles with DEFAULT scope for most permissions
+        with self.settings(
+            ROOT_SCOPE_PERMISSIONS="",
+            TENANT_SCOPE_PERMISSIONS="",
+            DEFAULT_SCOPE_PERMISSIONS="inventory:*:*,approval:*:*,notifications:*:*",
```
randomnetcat: This setting doesn't exist. Anything that isn't set to root or tenant scope by the other settings is assumed to be in the default scope.

tests/management/role/test_seeding_binding_scope_update.py
```diff
@@ -0,0 +1,325 @@
...
+            seed_group()
+            seed_roles()
+
+        # Find roles that should be at DEFAULT scope
+        roles_with_inventory = list(
+            Role.objects.filter(
+                system=True,
+                tenant=self.public_tenant,
+                access__permission__application="inventory",
+                platform_default=True,  # Only platform_default roles to avoid admin_default override complexity
```
randomnetcat: If you want to avoid the override, you either need to change the `ADMIN_DEFAULT_SEEDED_ROLES_FORCE_ROOT_SCOPE` or explicitly set `admin_default=False` here, since a role can be both platform-defauld and admin-default.

tests/management/role/test_seeding_binding_scope_update.py
```diff
@@ -0,0 +1,325 @@
...
+            ).distinct()
+        )
+
+        self.assertGreater(len(roles_with_inventory), 0, "Should find platform_default roles with inventory perms")
+
+        # Pick one role to test with
+        test_role = roles_with_inventory[0]
```
randomnetcat: Might as well just use `first()` and check for `None`, then?

tests/management/role/test_definer.py
```diff
@@ -1228,6 +1228,43 @@ def test_non_default_role_has_no_parent(self):
...
+                # Use a wildcard pattern that matches the role's permissions
+                permission_pattern = f"{role_permissions[0].split(':')[0]}:*:*"
+
+                # Re-seed with the permission now in ROOT scope
+                with self.settings(ROOT_SCOPE_PERMISSIONS=permission_pattern, TENANT_SCOPE_PERMISSIONS=""):
+                    seed_roles()
+
+                # Verify that migration was triggered for the scope change
+                # The migration should be called with the v1 role, old scope (DEFAULT), and new scope (ROOT)
+                if mock_migrate.called:
```
randomnetcat: Shouldn't we know whether it was called?

tests/management/role/test_seeding_binding_scope_update.py
```diff
@@ -0,0 +1,325 @@
...
+            TENANT_SCOPE_PERMISSIONS="inventory:*:*",
+            DEFAULT_SCOPE_PERMISSIONS="",
+        ):
+            seed_roles(force_update_relationships=True)
+
+        # The test goal: verify that when scopes change, migration is triggered
+        # We check if migration was called for ANY role with inventory permissions
+        # (the exact role and scopes may vary based on definition files)
+
+        if mock_migrate_bindings.call_count > 0:
```
randomnetcat: Shouldn't we know whether it was called? We changed from all permissions having default scope to having all `inventory` permissions (which we know the role has) having tenant scope, so the role should have changed scope, right?

rbac/management/role/definer.py
```diff
@@ -48,6 +48,110 @@
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
 
+def _determine_old_scope(existing_v2_role, platform_roles, resource_service=None):
+    """
+    Determine the scope of an existing V2 role by calculating it from the role's permissions.
+
+    Args:
+        existing_v2_role: The existing SeededRoleV2 instance (or None)
+        platform_roles: Dictionary mapping (DefaultAccessType, Scope) to platform roles (unused, kept for compatibility)
```
randomnetcat: Compatibility with what?

tests/management/role/test_seeding_binding_scope_update.py
```diff
@@ -0,0 +1,325 @@
...
+        mock_role = MagicMock()
+        mock_role.parents.values_list.return_value = [tenant_admin_uuid]
+        result = _determine_old_scope(mock_role, platform_roles)
+        self.assertEqual(result, Scope.TENANT, "Should detect TENANT scope from ADMIN parent UUID")
+
+        # Test 3: Role with no matching parents
+        mock_role = MagicMock()
+        mock_role.parents.values_list.return_value = [uuid4()]  # Random UUID
+        result = _determine_old_scope(mock_role, platform_roles)
+        self.assertIsNone(result, "Should return None when no parent matches")
```
randomnetcat: I'm not sure what we're testing here? `_determine_old_scope` doesn't appear to actually use `platform_roles`?

tests/management/role/test_seeding_binding_scope_update.py
```diff
@@ -0,0 +1,325 @@
...
+        mock_role.permissions.values_list.return_value = ["inventory:hosts:read", "inventory:hosts:write"]
+
+        # Create a mock resource service
+        mock_resource_service = MagicMock(spec=ImplicitResourceService)
+        mock_resource_service.highest_scope_for_permissions.return_value = Scope.DEFAULT
+
+        # Test that scope is determined from permissions
+        result = _determine_old_scope(mock_role, {}, mock_resource_service)
+        self.assertEqual(result, Scope.DEFAULT, "Should detect DEFAULT scope from permissions")
+
```
randomnetcat: We should probably just test it by updating `ROOT_SCOPE_PERMISSIONS` and `TENANT_SCOPE_PERMISSIONS`? (We could also just not test it, since it's an internal helper function.)

tests/management/role/test_seeding_binding_scope_update.py
```diff
@@ -0,0 +1,325 @@
...
+from management.tenant_mapping.model import DefaultAccessType
+from migration_tool.in_memory_tuples import InMemoryTuples, InMemoryRelationReplicator
+from tests.identity_request import IdentityRequest
+from tests.v2_util import bootstrap_tenant_for_v2_test
+
+from api.models import Tenant
+
+
+@override_settings(ATOMIC_RETRY_DISABLED=True, REPLICATION_TO_RELATION_ENABLED=True)
+class SystemRoleBindingScopeUpdateTests(IdentityRequest):
```
randomnetcat: This appears to test a lot of internal logic but not actually test what relations are added/removed? We should probably do that at least to some extent. (We should also add a test that checks what happens for V2 tenants.)

rbac/management/role/definer.py
randomnetcat: #2733

astrozzc:
oh, sorry, I just thought of a last question: if a system role is bound to a subworkspace on purpose, when the scope of the role is changed, would it update the binding also? that would be problematic

rbac/management/role/definer.py
```diff
@@ -48,6 +52,141 @@
...
+    Returns:
+        The Scope of the role, or None if scope cannot be determined
+    """
+    if existing_v2_role is None:
+        return None
+
+    # Get the UUIDs of parent platform roles
+    parent_uuids = set(existing_v2_role.parents.values_list("uuid", flat=True))
+    if not parent_uuids:
+        return None
```
randomnetcat: This doesn't solve the [issue that JayZ pointed out](https://github.com/RedHatInsights/insights-rbac/pull/2698/changes#r3033394441). If we want to do this, I think we're just not going to get away from having to cache the old scope somewhere.

tests/management/role/test_definer.py
```diff
@@ -1228,6 +1228,47 @@ def test_non_default_role_has_no_parent(self):
...
+        non_default_role = (
+            Role.objects.public_tenant_only().filter(platform_default=False, admin_default=False, system=True).first()
+        )
+        self.assertIsNotNone(non_default_role, "Should have at least one seeded non-default role")
+
+        # Mock the migration function
+        with patch("management.role.definer._migrate_bindings_for_scope_change") as mock_migrate:
+            # Change the scope for this role's permissions
+            role_permissions = list(non_default_role.access.values_list("permission__permission", flat=True))
+            if role_permissions:
```
randomnetcat: This risks not actually testing anything. We should either just pick a role by name from the test seed data or filter for roles with permissions above.

tests/management/role/test_seeding_binding_scope_update.py
```diff
@@ -0,0 +1,386 @@
...
+        test_role = Role.objects.filter(
+            system=True,
+            tenant=self.public_tenant,
+            access__permission__application="inventory",
+            platform_default=True,
+            admin_default=False,  # Avoid roles subject to ROOT scope override
+        ).first()
+
+        if not test_role:
+            self.skipTest("No suitable platform_default (non-admin_default) roles with inventory permissions found")
```
randomnetcat: This should likely fail here; we should know that such a role exists in our test data.

tests/management/role/test_seeding_binding_scope_update.py
```diff
@@ -0,0 +1,386 @@
...
+        group = Group.objects.create(name="Test Group", tenant=self.tenant, system=False)
+        add_roles(group, [test_role.uuid], self.tenant)
+
+        # Verify initial binding at DEFAULT workspace
+        v2_role = SeededRoleV2.objects.get(uuid=test_role.uuid)
+        initial_bindings = self.tuples.read_tuples(
+            resource_type="rbac/principal",
+            resource_id=f"group:{group.uuid}",
+            relation="binding",
+        )
```
randomnetcat: This function doesn't appear to exist, unless I'm missing something obvious? And `resource_id` wouldn't have a `group:` prefix regardless.

tests/management/role/test_seeding_binding_scope_update.py
```diff
@@ -0,0 +1,386 @@
...
+
+        # Should no longer have binding at default workspace
+        default_workspace_bindings_after = [
+            t
+            for t in final_bindings
+            if t.subject.object.object_id == str(self.default_workspace.id)
+            and str(v2_role.uuid) in t.subject.object.object_type
```
randomnetcat: Some of these fields don't exist.

tests/management/role/test_seeding_binding_scope_update.py
```diff
@@ -0,0 +1,386 @@
...
+            account_id="9999999",
+            org_id="9999999",
+            ready=True,
+        )
+        v2_tenant.save()
+
+        # Bootstrap V2 tenant with workspaces
+        from tests.v2_util import bootstrap_tenant_for_v2_test
+
+        bootstrap_result = bootstrap_tenant_for_v2_test(v2_tenant, tuples=self.tuples)
```
randomnetcat: This doesn't make the tenant a V2 tenant. See `ensure_v2_write_activated`.

tests/management/role/test_seeding_binding_scope_update.py
randomnetcat: Nevermind, we apparently did at it. Sorry!

rbac/management/role/definer.py
```diff
@@ -48,6 +52,230 @@
...
+                # Calculate scope from the existing permissions using the highest scope
+                calculated_scope = resource_service.highest_scope_for_permissions(existing_permissions)
+                logger.debug(
+                    "Calculated old scope for %s from %d existing V2 permissions: %s "
+                    "(note: uses current mapping, may not detect global mapping changes)",
+                    existing_v2_role.name,
+                    len(existing_permissions),
+                    calculated_scope.name if calculated_scope else None,
+                )
+                return calculated_scope
```
astrozzc: So if a role has neither platform/admin default parent, and its permission is changed from ROOT_SCOPE_PERMISSIONS to  TENANT_SCOPE_PERMISSIONS, we can't detect the scope change.
Otherwise lgtm

rbac/management/role/definer.py
```diff
@@ -48,6 +52,230 @@
...
+        except Exception:
+            logger.debug(
+                "Failed to calculate old scope from existing V2 permissions for %s",
+                existing_v2_role.name,
+                exc_info=True,
+            )
+
+    return None
```
randomnetcat: Okay! I think I have a better solution for determining the old scope without caching.

We can check what resources the role is actually bound to in V1 tenants:
```python
def _existing_scopes_for(role: Role):
    in_default = (
        RoleBinding.objects
        .filter(resource_type="workspace")
        .filter(tenant__tenant_mapping_v2_write_activated_at=None)
        .filter(role__v1_source=role)
        .filter(Exists(
            Workspace.objects
            .filter(type=Workspace.Types.DEFAULT)
            .filter(id=Cast(OuterRef("resource_id"), UUIDField()))
        ))
    )

    # similarly for in_root (probably just make it a function)
    in_root = ...

    in_tenant = (
        RoleBinding.objects
        .filter(resource_type="tenant")
        .filter(tenant__tenant_mapping_v2_write_activated_at=None)
        .filter(role__v1_source=role)
    )
    
    scopes = set()
    
    if in_default:
        scopes.add(Scope.DEFAULT)
        
    if in_root:
        scopes.add(Scope.ROOT)
        
    if in_tenant:
        scopes.add(Scope.TENANT)
        
    return scopes
```

We then need to migrate if `not _existing_scopes_for(migrated_role).issubset([expected_scope])`, that is, if the existing scope set is neither empty nor what we expect it to be.

rbac/management/role/definer.py
```diff
@@ -48,6 +52,230 @@
...
+    )
+
+    replicator = OutboxReplicator()
+    migrated_count = 0
+
+    for group in groups:
+        try:
+            # Use the existing migration function to migrate bindings for this group
+            # This will update all system role bindings for the group to the correct scope
+            result = migrate_system_role_bindings_for_group(group, replicator)
```
randomnetcat, line 262: Theoretically we should also handle CARs here.

tests/management/role/test_definer.py
```diff
@@ -1228,6 +1228,55 @@ def test_non_default_role_has_no_parent(self):
...
+            non_default_role,
+            "Expected at least one seeded non-default system role with permissions in test data",
+        )
+
+        # Mock the migration function
+        with patch("management.role.definer._migrate_bindings_for_scope_change") as mock_migrate:
+            # Change the scope for this role's permissions
+            role_permissions = list(non_default_role.access.values_list("permission__permission", flat=True))
+            self.assertGreater(len(role_permissions), 0, f"Role {non_default_role.name} should have permissions")
+            if role_permissions:
```
randomnetcat: Unnecessary branch, given the previous assert.

tests/management/role/test_definer.py
```diff
@@ -1228,6 +1228,55 @@ def test_non_default_role_has_no_parent(self):
...
+                permission_pattern = f"{role_permissions[0].split(':')[0]}:*:*"
```
randomnetcat: Could just use `role_permissions[0].permission` directly if all we care about is changing the scope.

tests/management/role/test_definer.py
```diff
@@ -1228,6 +1228,55 @@ def test_non_default_role_has_no_parent(self):
...
+                # Verify that migration was NOT called for the non-default role
+                # (though it may be called for other default roles with same permission pattern)
+                for call in mock_migrate.call_args_list:
+                    role_arg = call[0][0]
+                    self.assertNotEqual(
+                        role_arg,
+                        non_default_role,
+                        f"Migration should not be called for non-default role {non_default_role.name} "
+                        "because it has no automatic bindings to migrate",
+                    )
```
randomnetcat: We do still need to migrate roles that aren't `platform_default` or `admin_default`; what we care about is updating the scope of groups that have been manually assigned the role. (Default access groups get it for free by updating the role's V2 parents.)

tests/management/role/test_seeding_binding_scope_update.py
```diff
@@ -0,0 +1,333 @@
...
+        v2_role = SeededRoleV2.objects.get(uuid=test_role.uuid)
+
+        # Find bindings for this workspace at the default workspace
+        initial_bindings = self.tuples.find_tuples(
+            all_of(
+                resource("rbac", "workspace", str(self.default_workspace.id)),
+                relation("binding"),
+            )
+        )
+        self.assertGreater(len(initial_bindings), 0, "Should have binding at default workspace initially")
```
randomnetcat: We need something more clever than this, since this will also find default access bindings. Inheriting from `DualWriteTest` might be useful here.

tests/management/role/test_seeding_binding_scope_update.py
```diff
@@ -0,0 +1,333 @@
...
+            seed_roles(force_update_relationships=True)
+
+        # For TENANT scope, bindings should be at the tenant level
+        tenant_resource_id = Tenant.org_id_to_tenant_resource_id(self.tenant.org_id)
+        tenant_bindings = self.tuples.find_tuples(
+            all_of(
+                resource("rbac", "tenant", tenant_resource_id),
+                relation("binding"),
+            )
+        )
```
randomnetcat: As above, we need to be more clever than checking for the existence of any role bindings here.

tests/management/role/test_seeding_binding_scope_update.py
```diff
@@ -0,0 +1,333 @@
...
+        self.assertGreater(len(tenant_bindings), 0, "Should have binding at tenant after migration")
```
randomnetcat: We should also check that the default workspace binding was removed.

rbac/management/role/definer.py
Ellen-Yi-Dong: Oh! I was talking JayZ earlier today, and I had mentioned adding this field (though it requires a migration) to add a scope field here: https://github.com/RedHatInsights/insights-rbac/pull/2785

What do you think about this..?

rbac/management/role/definer.py
```diff
@@ -48,6 +51,257 @@
...
+        A set of Scopes where the role currently has bindings in V1 tenants.
+        Returns empty set if role is None, if no bindings exist, or if an error occurs.
+    """
+    if v1_role is None:
+        return set()
+
+    try:
+        from django.db.models import Exists, OuterRef
+        from django.db.models.functions import Cast
+        from django.db.models.fields import UUIDField
```
astrozzc: Can we put these to top of the file? NVM if there is circular dependency issue

rbac/management/role/definer.py
```diff
@@ -48,6 +51,257 @@
...
+
+        # Check for bindings at DEFAULT workspace (workspace type = DEFAULT)
+        in_default = has_bindings(
+            RoleBinding.objects.filter(resource_type="workspace")
+            .filter(tenant__tenant_mapping__v2_write_activated_at=None)
+            .filter(role__v1_source=v1_role)
+            .filter(
+                Exists(
+                    Workspace.objects.filter(type=Workspace.Types.DEFAULT).filter(
+                        id=Cast(OuterRef("resource_id"), UUIDField())
```
astrozzc, line 89: This can be into the has_bindings function, the input is v1_role and scope, and then if scope is in root or default, type = ROOT or DEFAULT append the filter
.filter(
                Exists(
                    Workspace.objects.filter(type=type).filter(
                        id=Cast(OuterRef("resource_id"), UUIDField())
                    )
                )

rbac/management/role/definer.py
```diff
@@ -48,6 +51,257 @@
...
+                "Failed to migrate bindings for group %s with role %s",
+                group.uuid,
+                v1_role.name,
+                exc_info=True,
+            )
+
+    # Migrate CAR bindings
+    for car in cars:
+        try:
+            result = _migrate_car_bindings(car, replicator)
```
randomnetcat: If you're using this function here, we should just make it public and remove the underscore.

tests/management/role/test_seeding_binding_scope_update.py
```diff
@@ -0,0 +1,372 @@
...
+        from management.role_binding.model import RoleBinding
+        from management.role.v2_model import RoleV2
+
+        v2_role = RoleV2.objects.get(v1_source=test_role)
+        RoleBinding.objects.get_or_create(
+            tenant=self.tenant,
+            role=v2_role,
+            resource_type="workspace",
+            resource_id=str(self.default_workspace.id),
+        )
```
randomnetcat: `add_roles` should do this; probably should just assert that it exists?

rbac/management/role/definer.py
```diff
@@ -48,6 +51,257 @@
...
+        new_scope: The new scope
+    """
+    from management.group.model import Group
+    from migration_tool.migrate_binding_scope import migrate_system_role_bindings_for_group
+    from management.relation_replicator.outbox_replicator import OutboxReplicator
+    from api.cross_access.model import CrossAccountRequest
+    from migration_tool.migrate_binding_scope import _migrate_car_bindings
+
+    # Find all groups (non-public tenant) that have this system role assigned
+    groups_with_role = Group.objects.filter(policies__roles=v1_role).exclude(tenant__tenant_name="public").distinct()
```
randomnetcat, line 232: Ugh, it's actually quite tricky to ensure we get all groups here. A concurrent V1 role assignment could determine the scope using the old permissions, then assign it after the list of groups to update is determined. The two obvious solutions would be (a) using a `SERIALIZABLE` transaction for V1 role assignment and (b) `SELECT FOR UPDATE` the role during seeding, then `SELECT FOR SHARE` during V1 role assignment. I'd personally lean towards (a).

@astrozzc, thoughts?

rbac/management/role/definer.py
```diff
@@ -25,29 +25,269 @@
...
+    Args:
+        v1_role: The V1 system role whose scope has changed
+        old_scopes: Set of scopes where the role currently has bindings
+        new_scope: The new scope
+    """
+    from management.group.model import Group
+    from migration_tool.migrate_binding_scope import migrate_system_role_bindings_for_group
+    from management.relation_replicator.outbox_replicator import OutboxReplicator
+    from api.cross_access.model import CrossAccountRequest
+    from migration_tool.migrate_binding_scope import _migrate_car_bindings
```
randomnetcat: Should put these at the top of the file if no circular dependency issue.

tests/management/role/test_seeding_binding_scope_update.py
```diff
@@ -0,0 +1,372 @@
...
+
+        # Bootstrap tenant with workspaces but keep it as V1 (no V2 activation)
+        # The scope change migration only applies to V1 tenants
+        bootstrap_result = bootstrap_tenant_for_v2_test(self.tenant, tuples=self.tuples)
+        self.root_workspace = bootstrap_result.root_workspace
+        self.default_workspace = bootstrap_result.default_workspace
+
+        # Deactivate V2 writes to make this a V1 tenant
+        # This ensures _determine_old_scopes can find the bindings
+        from management.tenant_mapping.model import TenantMapping
```
randomnetcat: Imports should just be at the top. (Below, as well.)

tests/management/role/test_seeding_binding_scope_update.py
```diff
@@ -0,0 +1,372 @@
...
+        )
+
+        # Change scope to TENANT
+        with self.settings(
+            ROOT_SCOPE_PERMISSIONS="",
+            TENANT_SCOPE_PERMISSIONS="inventory:*:*",
+        ):
+            seed_roles(force_update_relationships=True)
+
+        # For TENANT scope, bindings should be at the tenant level
```
randomnetcat: We should also assert that the old binding was removed.

tests/management/role/test_seeding_binding_scope_update.py
```diff
@@ -0,0 +1,372 @@
...
+
+            # All detected scopes should be valid
+            for scope in detected_scopes:
+                self.assertIn(
+                    scope,
+                    [Scope.DEFAULT, Scope.TENANT, Scope.ROOT],
+                    f"All scopes should be valid, got {scope}",
+                )
+        else:
+            self.skipTest("No platform_default seeded roles found to test with")
```
randomnetcat: We should assert instead; we control the test data, so we know what to expect.

tests/management/role/test_seeding_binding_scope_update.py
```diff
@@ -0,0 +1,372 @@
...
+
+            # All detected scopes should be valid
+            for scope in detected_scopes:
+                self.assertIn(
+                    scope,
+                    [Scope.DEFAULT, Scope.TENANT, Scope.ROOT],
+                    f"All scopes should be valid, got {scope}",
+                )
+        else:
+            self.skipTest("No admin_default seeded roles found to test with")
```
randomnetcat: Same here.

rbac/management/role/definer.py
astrozzc: Good point, (a) would require we update the other code path to use SERIALIZABLE? sounds very complicated. 
Not sure if worth the complication, I wonder if we just re-run in that case. Seems much easier

randomnetcat: Theoretically, re-running it doesn't actually guarantee we catch anything new. There's the pathological case where an arbitrary number of re-runs are scheduled before the concurrent transaction completes. So, I think it would only be fully correct to do one of the two things I mentioned above.

It's possible this could be good enough, though? We may at least want to create a ticket to track this.

rbac/management/role/definer.py
```diff
@@ -25,29 +25,329 @@
...
+                    "Failed to migrate bindings for group %s with role %s",
+                    group.uuid,
+                    v1_role.name,
+                    exc_info=True,
+                )
+
+        # Migrate CAR bindings
+        for car in cars:
+            try:
+                result = _migrate_car_bindings(car, replicator)
```
randomnetcat: Should make this public.

randomnetcat (approved):
LGTM pending the below and [this comment](https://github.com/RedHatInsights/insights-rbac/pull/2698#discussion_r3327893242).

rbac/management/role/definer.py
Ellen-Yi-Dong: Updated! Both this comment and the comment linked: based changes on the first option provided

rbac/management/group/definer.py
```diff
@@ -50,44 +50,61 @@
...
-            logger.info("Finished seeding default org admin group %s.", name)
+        # If we're already in a transaction (e.g., from a test), don't create a new atomic block
+        # to avoid "SET TRANSACTION ISOLATION LEVEL must be called before any query" errors
+        if connection.in_atomic_block:
+            # Already in a transaction, proceed without atomic_block
+            return _seed_group_impl(public_tenant)
 
-        return group, admin_group
+        with atomic_block():
+            return _seed_group_impl(public_tenant)
```
randomnetcat: I'm not sure what you're going for here? If `SERIALIZABLE` is necessary, then it's necessary and we have to actually ensure we do it. If it isn't, then `with transaction.atomic()` is fine (regardless of whether we are already in a transaction).

rbac/management/group/definer.py
```diff
@@ -157,7 +174,12 @@ def clone_default_group_in_public_schema(group, tenant) -> Optional[Group]:
 
 
 def add_roles(group, roles_or_role_ids, tenant, user=None):
-    """Process list of roles and add them to the group."""
+    """Process list of roles and add them to the group.
+
+    Note: This function should be called within a SERIALIZABLE transaction when adding
+    system roles to prevent race conditions during scope migrations. The caller is
+    responsible for ensuring the appropriate transaction isolation level.
+    """
```
randomnetcat: If we have to document this, we should probably just make it `@atomic` instead (to ensure it's actually enforced)?

rbac/management/atomic_transactions.py
```diff
@@ -63,8 +63,19 @@ def wrapper(*args, **kwargs):
...
+    """
+    from django.db import connection
+
     if is_atomic_disabled():
         return transaction.atomic()
 
+    # If we're already in a transaction, don't try to set isolation level
+    # as it will fail with "SET TRANSACTION ISOLATION LEVEL must be called before any query"
+    if connection.in_atomic_block:
+        return transaction.atomic()
```
randomnetcat: We can't do this. We use this decorator specifically where we need `SERIALIZABLE` isolation; if we just needed `READ COMMITTED`, we would use a plain `transaction.atomic()`.

tests/api/test_utils.py
```diff
@@ -10,7 +10,7 @@
 from management.tenant_service.v2 import V2TenantBootstrapService
 
 
-class TestAPIUtils(TestCase):
+class TestAPIUtils(TransactionTestCase):
```
randomnetcat: Why the change?

tests/internal/test_migrate_binding_scope_api.py
```diff
@@ -874,12 +875,34 @@ def test_migration_creates_bindings_for_roles_with_no_bindings(self):
...
     def test_v2_tenant_not_migrated(self):
-        fixture = RbacFixture()
+        # Create system role manually instead of using RbacFixture to avoid transaction issues
+        system_role = Role.objects.create(
+            name="system role",
+            system=True,
+            platform_default=False,
+            admin_default=False,
+            tenant=self.public_tenant,
+        )
```
randomnetcat: Setting `ATOMIC_RETRY_DISABLED=True` should be sufficient to avoid having to do this? (And likewise below.)

tests/management/role/test_seeding_binding_scope_update.py
```diff
@@ -0,0 +1,326 @@
...
+    """Test that system role bindings are updated when role scope changes during seeding."""
+
+    def setUp(self):
+        """Set up test data."""
+        super().setUp()
+        self.public_tenant = Tenant.objects.get(tenant_name="public")
+
+        # Deactivate V2 writes to make this a V1 tenant
+        # This ensures _determine_old_scopes can find the bindings
+        TenantMapping.objects.filter(tenant=self.tenant).update(v2_write_activated_at=None)
```
randomnetcat: I wouldn't do this, but if we need to, you can use `assert_v1_write_allowed` instead.

rbac/management/role/definer.py
```diff
@@ -25,29 +25,284 @@
...
+        old_scopes: Set of scopes where the role currently has bindings
+        new_scope: The new scope
+    """
+    replicator = OutboxReplicator()
+
+    # Find all groups (non-public tenant) that have this system role assigned
+    groups_with_role = Group.objects.filter(policies__roles=v1_role).exclude(tenant__tenant_name="public").distinct()
+
+    # Find all approved CARs that have this system role
+    cars_with_role = CrossAccountRequest.objects.filter(roles=v1_role, status="approved")
```
randomnetcat, line 235: Okay, this still isn't guaranteed to catch everything, but at this point it's probably Good Enough. We can always just run the scope migration manually to fix any stragglers. I'll just create a ticket to track this.

randomnetcat (approved):

rbac/management/role/definer.py
randomnetcat: [RHCLOUD-48478](https://redhat.atlassian.net/browse/RHCLOUD-48478)

rbac/management/role/definer.py
randomnetcat: [RHCLOUD-48478](https://redhat.atlassian.net/browse/RHCLOUD-48478)

PR #2722, opened 2026-04-08
[RHCLOUD-46189] Add database indexes for case-insensitive name filtering

dsova06:
## Big Picture

Recent changes (RHCLOUD-46140, RHCLOUD-46177) switched V2 name filtering to case-insensitive operations (iexact, iregex, icontains). Without proper indexes, PostgreSQL falls back to sequential scans for these queries, which gets progressively slower as tables grow. This PR adds GIN trigram indexes -- a single index type that efficiently accelerates all three case-insensitive query patterns -- keeping name-based lookups fast regardless of table size.

## Summary

RHCLOUD-46140 and RHCLOUD-46177 changed V2 name filtering to case-insensitive (`iexact`, `iregex`, `icontains`). PostgreSQL's default B-tree indexes cannot accelerate these lookups, causing sequential scans that degrade with table size.

This PR adds GIN trigram indexes (`pg_trgm`) on the `name` column of both `RoleV2` and `Workspace` models -- a single index type that efficiently supports all three case-insensitive query patterns.

## Changes

1. **Migration 0084** -- enables `pg_trgm` extension (idempotent) and creates `rolev2_name_trgm_idx` and `workspace_name_trgm_idx`
2. **Model Meta** -- declares both indexes in `RoleV2.Meta.indexes` and `Workspace.Meta.indexes` so Django stays in sync
3. **Tests** -- 8 new tests verifying: extension enabled, indexes exist with correct type, model Meta definitions, and case-insensitive queries return correct results

## Test Plan

- `pipenv run tox -e py312-fast -- tests.management.test_name_indexes` -- all 8 new tests pass
- `pipenv run tox -e py312-fast` -- full test suite passes (no regressions)
- Migration applies cleanly on a fresh database

## Verification

```sql
-- Check indexes exist after migration
SELECT indexname, indexdef FROM pg_indexes
WHERE tablename IN ('management_rolev2', 'management_workspace')
AND indexname LIKE '%trgm%';

-- Verify index usage (look for Bitmap Index Scan instead of Seq Scan)
EXPLAIN ANALYZE SELECT * FROM management_workspace
WHERE UPPER("name"::text) LIKE UPPER('%search term%');
```

Note: PostgreSQL may still choose sequential scan on very small tables where it is cheaper.

## JIRA

https://redhat.atlassian.net/browse/RHCLOUD-46189


EvanCasey13 (approved):
LGTM!

PR #2767, opened 2026-04-20
[RHCLOUD-45423] Add api to run migration to replicate role to tenant relationship

astrozzc:
## Link(s) to Jira
- https://redhat.atlassian.net/browse/RHCLOUD-45423

For existing roles, replicate their relationships to tenants

## Local Testing
Unit test

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

## Summary by Sourcery

Add an internal utility endpoint and migration routine to backfill owner relationships between custom V2 roles and their tenants using relation replication.

New Features:
- Introduce a POST-only internal API endpoint to trigger migration of custom V2 role owner relationships to tenants.

Enhancements:
- Add a migration helper that scans custom V2 roles and emits owner relationship replication events in paginated chunks, with support for real or no-op replicators.

Tests:
- Add unit tests covering successful replication of owner tuples, error handling when tenants lack resource IDs, and processing of multiple tenants.

### DIFF
--- a/rbac/internal/custom_v2_role_tenant_migration.py
+++ b/rbac/internal/custom_v2_role_tenant_migration.py
@@ -1,0 +1,133 @@
+#
+# Copyright 2026 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+"""Backfill rbac/role#owner@rbac/tenant for custom V2 roles from each role's tenant row.
+
+Scans the ``RoleV2`` table only (custom roles are on the order of thousands), not the tenant table
+(millions of rows). Uses keyset pagination on ``rolev2.pk``.
+"""
+
+from __future__ import annotations
+
+import logging
+from typing import Any
+
+from django.conf import settings
+from management.atomic_transactions import atomic_block
+from management.relation_replicator.noop_replicator import NoopReplicator
+from management.relation_replicator.outbox_replicator import OutboxReplicator
+from management.relation_replicator.relation_replicator import (
+    PartitionKey,
+    RelationReplicator,
+    ReplicationEvent,
+    ReplicationEventType,
+)
+from management.role.relations import role_owner_relationship
+from management.role.v2_model import RoleV2
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_CHUNK_SIZE = 500
+
+
+class TenantResourceIdMissingError(ValueError):
+    """Raised when a custom role's tenant has no ``tenant_resource_id()`` (no ``org_id``)."""
+
+
+def _require_tenant_resource_id(role: RoleV2, resource_id: str | None) -> str:
+    if resource_id:
+        return resource_id
+    tenant = role.tenant
+    raise TenantResourceIdMissingError(
+        f"Tenant for custom role has no tenant_resource_id (org_id missing): "
+        f"role_uuid={role.uuid} role_name={role.name!r} tenant_id={tenant.pk}"
+    )
+
+
+def _base_queryset():
+    """Return all custom V2 roles (custom roles are only created for org tenants in practice)."""
+    return RoleV2.objects.filter(type=RoleV2.Types.CUSTOM).select_related("tenant").order_by("pk")
+
+
+def _process_chunk(
+    *,
+    chunk: list[RoleV2],
+    replicator: RelationReplicator,
+) -> None:
+    for role in chunk:
+        tenant = role.tenant
+        resource_id = _require_tenant_resource_id(role, tenant.tenant_resource_id())
+
+        replicator.replicate(
+            ReplicationEvent(
+                event_type=ReplicationEventType.UPDATE_CUSTOM_ROLE,
+                info={
+                    "role_uuid": str(role.uuid),
+                    "org_id": str(tenant.org_id) if tenant.org_id else "",
+                },
+                partition_key=PartitionKey.byEnvironment(),
+                add=[role_owner_relationship(role.uuid, resource_id)],
+                remove=[],
+            )
+        )
+        logger.info(
+            "Replicated owner tuple for custom V2 role uuid=%s name=%r tenant org_id=%s",
+            role.uuid,
+            role.name,
+            tenant.org_id,
+        )
+
+
+def replicate_custom_v2_role_owner_relationships(
+    *,
+    replicator: RelationReplicator | None = None,
+) -> dict[str, Any]:
+    """Emit owner tuples for every custom V2 role from ``role.tenant.tenant_resource_id()``.
+
+    Iterates **custom roles only** (typically thousands of rows), not tenants. Does not change PostgreSQL
+    role rows; only replicates ``rbac/role:<uuid>#owner@rbac/tenant:<id>``.
+
+    Raises ``TenantResourceIdMissingError`` if any custom role's tenant has no ``tenant_resource_id()``.
+
+    Fetches and processes in fixed-size chunks (``DEFAULT_CHUNK_SIZE``, keyset pagination on ``rolev2.pk``).
+    Each chunk is read and processed inside a separate SERIALIZABLE transaction (see ``atomic_block``).
+    """
+    if replicator is None:
+        if settings.REPLICATION_TO_RELATION_ENABLED:
+            replicator = OutboxReplicator()
+        else:
+            replicator = NoopReplicator()
+
+    base = _base_queryset()
+    replicated_count = 0
+
+    last_pk = 0
+    while True:
+        with atomic_block():
+            chunk = list(base.filter(pk__gt=last_pk)[:DEFAULT_CHUNK_SIZE])
+            if not chunk:
+                break
+            last_pk = chunk[-1].pk
+            _process_chunk(
+                chunk=chunk,
+                replicator=replicator,
+            )
+            replicated_count += len(chunk)
+
+    return {
+        "replicated_count": replicated_count,
+    }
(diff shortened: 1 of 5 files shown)


rbac/internal/specs/openapi.json
```diff
@@ -3794,21 +3794,44 @@
           "Relations"
         ],
         "summary": "Read tuples from the relations api.",
-        "description": "Reads relation tuples from the relations api based on filter criteria.",
+        "description": "Reads relation tuples from the relations api based on filter criteria. Custom roles include an `owner` edge from `rbac/role` to `rbac/tenant` (tenant resource id is `{PRINCIPAL_USER_DOMAIN}/{org_id}`); RBAC replicates this so `role#read` / `role#write` can derive from the tenant per the rbac schema.",
```
randomnetcat: This doesn't seem like the best place to document this? We don't mention any other specific relations here.

rbac/internal/specs/openapi.json
```diff
@@ -3794,21 +3794,44 @@
...
+                      "resource_type": "role",
+                      "resource_id": "019d8dab-4282-7141-b898-94f02d1986a4",
+                      "relation": "owner",
+                      "subject_filter": {
+                        "subject_namespace": "rbac",
+                        "subject_type": "tenant",
+                        "subject_id": "localhost/1234567",
+                        "relation": null
+                      }
+                    }
```
randomnetcat: Similarly, it doesn't seem necessary to put this here?

rbac/internal/custom_v2_role_tenant_migration.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def _require_tenant_resource_id(role: RoleV2, resource_id: str | None) -> str:
+    if resource_id:
+        return resource_id
+    tenant = role.tenant
+    raise TenantResourceIdMissingError(
+        f"Tenant for custom role has no tenant_resource_id (org_id missing): "
+        f"role_uuid={role.uuid} role_name={role.name!r} tenant_id={tenant.pk}"
+    )
```
randomnetcat: Could just take the `role` here and both get and check the ID in one operation?

rbac/internal/custom_v2_role_tenant_migration.py
```diff
@@ -0,0 +1,133 @@
...
+    base = _base_queryset()
+    replicated_count = 0
+
+    last_pk = 0
+    while True:
+        with atomic_block():
+            chunk = list(base.filter(pk__gt=last_pk)[:DEFAULT_CHUNK_SIZE])
+            if not chunk:
+                break
+            last_pk = chunk[-1].pk
```
randomnetcat: Any reason not to let Django handle this itself by just using a `QuerySet` `iterator()`?

rbac/internal/custom_v2_role_tenant_migration.py
```diff
@@ -0,0 +1,133 @@
...
+    """Return all custom V2 roles (custom roles are only created for org tenants in practice)."""
+    return RoleV2.objects.filter(type=RoleV2.Types.CUSTOM).select_related("tenant").order_by("pk")
+
+
+def _process_chunk(
+    *,
+    chunk: list[RoleV2],
+    replicator: RelationReplicator,
+) -> None:
+    for role in chunk:
```
randomnetcat: We do still have to lock all the roles here. V1 role operation can update V2 roles concurrently while not in a `SERIALIZABLE` transaction. This should probably do something similar to what `migrate_custom_role_bindings` does.

rbac/internal/custom_v2_role_tenant_migration.py
```diff
@@ -0,0 +1,133 @@
...
+    Raises ``TenantResourceIdMissingError`` if any custom role's tenant has no ``tenant_resource_id()``.
+
+    Fetches and processes in fixed-size chunks (``DEFAULT_CHUNK_SIZE``, keyset pagination on ``rolev2.pk``).
+    Each chunk is read and processed inside a separate SERIALIZABLE transaction (see ``atomic_block``).
+    """
+    if replicator is None:
+        if settings.REPLICATION_TO_RELATION_ENABLED:
+            replicator = OutboxReplicator()
+        else:
+            replicator = NoopReplicator()
```
randomnetcat: If the replicator is a `NoopReplicator`, this doesn't actually *do* anything. We should probably error out instead?

tests/internal/test_custom_v2_role_tenant_migration.py
```diff
@@ -0,0 +1,119 @@
...
+            verb="read",
+            permission="inventory:hosts:read",
+        )
+
+    def test_replicates_owner_tuple_using_tenant_resource_id(self):
+        """Each custom role with an org tenant gets one owner tuple replicated."""
+        role = CustomRoleV2.objects.create(name="custom_a", tenant=self.org_tenant)
+        role.permissions.add(self.perm)
+
+        replicator = MagicMock()
```
randomnetcat: We should probably just test this with an `InMemoryRelationReplicator` and actually check what tuples are being replicated?

tests/internal/test_custom_v2_role_tenant_migration.py
```diff
@@ -0,0 +1,119 @@
...
+            account_id="acc-ot2",
+            org_id="org-owner-002",
+            ready=True,
+        )
+        bootstrap_tenant_for_v2_test(other)
+
+        CustomRoleV2.objects.create(name="r1", tenant=self.org_tenant).permissions.add(self.perm)
+        CustomRoleV2.objects.create(name="r2", tenant=other).permissions.add(self.perm)
+
+        replicator = MagicMock()
```
randomnetcat: Similarly to above, we should probably do more than this. What's interesting isn't whether the replicator is being called; it's what tuples are being replicated.

rbac/internal/custom_v2_role_tenant_migration.py
astrozzc: Good idea

rbac/internal/custom_v2_role_tenant_migration.py
astrozzc: Updated

rbac/internal/custom_v2_role_tenant_migration.py
astrozzc: The migration already runs each chunk inside atomic_block() which uses SERIALIZABLE isolation. If a concurrent V1 operation touches a role we've already read, PostgreSQL will abort one of the transactions. Since this is a one-shot backfill that only reads roles (no Postgres writes to role rows), the risk is minimal and a retry of the migration is acceptable.

rbac/internal/custom_v2_role_tenant_migration.py
astrozzc: Agree, I would blame AI for this 😛

tests/internal/test_custom_v2_role_tenant_migration.py
astrozzc: Updated

rbac/internal/custom_v2_role_tenant_migration.py
randomnetcat: That's not true; [the Postgres docs](https://www.postgresql.org/docs/current/transaction-iso.html#XACT-SERIALIZABLE) don't say it super explicitly, but they do make clear that `SERIALIZABLE` isolation is only effective *between `SERIALIZABLE` transactions*. The V1 code doesn't use `SERIALIZABLE`, so there won't be any protection there.

rbac/internal/specs/openapi.json
```diff
@@ -3799,16 +3799,18 @@
...
+                    "resource_namespace": "rbac",
+                    "resource_type": "group",
+                    "resource_id": "bob_club",
+                    "relation": "member",
+                    "subject_filter": {
+                      "subject_namespace": "rbac",
+                      "subject_type": "principal",
+                      "subject_id": "bob",
+                      "relation": null
+                    }
```
randomnetcat: No need for the change here at all?

rbac/internal/custom_v2_role_tenant_migration.py
```diff
@@ -0,0 +1,122 @@
...
+    Raises ``TenantResourceIdMissingError`` if any custom role's tenant has no ``tenant_resource_id()``.
+
+    Streams rows with ``iterator()`` (see Django docs for fetch behavior with ``select_related``).
+    Each batch of up to ``DEFAULT_CHUNK_SIZE`` roles is processed inside a separate SERIALIZABLE
+    transaction (see ``atomic_block``).
+    """
+    if replicator is None:
+        if settings.REPLICATION_TO_RELATION_ENABLED:
+            replicator = OutboxReplicator()
+        raise ValueError("Replication to relations is disabled")
```
randomnetcat:
```suggestion
        if not settings.REPLICATION_TO_RELATION_ENABLED:
            raise ValueError("Replication to relations is disabled")
        replicator = OutboxReplicator()
```

rbac/internal/specs/openapi.json
astrozzc: Oops, some legacy missed

rbac/internal/custom_v2_role_tenant_migration.py
astrozzc: You are right! I did some research and indeed they have to be [same SERIALIZABLE transactions](https://github.com/postgres/postgres/blob/master/src/backend/storage/lmgr/predicate.c#L58). Good to know!
Updated

rbac/internal/custom_v2_role_tenant_migration.py
astrozzc: Now we have to put select_for_update into transaction, can't use the iterator now

rbac/internal/custom_v2_role_tenant_migration.py
```diff
@@ -0,0 +1,127 @@
...
+
+    while True:
+        with atomic_block():
+            chunk = list(_base_queryset().filter(pk__gt=last_pk).select_for_update()[:DEFAULT_CHUNK_SIZE])
+            if not chunk:
+                break
+            for role in chunk:
+                _replicate_role(role, replicator)
+                replicated_count += 1
+            last_pk = chunk[-1].pk
```
randomnetcat, line 123: Ah, we're back to doing this manually, I see. I think I'd still prefer to not do it this way, but it does look like this will work.

randomnetcat (approved):

rbac/internal/custom_v2_role_tenant_migration.py
astrozzc: Right, cause for the Django handled one, it would select_for_update to lock all roles at once

PR #2794, opened 2026-04-24
RHCLOUD-47270 - MCP - add authorization context to list_audit_logs mcp endpoint

EvanCasey13:
## Link(s) to Jira
https://redhat.atlassian.net/browse/RHCLOUD-47270

Enhance the list_audit_logs MCP tool with filtering capabilities and an optional include_authorization flag that enriches audit entries with the role and permission that authorized each action.

 1. New Filtering Parameters (mcp_views.py:437-446)

  The tool now accepts:
  - principal_username — filter by who made the change (partial match via icontains)
  - resource_type — filter by resource (group/role/workspace/etc.)
  - action — filter by action (add/edit/delete/create/remove)

  2. Authorization Enrichment (mcp_views.py:468-515)

  When include_authorization=True:
  - For each audit log entry, the tool determines which role and permission authorized the action
  - Returns an authorized_by object containing:
    - role — the role display name
    - via_group — the group that granted the role
    - permission — the specific permission string (e.g., rbac:group:write)
  - If the user no longer exists, a note field explains this

## Local Testing

Run all MCP tests

```
```

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

## Summary by Sourcery

Add filtering and optional authorization context enrichment to the MCP audit log listing endpoint.

New Features:
- Allow filtering audit logs by principal username, resource type, and action in the MCP list_audit_logs endpoint.
- Support an include_authorization flag to return the current role and permission that authorized each audited action, including org admin handling.

Enhancements:
- Inline the MCP audit log listing logic instead of delegating to the AuditLogViewSet, adding tenant scoping, pagination, and structured JSON responses.
- Provide descriptive API documentation for new filters, sorting behavior, and the semantics of the authorization context in audit log entries.

### DIFF
--- a/rbac/management/mcp_views.py
+++ b/rbac/management/mcp_views.py
@@ -490,7 +490,11 @@ def list_audit_logs(
     # Enrich with authorization context
     auth_cache: dict[str, dict[str, dict[str, Any] | None]] = {}
     org_admin_cache: dict[str, bool] = {}
-    org_admin_auth = {"role": "Org Admin", "via_group": None, "permission": "(bypasses all RBAC checks)"}
+    org_admin_auth: dict[str, Any] = {
+        "role": "Org Admin",
+        "via_group": None,
+        "permission": "(bypasses all RBAC checks)",
+    }
     results = []
 
     for entry in entries:
@@ -506,6 +510,7 @@ def list_audit_logs(
         if actor not in org_admin_cache:
             org_admin_cache[actor] = _is_org_admin(actor, tenant.org_id)
 
+        auth_info: dict[str, Any] | None
         if org_admin_cache[actor]:
             auth_info = org_admin_auth
         elif cache_key not in auth_cache[actor]:
--- a/rbac/management/mcp_views.py
+++ b/rbac/management/mcp_views.py
@@ -34,10 +34,13 @@ from django.urls import reverse
 from django.utils.decorators import method_decorator
 from django.views import View
 from django.views.decorators.csrf import csrf_exempt
+from management.access.model import Access
 from management.access.view import AccessView
 from management.audit_log.view import AuditLogViewSet
 from management.group.view import GroupViewSet
+from management.models import AuditLog
 from management.permission.view import PermissionViewSet
+from management.principal.model import Principal
 from management.principal.view import PrincipalView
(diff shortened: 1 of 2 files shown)


dsova06 (commented):
Nice feature -- being able to trace *why* someone was authorized for an action is genuinely useful for the MCP agent. I went through the changes and have some thoughts:

1. **Org admin handling got removed.** The latest commit strips the org admin check from `_find_authorizing_role`, so org admins now show `authorized_by: null` with "User not found or no longer has permissions." Since org admins bypass RBAC entirely, this would be pretty confusing for anyone reading the output. The earlier version that returned "Org Admin (bypasses all RBAC checks)" felt like the right approach -- would be great to bring that back.

2. **Empty page returns `count: 0` instead of the actual total.** When offset is past the end of results, the early return hardcodes `{"meta": {"count": 0}}` even though `total_count` might be non-zero. This would break pagination for consumers -- I think we'd want to always return the real `total_count` here.

3. **`_REQUIRED_PERMISSIONS` is missing `user` and `permission` resource types.** The AuditLog model defines both, but the mapping doesn't cover them. The fallback ends up producing things like `rbac:user:write` which isn't a real permission -- user operations typically need `rbac:principal:write`. Might be worth filling those in to avoid silent wrong results.

### Would be good to clean up

4. **`order_by` isn't validated.** It goes straight into `queryset.order_by()` -- a bad value would throw a `FieldError` and 500. The old ViewSet delegation handled this internally. An allowlist of valid sort fields would make this more robust.

5. **`_permission_satisfies` duplicates `_permission_matches`.** There's already an identical function at line ~994 in this file -- same logic, same wildcard support. We could just reuse `_permission_matches` and avoid the duplication.

### Something to think about

6. **`links` field dropped from response.** The old response format included `{meta, links, data}` -- the new one only returns `{meta, data}`. Probably fine for MCP consumers but wanted to flag it in case it was unintentional.

dsova06 (approved):
LGTM

PR #2800, opened 2026-04-27
 RHCLOUD-47270 - Add group_name filter to MCP audit tools

EvanCasey13:
## Link(s) to Jira
https://redhat.atlassian.net/browse/RHCLOUD-47270

  Improves MCP audit tool accuracy by adding targeted filters that help the LLM narrow results instead of scanning
  hundreds of entries:

  - Add group_name and role_name filters to list_audit_logs for precise lookups
  - Allow list_group_roles to accept group_name instead of requiring UUID lookup first
  - Update tool descriptions to guide the LLM toward filtered queries

  Test plan

  - Verify list_audit_logs(group_name='...') filters entries containing that group name
  - Verify list_audit_logs(role_name='...') filters entries containing that role name
  - Verify list_group_roles(group_name='...') resolves the group and returns its roles
  - Verify error returned when group_name not found

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

## Summary by Sourcery

Add support for group-based filtering in MCP audit and group role tools to make investigations and role lookups more precise.

New Features:
- Allow list_audit_logs to filter results by group_name via a case-insensitive search of the description field.
- Allow list_group_roles to accept either group_uuid or group_name (case-insensitive) for resolving the target group, including appropriate error responses when resolution fails.

Tests:
- Add tests covering audit log filtering by group_name and group role lookups by group_name, case-insensitivity, and error conditions when parameters are missing or groups are not found.

### DIFF
--- a/rbac/management/mcp_views.py
+++ b/rbac/management/mcp_views.py
@@ -730,10 +730,12 @@ def list_permission_options(
 
 @register_tool(
     description=(
-        "List roles assigned to a specific group. Provide either group_uuid OR group_name. "
+        "List roles assigned to a specific group. Provide either group_uuid OR group_name "
+        "(if both are provided, group_uuid takes precedence). "
         "Use group_name for convenience (case-insensitive lookup). "
         "Example: list_group_roles(group_name='<group-name>') to see roles assigned to that group. "
         "Filter results by role_name, role_description, role_display_name, or role_system. "
+        "Set exclude='true' to list roles NOT in the group. "
         "Order by: 'name', 'display_name', 'modified', 'policyCount' (prefix with '-' to reverse). "
         "Returns: {meta: {count}, links, data: [{uuid, name, description, system, ...}]}."
     ),
--- a/rbac/management/mcp_views.py
+++ b/rbac/management/mcp_views.py
@@ -421,17 +421,17 @@ def list_permissions(
 
 @register_tool(
     description=(
-        "List audit log entries showing who made changes to RBAC resources. "
-        "Use this to answer: 'Who added a role to group X?' or 'Who modified group Y?' "
-        "Filter by: principal_username (who made the change), resource_type "
+        "List audit log entries recording RBAC changes for the authenticated organization. "
+        "Each entry records who changed what (principal, resource_type, action). "
+        "Filter by principal_username (who made the change), resource_type "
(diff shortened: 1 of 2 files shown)


dsova06 (commented):
This PR adds `group_name` and `role_name` filters to `list_audit_logs` and allows `list_group_roles` to accept a `group_name` instead of requiring a UUID upfront. Both are practical improvements for LLM-driven audit workflows -- having to do a UUID lookup just to inspect a group's roles is exactly the kind of extra hop that wastes tokens and confuses the model. The implementation is clean and follows existing patterns well; tenant scoping is correct throughout.

1. **Test assertions for `list_group_roles` by-name tests don't verify the actual data.** `test_list_group_roles_by_name` and `test_list_group_roles_by_name_case_insensitive` check that the response has no error and contains a `data` key, but they don't verify the count or the role names. These tests would pass even if the resolution returned the wrong group's roles or an empty list. It'd be great to assert `tool_output["meta"]["count"] == 1` and that the returned role name matches what was set up (Sourcery flagged this too, and the suggestion looks right).

### Would be good to clean up

2. **The `exclude` parameter got dropped from the tool description.** The old description mentioned `Set exclude='true' to list roles NOT in the group`, which is useful for LLMs that want to find roles *not* assigned to a group. The parameter still works, but the LLM won't know to use it unless it inspects the schema. Would be worth keeping a brief mention.

### Something to think about

3. **Both `group_name` and `role_name` filter on `description__icontains`, so they produce an AND when used together.** This is actually useful for the main use case (finding entries where a specific role was added to a specific group), but it also means a group named "admin" would match any description containing "admin" -- including role-related descriptions. For MCP tool purposes this is probably fine since the LLM can further narrow results, but worth being aware of.

4. **When both `group_uuid` and `group_name` are passed to `list_group_roles`, `group_uuid` wins silently.** Sourcery flagged this. For an MCP tool this is unlikely to happen in practice, but a one-liner note in the description ("if both are provided, group_uuid takes precedence") would prevent any confusion.

dsova06 (approved):
LGTM

PR #2827, opened 2026-05-05
Handle empty type name in tuple lookup

randomnetcat:
## Link(s) to Jira
N/A

Don't pass `rbac` as a type namespace when looking up tuples with unconstrained type. This was breaking the orphan removal script, which, since #2748, looks up tuples with an unconstrained subject type.

## Summary by Sourcery

Adjust tuple reading logic to avoid assuming the 'rbac' namespace when type names are unconstrained and align tests with SpiceDB filtering semantics.

Bug Fixes:
- Prevent orphan removal and tuple lookup failures when subject or resource type names are empty by not defaulting to the 'rbac' namespace in those cases.

Enhancements:
- Update test tuple filtering utilities to construct filters only for provided fields and to manually enforce subject type constraints consistent with SpiceDB behavior.

### DIFF
--- a/rbac/management/relation_replicator/relations_api_replicator.py
+++ b/rbac/management/relation_replicator/relations_api_replicator.py
@@ -264,6 +264,9 @@ class RelationsApiReplicator(RelationReplicator):
         if (pagination_limit is None) and (continuation_token is not None):
             raise TypeError("A pagination limit must be provided if a continuation token is.")
 
+        if not resource_type and not subject_type:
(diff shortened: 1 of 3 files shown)


astrozzc:
should we restrict the resource_type and subject_type at least one is not empty? or it might accidentally read everything for removal?

randomnetcat:
Done.

astrozzc (approved):

PR #2869, opened 2026-05-12
[RHCLOUD-47086][RHCLOUD-47087] Add MCP write infrastructure and CREATE tools

dsova06:
## Link(s) to Jira
- [RHCLOUD-47086](https://issues.redhat.com/browse/RHCLOUD-47086)
- [RHCLOUD-47087](https://issues.redhat.com/browse/RHCLOUD-47087)

Adds write support to the RBAC MCP server, fulfilling two requirements:

1. **Read-only mode toggle (RHCLOUD-47086)**: Adds `MCP_WRITE_ENABLED` env var (default `False`) so the MCP server deploys in read-only mode by default, per RH MCP Server Guidelines. When disabled, write tools are hidden from `tools/list` and calls are rejected with a JSON-RPC error.

2. **8 CREATE write tools (RHCLOUD-47087)**: Implements write tools that delegate to existing Django views via internal POST requests. All write tools require authentication and respect existing view-level permission checks.

**Write infrastructure changes:**
- Extended `ToolConfig` dataclass with `write: bool` field
- Extended `register_tool` decorator to accept `write=True`
- Extended `_clone_request` for POST support with JSON body and CSRF bypass
- Added `_call_view_write` helper for POST delegation to Django views
- Extracted `_resolve_group_uuid` helper (deduplicated from 3 occurrences)
- Updated `_handle_tools_list` to filter write tools when disabled
- Updated `_handle_tools_call` to reject write calls when disabled

**8 CREATE tools implemented:**

| Tool | API Endpoint | Gating |
|------|-------------|--------|
| `create_group` | POST /api/v1/groups/ | both |
| `add_principals_to_group` | POST /api/v1/groups/{uuid}/principals/ | both |
| `add_roles_to_group` | POST /api/v1/groups/{uuid}/roles/ | V1 only |
| `create_role_v1` | POST /api/v1/roles/ | V1 only |
| `create_role` | POST /api/v2/roles/ | V2 only |
| `create_role_bindings` | POST /api/v2/role-bindings/:batchCreate | V2 only |
| `create_workspace` | POST /api/v2/workspaces/ | V2 only |
| `create_cross_account_request` | POST /api/v1/cross-account-requests/ | both |

**Three-level write gating:** `MCP_WRITE_ENABLED` flag -> `api_version` gating (V1/V2) -> Django view permission checks.

## Local Testing

```bash
# Run MCP-specific tests

# Verify write tools are hidden by default
http POST localhost:8000/_private/_a2s/mcp/ \
  jsonrpc="2.0" method="tools/list" id:=1 params:='{}' \
  x-rh-identity:<base64-identity>
# Should NOT include create_group, add_principals_to_group, etc.

# Enable write mode and verify tools appear
MCP_WRITE_ENABLED=True make serve
# Same tools/list call should now include write tools

# Test write tool execution
http POST localhost:8000/_private/_a2s/mcp/ \
  jsonrpc="2.0" method="tools/call" id:=2 \
  params:='{"name":"create_group","arguments":{"name":"Test Group"}}' \
  x-rh-identity:<base64-identity>
```

## Checklist
  - No action required. `MCP_WRITE_ENABLED` defaults to `False`, so write tools are disabled until explicitly enabled in deployment config.
  - 16 new tests across 4 test classes covering write gating, tool execution, auth, and V1/V2 behavior

- [x] Input Validation
  - All input validated by existing Django view serializers via delegation
- [x] Output Encoding
  - Responses produced by existing DRF serializers
  - Write tools require `requires_auth=True`; identity header forwarded to delegated views
- [x] Session Management
  - No sessions; stateless MCP over HTTP POST
- [x] Access Control
  - Three-level gating: MCP_WRITE_ENABLED flag, API version gating, Django view permission checks
  - Tenant isolation preserved through request cloning (tenant context forwarded)
- [x] Cryptographic Practices
  - No cryptographic operations added
  - Write rejections logged at WARNING level; tool calls logged at INFO level
- [x] Data Protection
  - No sensitive data exposed; all data access through existing views
- [x] Communication Security
  - Internal-only endpoint (`/_private/_a2s/mcp/`)
- [x] System Configuration
  - Write mode disabled by default (`MCP_WRITE_ENABLED=False`)
- [x] Database Security
  - All database access through existing Django ORM queries with tenant filtering
- [x] File Management
  - No file operations
- [x] Memory Management
  - View callables cached at module level (stateless, reusable)
  - Follows existing MCP delegation pattern; CSRF bypass for internal requests

[RHCLOUD-47086]: https://redhat.atlassian.net/browse/RHCLOUD-47086?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RHCLOUD-47087]: https://redhat.atlassian.net/browse/RHCLOUD-47087?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Add write-capable MCP tools for RBAC that delegate to existing Django views, gated by a new MCP write enable flag.

New Features:
- Introduce eight MCP write tools for creating groups, roles, role bindings, workspaces, and cross-account requests via existing RBAC APIs.
- Add a new MCP write mode flag to expose or hide write tools and enforce write call rejection when disabled.

Enhancements:
- Extend MCP tool registration and request cloning infrastructure to support authenticated POST delegation, CSRF bypass, and write/read classification.
- Factor out shared group UUID resolution logic for reuse across read and write tools.

Tests:
- Add MCP test coverage for write-mode gating, authentication requirements, and successful execution paths for new V1 and V2 write tools.


lpichler:
Mypy is failing:
```
py312: OK ✔ in 3 minutes 6.93 seconds
mypy: install_deps> python -I -m pip install mypy types-python-dateutil
mypy: commands[0]> mypy --install-types --non-interactive rbac
rbac/management/mcp_views.py:852: error: Argument "uuid" to "_call_view" has incompatible type "str | None"; expected "str"  [arg-type]
rbac/management/mcp_views.py:3139: error: Argument "uuid" to "_call_view_write" has incompatible type "str | None"; expected "str"  [arg-type]
rbac/management/mcp_views.py:3174: error: Argument "uuid" to "_call_view_write" has incompatible type "str | None"; expected "str"  [arg-type]
Found 3 errors in 1 file (checked 334 source files)
```

rbac/management/mcp_views.py
```diff
@@ -3171,19 +3486,30 @@ def _is_v2_available() -> bool:
...
-        }
-        for tool in _get_tools()
-        if v2_available or _TOOL_CONFIG.get(tool.name, ToolConfig(fn=lambda: "")).api_version != ApiVersion.V2
-    ]
+    tools_data = []
+    for tool in _get_tools():
+        config = _TOOL_CONFIG.get(tool.name, ToolConfig(fn=lambda: ""))
+        if not v2_available and config.api_version == ApiVersion.V2:
+            continue
+        if not write_enabled and config.write:
```
lpichler: @dsova06 In write mode enable, LLM has to be able to generate suggestion for edit - LLM has to know which write are available regardless it is write mode enabled or disabled. So maybe it would work better if this check would in call of each write tool and write tools with description would be available. What do you think about ?

rbac/rbac/settings.py
```diff
@@ -639,6 +639,7 @@
 # MCP settings
 MCP_TOOL_TIMEOUT_SECONDS = ENVIRONMENT.int("MCP_TOOL_TIMEOUT_SECONDS", default=30)
 MCP_TOOL_MAX_WORKERS = ENVIRONMENT.int("MCP_TOOL_MAX_WORKERS", default=10)
+MCP_WRITE_ENABLED = ENVIRONMENT.bool("MCP_WRITE_ENABLED", default=False)
```
lpichler, line 642: it is possible to put it into clowder rbac, so we can control it by just app-interface - this is prefered options in those cases.

dsova06 (commented):

1. **`test_add_principals_by_name` only asserts HTTP 200, which always passes.** The MCP endpoint returns 200 for every JSON-RPC response, including errors. This test would pass even if the group name resolution failed or the principals call errored out. It would be great to at least assert `"result"` is in the response and check `isError` is `False`, like the other write tests do.

```python
def test_add_principals_by_name(self):
    response = self._call_tool(...)
    self.assertEqual(response.status_code, 200)
    data = response.json()
    self.assertIn("result", data, f"Expected result but got: {data}")
```

### Would be good to clean up

2. **No execution tests for 5 of the 8 write tools.** `add_roles_to_group`, `create_role_v1`, `create_role_bindings`, `create_workspace`, and `create_cross_account_request` have no success-path tests. The gating tests verify the infrastructure works and the tools are thin delegation to existing views, so this isn't blocking -- but even a single happy-path test per tool would catch wiring bugs (wrong URL name, wrong body shape, missing kwargs). The `create_role_bindings` tool in particular constructs a `{"bindings": [...]}` payload that's unusual enough to warrant a test confirming the view accepts it.

### Something to think about

3. **`_call_view_write` doesn't surface HTTP error status from the delegated view.** If a view returns 400 (validation error) or 403 (permission denied), the tool returns the error body as a "successful" result with `isError: false`. An MCP client would need to parse the response text to know if the create actually worked. This is consistent with how `_call_view` works for reads, so it's not new behavior -- but for write operations where the caller really needs to know if something was created, it might be worth checking the response status and setting `isError: true` for 4xx/5xx. No need to address now, just flagging as a future consideration.

Looks good otherwise, nice work -- the three-level gating design is solid and the `_resolve_group_uuid` extraction is a clean dedup.

rbac/management/mcp_views.py
dsova06: Good point -- write tools are now always visible in tools/list so the LLM knows what's available. When write mode is disabled, descriptions are prefixed with `[DISABLED -- write mode off]`. The rejection still happens at call time in `_handle_tools_call`.

rbac/rbac/settings.py
dsova06: Agreed, adding it to clowdapp makes sense for app-interface control. I'll add the env var to `deploy/rbac-clowdapp.yml` in a follow-up once we confirm the variable naming with the team.

dsova06:
@lpichler Thanks for the review -- all feedback addressed:

**Mypy errors**: Fixed all 3 by adding `assert resolved_uuid is not None` after the early-return guard on `_resolve_group_uuid`. The function contract guarantees non-None on success, but mypy couldn't narrow the type through the tuple unpacking.

**Write tools visibility**: Write tools are now always listed in `tools/list` with a `[DISABLED -- write mode off]` prefix when write mode is off. The LLM can see what's available and suggest enabling write mode. Rejection still happens at call time.

**Clowdapp**: Will add `MCP_WRITE_ENABLED` to the clowdapp in a follow-up.

Also addressed my own self-review findings:
- Strengthened `test_add_principals_by_name` assertions
- Added happy-path tests for all 5 remaining write tools (`add_roles_to_group`, `create_role_v1`, `create_workspace`, `create_role_bindings`, `create_cross_account_request`)
- Fixed a wiring bug in `create_role_bindings` -- was sending `{"bindings": ...}` but serializer expects `{"requests": ...}` (caught by adding the test)
- `_call_view_write` now wraps 4xx/5xx responses in an error structure so MCP clients can distinguish success from failure

lpichler (approved):

rbac/rbac/settings.py
dsova06: Added MCP_WRITE_ENABLED to the clowdapp -- both in the service deployment env section and in parameters, defaulting to False.

dsova06:
@lpichler All findings addressed, including the clowdapp addition. MCP_WRITE_ENABLED is now in deploy/rbac-clowdapp.yml (service env + parameters, default False). Ready to merge.

PR #2889, opened 2026-05-15
RHCLOUD-47526 - Add list_principals_by_name to mcp tooling

EvanCasey13:
## Link(s) to Jira
https://redhat.atlassian.net/browse/RHCLOUD-47526

  Add name parameter to list_principals MCP tool to search principals by display name (first name + last name), enabling
   the AI assistant to resolve users when only their human-readable name is known.

  Changes

  - rbac/management/mcp_views.py: Added name parameter to list_principals tool and _list_principals_by_name() helper
  that fetches users from BOP and filters by case-insensitive substring match on combined first/last name
  - tests/management/test_mcp_views.py: Added 4 tests for name filtering (partial match, multiple matches,
  case-insensitive, no match)

  Usage

  # Find user by display name
  list_principals(name="RBAC Normal")  # matches "RBAC Normal For V2"
  list_principals(name="John Smith")   # matches any user with "john smith" in their name

  Why

  BOP doesn't support name-based search natively, so filtering is done client-side in the MCP layer. The implementation
  fetches users in batches (500 at a time, max 2000) and filters by substring match on "{first_name} {last_name}".

## Local Testing

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

## Summary by Sourcery

Add support for searching principals by human-readable display name via the MCP list_principals tool.

New Features:
- Allow list_principals MCP tool to filter principals by display name using a new name parameter.

Tests:
- Add unit tests covering partial, multiple, case-insensitive, and no-match scenarios for the name-based principal search.


dsova06 (commented):

1. **`meta.count` can undercount when the early-break fires** -- The loop in `_list_principals_by_name` breaks as soon as `len(filtered_users) >= offset + limit` (enough results for the current page). That means `meta.count` reports "matches found so far," not the true total. If an org has 1500 users and 200 match the name filter, but the loop stops after finding 10, the AI would report "found 10 users matching 'Smith'" when there are actually 200. Sourcery flagged this too. For the MCP use case, one option is to just keep scanning all BOP pages (up to the 2000 cap) so the count is accurate within the cap, and only early-break when `bop_offset >= total_count`. The perf cost is real but bounded, and accurate counts matter for the AI to give good answers.

2. **Pagination links don't carry the `name` filter** -- The generated `next`/`previous`/`last` links only include `limit` and `offset`, so following them would return unfiltered results. In practice the MCP consumer constructs its own tool calls, so these links are mostly decorative. Still, it would be good to either include the active filters in the query string or (simpler) just not emit links for the name-filter path since they can't work correctly without re-doing the client-side filtering.

3. **Missing test for `username_only` with name filter** -- Commit c474c40 specifically added `username_only` support to the name path, which was the right fix for the Sourcery bug report. But there's no test validating that `list_principals(name="Smith", username_only="true")` returns `[{"username": "..."}]` instead of full user objects. Would be great to cover this since it was a deliberate addition.

### Would be good to clean up

4. **No test for BOP error during name search** -- The name path has its own error handling (returns a JSON error dict instead of delegating to the view). A test with `status_code: 500` on the mock would confirm this path works and documents the error format for the name-filter code path.

5. **No test for `name` + `usernames` precedence** -- The code silently ignores `name` when `usernames` is provided, falling through to the PrincipalView path. A quick test documenting this behavior would prevent future confusion about which param wins.

### Something to think about

6. **Behavior for names with only first or last name** -- `f"{first} {last}"` always produces a space in between. If a user has no last name (empty string), the full_name becomes `"john "` with a trailing space. Substring matching still works (`"john" in "john "` is True), so this isn't a bug, but worth being aware of. A `.strip()` on `full_name` would make it slightly more defensive.

Good approach overall -- BOP doesn't support name search natively, so client-side filtering is the right call. The batched fetching with a 2000-user cap is a reasonable safeguard.

dsova06 (approved):
All findings addressed cleanly across the two follow-up commits. The early-break removal, pagination link fix, conflict rejection, and added test coverage all look good. Nice work.

PR #2906, opened 2026-05-19
[RHCLOUD-35383] Add v2_role FK, indexes, and unique constraint to BindingMapping

dsova06:
## Link(s) to Jira
- [RHCLOUD-35383](https://redhat.atlassian.net/browse/RHCLOUD-35383)

**WHAT:** Add a ForeignKey from BindingMapping to RoleV2, compound indexes, and a unique constraint to prevent duplicate binding mappings.

**WHY:** BindingMapping currently stores the V2 role ID buried inside a JSON field (`mappings["role"]["id"]`) and has zero indexes or constraints. This causes inefficient lookups, no database-level prevention of duplicate bindings, and required a complex locking workaround in the dual-write handler to prevent concurrent duplicate creation.

**HOW:**
- Add nullable `v2_role` ForeignKey to BindingMapping pointing to RoleV2 (SET_NULL on delete)
- Add compound index on `(role, resource_type_namespace, resource_type_name, resource_id)` for V1 role queries used in the locking code and workspace service
- Add unique constraint on `(v2_role, resource_type_namespace, resource_type_name, resource_id)` to prevent duplicate bindings per V2 role and resource
- Data migration backfills `v2_role` from `mappings["role"]["id"]` by looking up RoleV2 records by UUID
- Update `for_role_binding()` factory method to accept optional `v2_role` parameter
- Update `sharedSystemRolesReplicatedRoleBindings` caller to pass `v2_role`

Three migrations ordered for safety: schema (add field + index) -> data (backfill) -> constraint (unique). PostgreSQL treats NULL as distinct in unique constraints, so rows without a v2_role won't conflict.

SET_NULL is used instead of CASCADE to preserve BindingMappings when a RoleV2 is deleted. The V1 `role` FK already handles cleanup via CASCADE when a V1 role is deleted. SET_NULL allows lazy V2 role re-creation from existing mappings (used by the dual-write handler's custom role migration path).

**Out of scope (follow-up):**
- Removing the resource-level locking workaround in `relation_api_dual_write_subject_handler.py` (behavioral change, separate PR)
- Replacing `mappings["role"]["id"]` JSON lookups with `v2_role` FK queries
- Setting `v2_role` in `_create_default_mapping_for_system_role`

## Local Testing
- `pipenv run tox -e lint` -- passes
- `pipenv run tox -e py312-fast -- tests.management.role.test_model.BindingMappingTests` -- all 20 tests pass (6 new)
- `pipenv run tox -e py312-fast` -- full suite passes
- Migrations applied and verified against local PostgreSQL: FK column, compound index, and unique constraint all present

## Checklist
- [ ] if API spec changes are required, is the spec updated? N/A
- [ ] are there any pre/post merge actions required? No
- [x] are these changes covered by unit tests?
  - [x] if yes, are they backwards compatible? Yes -- new nullable FK, additive index and constraint
- [ ] is there known, direct impact to dependent teams/components? No

- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

[RHCLOUD-35383]: https://redhat.atlassian.net/browse/RHCLOUD-35383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rbac/management/migrations/0089_backfill_bindingmapping_v2_role.py
+++ b/rbac/management/migrations/0089_backfill_bindingmapping_v2_role.py
@@ -1,0 +1,55 @@
+import itertools
+
+from django.db import migrations, transaction
+
+
+def backfill_v2_role(apps, schema_editor):
+    """Backfill v2_role FK from mappings JSON field."""
+    BindingMapping = apps.get_model("management", "BindingMapping")
+    RoleV2 = apps.get_model("management", "RoleV2")
+
+    needed_uuids = set(
+        BindingMapping.objects.filter(v2_role__isnull=True)
+        .exclude(mappings__role__id=None)
+        .values_list("mappings__role__id", flat=True)
+    )
+
+    v2_role_lookup = {str(r.uuid): r.pk for r in RoleV2.objects.filter(uuid__in=needed_uuids).only("id", "uuid")}
+
+    unprocessed_pks = list(BindingMapping.objects.filter(v2_role__isnull=True).values_list("pk", flat=True))
+
+    for pk_chunk in itertools.batched(unprocessed_pks, 1000):
+        with transaction.atomic():
+            rows = BindingMapping.objects.filter(pk__in=pk_chunk, v2_role__isnull=True).select_for_update()
+
+            batch = []
+            for bm in rows:
+                role_data = bm.mappings.get("role")
+                if not role_data or "id" not in role_data:
+                    continue
+
+                role_uuid = str(role_data["id"])
+                v2_role_pk = v2_role_lookup.get(role_uuid)
+                if v2_role_pk is None:
+                    v2 = RoleV2.objects.filter(uuid=role_uuid).values_list("pk", flat=True).first()
+                    if v2 is None:
+                        continue
+                    v2_role_pk = v2
+                    v2_role_lookup[role_uuid] = v2_role_pk
+
+                bm.v2_role_id = v2_role_pk
+                batch.append(bm)
+
+            if batch:
+                BindingMapping.objects.bulk_update(batch, ["v2_role_id"])
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("management", "0088_bindingmapping_v2_role_and_indexes"),
+    ]
+
+    operations = [
+        migrations.RunPython(backfill_v2_role, migrations.RunPython.noop),
+    ]
--- a/rbac/management/role/model.py
+++ b/rbac/management/role/model.py
@@ -18,7 +18,7 @@
 """Model for role management."""
 
 import logging
-from typing import Optional, Union
+from typing import Optional, TYPE_CHECKING, Union
 from uuid import uuid4
 
 from django.conf import settings
@@ -40,6 +40,9 @@ from migration_tool.models import (
 )
 
 from api.models import FilterQuerySet, TenantAwareModel
+
+if TYPE_CHECKING:
+    from management.role.v2_model import RoleV2
 
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
@@ -153,9 +156,30 @@ class BindingMapping(models.Model):
     # JSON encoding of migration_tool.models.V2rolebinding
     mappings = models.JSONField(default=dict)
     role = models.ForeignKey(Role, on_delete=models.CASCADE, related_name="binding_mappings")
+    v2_role = models.ForeignKey(
+        "management.RoleV2",
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="binding_mappings",
+    )
     resource_type_namespace = models.CharField(max_length=256, null=False)
     resource_type_name = models.CharField(max_length=256, null=False)
     resource_id = models.CharField(max_length=256, null=False)
+
+    class Meta:
+        indexes = [
+            models.Index(
+                fields=["role", "resource_type_namespace", "resource_type_name", "resource_id"],
+                name="bm_role_resource_idx",
+            ),
+        ]
+        constraints = [
+            models.UniqueConstraint(
+                fields=["v2_role", "resource_type_namespace", "resource_type_name", "resource_id"],
+                name="unique_bindingmapping_v2role_resource",
+            ),
+        ]
 
     def save(self, *args, **kwargs):
         """Validate and save this BindingMapping."""
@@ -167,7 +191,7 @@ class BindingMapping(models.Model):
         super().save(*args, **kwargs)
 
     @classmethod
-    def for_role_binding(cls, role_binding: V2rolebinding, v1_role: Union[Role, str]):
+    def for_role_binding(cls, role_binding: V2rolebinding, v1_role: Union[Role, str], v2_role: Optional["RoleV2"]):
         """Create a new BindingMapping for a V2rolebinding."""
         mappings = role_binding.as_minimal_dict()
         resource = role_binding.resource
@@ -180,6 +204,7 @@ class BindingMapping(models.Model):
         return cls(
             mappings=mappings,
             **role_arg,
+            v2_role=v2_role,
             resource_type_namespace=resource_type_namespace,
             resource_type_name=resource_type_name,
             resource_id=resource_id,
--- a/rbac/management/group/relation_api_dual_write_subject_handler.py
+++ b/rbac/management/group/relation_api_dual_write_subject_handler.py
@@ -228,7 +228,8 @@ class RelationApiDualWriteSubjectHandler:
             resource,
             **subject,
         )
-        mapping = BindingMapping.for_role_binding(binding, system_role)
+        v2_role = SeededRoleV2.objects.filter(uuid=system_role.uuid).first()
+        mapping = BindingMapping.for_role_binding(binding, system_role, v2_role=v2_role)
         self.relations_to_add.extend(mapping.as_tuples())
         return mapping
 
--- a/rbac/management/migrations/0088_bindingmapping_v2_role_and_indexes.py
+++ b/rbac/management/migrations/0088_bindingmapping_v2_role_and_indexes.py
@@ -1,0 +1,37 @@
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("management", "0087_alter_extrolerelation_ext_id_alter_exttenant_name"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="bindingmapping",
+            name="v2_role",
+            field=models.ForeignKey(
+                blank=True,
+                null=True,
+                on_delete=django.db.models.deletion.SET_NULL,
+                related_name="binding_mappings",
+                to="management.rolev2",
+            ),
+        ),
+        migrations.AddIndex(
+            model_name="bindingmapping",
+            index=models.Index(
+                fields=["role", "resource_type_namespace", "resource_type_name", "resource_id"],
+                name="bm_role_resource_idx",
+            ),
+        ),
+        migrations.AddConstraint(
+            model_name="bindingmapping",
+            constraint=models.UniqueConstraint(
+                fields=["v2_role", "resource_type_namespace", "resource_type_name", "resource_id"],
+                name="unique_bindingmapping_v2role_resource",
+            ),
+        ),
+    ]
--- a/rbac/migration_tool/sharedSystemRolesReplicatedRoleBindings.py
+++ b/rbac/migration_tool/sharedSystemRolesReplicatedRoleBindings.py
@@ -258,6 +258,7 @@ def _get_or_create_binding(
         existing_mapping.update_mappings_from_role_binding(
             role_binding.as_migration_value(force_group_uuids=group_uuids)
         )
+        existing_mapping.v2_role = v2_role
 
         return existing_mapping, role_binding
     else:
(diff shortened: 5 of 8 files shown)


dsova06 (commented):
The migration ordering (schema -> data -> constraint) is solid, and the test coverage hits the important edge cases -- NULL semantics, cascade behavior, factory backward compatibility. Clean scope with follow-ups explicitly called out.

### Would be good to clean up

1. **Data migration backfill does individual saves** -- The backfill in `0089` does `bm.save(update_fields=["v2_role_id"])` per row inside an `iterator(chunk_size=1000)` loop. The chunking helps on the read side but writes still go one at a time. If there are many BindingMapping rows in production, this could slow down the deployment window. Batching with `bulk_update` would cut the DB round-trips significantly:

```python
batch = []
for bm in BindingMapping.objects.filter(v2_role__isnull=True).iterator(chunk_size=1000):
    # ... lookup logic ...
    bm.v2_role_id = v2_role_pk
    batch.append(bm)
    if len(batch) >= 1000:
        BindingMapping.objects.bulk_update(batch, ["v2_role_id"])
        updated += len(batch)
        batch = []
if batch:
    BindingMapping.objects.bulk_update(batch, ["v2_role_id"])
    updated += len(batch)
```

### Something to think about

1. **Type hint on `v2_role` parameter** -- `for_role_binding` has type annotations for the other parameters but `v2_role=None` is untyped. An `Optional["RoleV2"] = None` would keep it consistent with the rest of the method signature.

2. **Dual-write handler still creates NULL v2_role rows** -- `_create_default_mapping_for_system_role` in `relation_api_dual_write_subject_handler.py` calls `for_role_binding(binding, system_role)` without passing `v2_role`. This is explicitly out of scope per the PR description, but worth tracking since the dual-write path is the main runtime creator of new BindingMappings -- without a follow-up, new rows will keep accumulating with NULL v2_role after the backfill runs.

Looks good otherwise. Nice clean separation of the three migrations.

dsova06:
Addressing Sourcery's feedback:

1. **`v2_role` not saved in `_get_or_create_binding`**: This is by design. `update_mappings_from_role_binding` also doesn't persist -- it just sets `self.mappings` in memory. Both branches of `_get_or_create_binding` return unsaved instances. The caller (`permission_groupings_to_v2_role_bindings`) collects them into `MigrateCustomRoleResult` and the ultimate consumer saves. Same pattern as the new-mapping branch which creates via `for_role_binding()` without saving.

2. **In-memory dict of RoleV2**: Fixed. The backfill now collects needed UUIDs from BindingMapping first, then queries only those RoleV2 records.

3. **Constraint name with spaces**: Fixed. Renamed to `unique_bindingmapping_v2role_resource`.

Also addressed self-review feedback:
- Data migration now uses `bulk_update` with batching instead of individual `save()` calls
- Added `Optional["RoleV2"]` type hint to `for_role_binding`'s `v2_role` parameter

dsova06:
@sourcery-ai resolve

randomnetcat:
Keeping two redundant ways of storing the V2 role around unnerves me a bit. I think the way to do this while keeping everything normalized would be to add a computed column for the role UUID in `BindingMapping`, then have the foreign be from the computed column onto the `RoleV2` UUID.

Before doing this, we'd also need to remove old BindingMappings for V2 tenants (and delete any new ones on conversion), since they could reference since-deleted V2 roles.

dsova06:
Good point about the redundancy @randomnetcat. I looked into the computed column approach -- PostgreSQL does support FKs on generated columns, but Django's ORM can't express it (GeneratedField can't be used as a ForeignKey). We'd need raw SQL migrations and a model that pretends the column is a regular field, which adds maintenance burden for anyone touching migrations later.

The way I see the tradeoff: the redundancy is temporary. There are only ~2 files outside the model that read `mappings["role"]["id"]` (`internal/utils.py` and the migration tool). The plan is to replace those lookups with `v2_role` FK queries in follow-up PRs, at which point the JSON role data becomes dead weight we can stop writing.

Going the computed column route solves the sync problem, but that benefit disappears once we stop writing the role into JSON. We'd be fighting the ORM to avoid a transitional state that gets cleaned up anyway.

On the orphaned references -- the data migration already handles this by skipping rows where the RoleV2 doesn't exist (they keep `v2_role=NULL`). The FK is nullable with SET_NULL, so no constraint violations.

Happy to track the follow-up cleanup as a separate ticket if that helps make the path clearer.

rbac/management/migrations/0089_backfill_bindingmapping_v2_role.py
```diff
@@ -0,0 +1,53 @@
...
+            skipped += 1
+            continue
+
+        bm.v2_role_id = v2_role_pk
+        batch.append(bm)
+
+        if len(batch) >= 1000:
+            BindingMapping.objects.bulk_update(batch, ["v2_role_id"])
+            updated += len(batch)
+            batch = []
```
randomnetcat: `itertools.batched` might be a cleaner way to express this?

rbac/management/migrations/0089_backfill_bindingmapping_v2_role.py
```diff
@@ -0,0 +1,53 @@
...
+        if role_data and "id" in role_data:
+            needed_uuids.add(str(role_data["id"]))
+
+    v2_role_lookup = {str(r.uuid): r.pk for r in RoleV2.objects.filter(uuid__in=needed_uuids).only("id", "uuid")}
+
+    batch = []
+    updated = 0
+    skipped = 0
+
+    for bm in BindingMapping.objects.filter(v2_role__isnull=True).iterator(chunk_size=1000):
```
randomnetcat: We need to lock the `BindingMapping`s `FOR UPDATE` here.

rbac/management/migrations/0089_backfill_bindingmapping_v2_role.py
```diff
@@ -0,0 +1,53 @@
...
+def backfill_v2_role(apps, schema_editor):
+    """Backfill v2_role FK from mappings JSON field."""
+    BindingMapping = apps.get_model("management", "BindingMapping")
+    RoleV2 = apps.get_model("management", "RoleV2")
+
+    needed_uuids = set()
+    for bm in BindingMapping.objects.filter(v2_role__isnull=True).iterator(chunk_size=1000):
+        role_data = bm.mappings.get("role")
+        if role_data and "id" in role_data:
+            needed_uuids.add(str(role_data["id"]))
```
randomnetcat: `BindingMapping.objects.values_list("mappings__role__id", flat=True).distinct()`?

rbac/management/migrations/0089_backfill_bindingmapping_v2_role.py
```diff
@@ -0,0 +1,53 @@
...
+    for bm in BindingMapping.objects.filter(v2_role__isnull=True).iterator(chunk_size=1000):
+        role_data = bm.mappings.get("role")
+        if not role_data or "id" not in role_data:
+            skipped += 1
+            continue
+
+        v2_role_pk = v2_role_lookup.get(str(role_data["id"]))
+        if v2_role_pk is None:
+            skipped += 1
+            continue
```
randomnetcat: There are race conditions here:
* We could find a `BindingMapping` created after we pre-loaded all of the roles (in which case we'd need to fall back to looking it up).
* Currently, we could find a `BindingMapping` that references a since-deleted role, but I believe that would be fixed by locking.

rbac/management/role/model.py
```diff
@@ -153,10 +156,31 @@ class BindingMapping(models.Model):
     # JSON encoding of migration_tool.models.V2rolebinding
     mappings = models.JSONField(default=dict)
     role = models.ForeignKey(Role, on_delete=models.CASCADE, related_name="binding_mappings")
+    v2_role = models.ForeignKey(
+        "management.RoleV2",
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="binding_mappings",
+    )
```
randomnetcat, line 165: So, logically, the correct `on_delete` value here would be `PROTECT`. As I alluded to in my initial comment, doing this would require us to fully tear down `BindingMapping`s when a tenant migrates to V2. We could proceed like this for now, but we should work on getting it updated in the near future.

rbac/management/role/model.py
```diff
@@ -167,7 +191,9 @@ def save(self, *args, **kwargs):
         super().save(*args, **kwargs)
 
     @classmethod
-    def for_role_binding(cls, role_binding: V2rolebinding, v1_role: Union[Role, str]):
+    def for_role_binding(
+        cls, role_binding: V2rolebinding, v1_role: Union[Role, str], v2_role: Optional["RoleV2"] = None
```
randomnetcat: `v2_role` seems like it shouldn't be optional?

rbac/management/migrations/0090_bindingmapping_unique_v2_role_resource.py
randomnetcat, line 1: This doesn't need to be a separate migration. The unique constraint won't apply to rows that contain a null in one of the relevant columns.

rbac/management/migrations/0089_backfill_bindingmapping_v2_role.py
dsova06: Fixed: switched to `itertools.batched` from stdlib (Python 3.12+). Removes manual batch tracking entirely.

rbac/management/migrations/0089_backfill_bindingmapping_v2_role.py
dsova06: Fixed: added `select_for_update()` on the update queryset to prevent concurrent modification during backfill.

rbac/management/migrations/0089_backfill_bindingmapping_v2_role.py
dsova06: Fixed: replaced the Python loop with `values_list("mappings__role__id", flat=True)` to collect UUIDs in a single query.

rbac/management/migrations/0089_backfill_bindingmapping_v2_role.py
dsova06: Fixed: added a fallback individual lookup when a UUID is missing from the pre-built lookup dict. This handles rows that appeared after the initial UUID collection pass. The result is cached in the lookup dict so repeated misses for the same UUID don't cause N+1 queries.

rbac/management/role/model.py
dsova06: Agreed -- PROTECT is the right long-term target. For now SET_NULL is needed because the dual-write handler's lazy migration path (relation_api_dual_write_subject_handler.py:414) deletes and re-creates CustomRoleV2 records, and PROTECT would block that deletion. Once we remove the lazy migration path (after full backfill), switching to PROTECT becomes safe. Tracking as a follow-up.

rbac/management/role/model.py
dsova06: Agreed on the direction -- v2_role should eventually be non-nullable. It depends on the backfill completing across all tenants and the lazy migration path being removed. Once those are done, we can add a migration to enforce NOT NULL. Keeping it optional for now to allow incremental rollout.

rbac/management/migrations/0090_bindingmapping_unique_v2_role_resource.py
dsova06: Fixed: merged the unique constraint from migration 0090 into 0088 (the schema migration). PostgreSQL excludes NULLs from unique constraints, so this is safe to apply before the backfill. Deleted migration 0090.

rbac/management/migrations/0089_backfill_bindingmapping_v2_role.py
```diff
@@ -0,0 +1,53 @@
...
+
+    needed_uuids = set(
+        BindingMapping.objects.filter(v2_role__isnull=True)
+        .exclude(mappings__role__id=None)
+        .values_list("mappings__role__id", flat=True)
+    )
+
+    v2_role_lookup = {str(r.uuid): r.pk for r in RoleV2.objects.filter(uuid__in=needed_uuids).only("id", "uuid")}
+
+    rows = BindingMapping.objects.filter(v2_role__isnull=True).select_for_update()
```
randomnetcat: First, does this run in a transaction? `select_for_update` is meaningless if not.

Second, this will eventually result in us holding locks on *all* of the BindingMappings; that isn't a good idea. See, e.g., `_remove_orphaned_role_bindings` for a way to avoid this (do a queryset for all of the entries, then lock and release them in batches).

rbac/management/role/model.py
randomnetcat: The *field* being optional doesn't mean this parameter has to be. Do we not want to enforce that it's being set in new `BindingMapping`s from here on out?

rbac/management/migrations/0089_backfill_bindingmapping_v2_role.py
dsova06: Fixed: restructured to batch lock-and-release. Collect all unprocessed PKs upfront, then iterate in chunks of 1000 with `transaction.atomic()` + `select_for_update()` inside each chunk. Locks are released at the end of each batch's transaction, following the same pattern as `_remove_incorrect_workspace_parent_relations`.

rbac/management/role/model.py
dsova06: Fixed: made `v2_role` a required parameter (no default). Callers must now explicitly pass it -- even if they pass `None`. Updated the dual-write handler to look up the `SeededRoleV2` by UUID for system roles, and updated all test call sites to pass `v2_role=None` explicitly.

PR #2878, opened 2026-05-13
Adding New Panels for Rbac Consumer

Ellen-Yi-Dong:
Related JIRA ticket: https://redhat.atlassian.net/browse/RHCLOUD-43550

Adding new panels for RBAC Consumer

## Summary by Sourcery

Update RBAC operations Grafana dashboard for stage with new consumer and replication monitoring panels and refreshed layout.

Enhancements:
- Rename the RBAC operations dashboard ConfigMap payload to a stage-specific JSON and update the dashboard metadata ID and version.
- Replace several legacy consumer timeseries panels with gauges tracking replication events, relations message increases, and average replication latency over 5 minutes.
- Add new latency-focused panels, including 95th and 99th percentile replication latency visualizations, and adjust panel sizing and grid positions accordingly.

PR #2908, opened 2026-05-19
Update grafana dashboard with RBAC Consumer Queries PT 2

Ellen-Yi-Dong:
Creating this PR based off of the comments made on this PR: https://github.com/RedHatInsights/insights-rbac/pull/2878#issuecomment-4449216975

Updating the query of a specific stage panel

## Summary by Sourcery

Update the RBAC operations Grafana dashboard config to point to the latest dashboard definition and refine a latency panel.

Enhancements:
- Rename the embedded RBAC Grafana dashboard JSON from the stage variant to the latest variant and update its ID and version metadata.
- Adjust the 95th percentile replication event latency panel to aggregate latency by event type and update its title to reflect the new breakdown.

### DIFF
--- a/dashboards/grafana-dashboard-insights-rbac-operations.configmap.yaml
+++ b/dashboards/grafana-dashboard-insights-rbac-operations.configmap.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 data:
-  rbac-grafana-latest.json: |-
+  rbac-grafana.json: |-
     {
       "annotations": {
         "list": [
--- a/dashboards/grafana-dashboard-insights-rbac-operations.configmap.yaml
+++ b/dashboards/grafana-dashboard-insights-rbac-operations.configmap.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 data:
-  rbac-grafana-stage.json: |-
+  rbac-grafana-latest.json: |-
     {
       "annotations": {
         "list": [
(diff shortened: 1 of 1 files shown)


dashboards/grafana-dashboard-insights-rbac-operations.configmap.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 data:
-  rbac-grafana-stage.json: |-
+  rbac-grafana-latest.json: |-
```
randomnetcat: Why is this name changing?

Ellen-Yi-Dong: Oh! During step 4 here: https://redhat.atlassian.net/wiki/spaces/PLATFORMAM/pages/101025479/Grafana 
it was just the name of the json file that I exported. I believe that it shouldn't matter if the name is different, but I can also change it back!

randomnetcat: Ah, I see. It probably makes sense to just put it back to `rbac-grafana.json`? (Since the same file will also eventually be used in prod.)

Ellen-Yi-Dong: That works for me!

lpichler (approved):

PR #2925, opened 2026-05-25
feat(mcp): always show remediation guidance in MCP instructions

lpichler:
## Summary
- Split the MCP suggestion layer into two independent instruction blocks:
  - **`_MCP_INSTRUCTIONS_REMEDIATION`** — always included, tells the LLM to suggest concrete fix steps after diagnosing problems (permission gaps, stale memberships, unauthorized changes, etc.)
  - **`_MCP_INSTRUCTIONS_WRITE_ACTIONS`** — only included when `MCP_WRITE_ENABLED=True`, adds the "Want me to execute this?" prompt
- Previously, the entire suggestion layer was gated behind `MCP_WRITE_ENABLED`, so read-only deployments gave no actionable remediation guidance
- Updated 3 existing tests to match the new instruction structure (remediation always present, write actions conditionally present)

## Test plan
- [x] `test_initialize_instructions_always_include_remediation_guidance` — verifies remediation guidance appears without write mode
- [x] `test_initialize_instructions_include_write_actions_when_writes_enabled` — verifies write action prompt appears when writes on
- [x] `test_initialize_instructions_omit_write_actions_when_writes_disabled` — verifies write action prompt is absent when writes off, but remediation still present
- [x] Existing honest caveats and base instruction tests still pass
- [x] Pre-commit hooks (flake8, black, trailing whitespace) pass

## Summary by Sourcery

Always include remediation guidance in MCP server instructions and gate only the write-execution prompt behind the write-enabled flag.

New Features:
- Introduce a dedicated remediation guidance instruction block that is always present in MCP instructions, regardless of write mode.
- Introduce a separate write-actions instruction block that is included only when MCP write mode is enabled.

Enhancements:
- Update MCP instruction composition to include remediation guidance by default and conditionally append write-action prompts based on the MCP write flag.

Tests:
- Add and update MCP initialization tests to assert remediation guidance is always included and write-action prompts appear or are omitted according to the MCP write-enabled setting.

### DIFF
--- a/rbac/management/mcp_views.py
+++ b/rbac/management/mcp_views.py
@@ -4912,33 +4912,40 @@ _MCP_INSTRUCTIONS_HONEST_CAVEATS = (
     "accurate guidance."
 )
 
-_MCP_INSTRUCTIONS_SUGGESTION_LAYER = (
-    "\n\n## Suggestion Layer\n\n"
-    "After completing a readonly analysis, present the user with numbered write-action "
-    "options they can select from. Format suggestions as:\n\n"
-    '"Want me to: (1) <action>, or (2) <action>, or (3) <action>? '
-    "Reply 1, 2, or 3 -- or 'no'.\"\n\n"
-    "Guidelines:\n"
-    "- Always include a 'do nothing' or 'audit first' option when the action is irreversible.\n"
-    "- For permission gaps: offer to add the user to an existing group with the right role, "
-    "create a narrow custom role, or add the role to the user's current group.\n"
-    "- For group dissolution: offer immediate deletion, a transition group for stranded members, "
+_MCP_INSTRUCTIONS_REMEDIATION = (
+    "\n\n## Remediation Guidance\n\n"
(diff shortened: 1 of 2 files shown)


dsova06 (approved):
Clean split -- the distinction between "here's what you could do" (remediation) and "want me to do it now?" (write actions) is the right decomposition. Good call having the writes-disabled test also assert that remediation guidance is still present.

### Something to think about

1. **Sourcery's review-only scenario flag** -- Sourcery noted a tension between the review-only carve-out in remediation ("do NOT suggest write actions") and the write-actions section ("offer to execute them directly"). In practice an LLM would respect the scenario-specific exception over the general instruction, so I don't think this is a real problem. But if you wanted to be belt-and-suspenders about it, a short qualifier in `_MCP_INSTRUCTIONS_WRITE_ACTIONS` like "When presenting remediation options (other than review-only scenarios), offer to execute them directly" would make it unambiguous. Not blocking on this.

LGTM

PR #2927, opened 2026-05-25
[RHCLOUD-47930] feat(mcp): auto-resolve display names to usernames in MCP tools

lpichler:
## Problem

When an LLM (especially Gemini) receives a natural language prompt referencing a user by display name, it passes that name directly as the `username` parameter to MCP tools. Since the Principal model only stores usernames (e.g. `jsmith`), not display names (e.g. "Joe Smith"), the tool fails on the first attempt:

```
❯ Joe is in the 'Compliance Auditors' group AND the 'Compliance Admins' group.
  He says he still can't edit compliance policies. What's blocking him?

→ User 'Joe' not found in this organization. Please double-check the username.
  You can use list_principals(usernames='<partial>', match_criteria='partial') to search for the user.
```

The LLM then has to make a second call to `list_principals(name='Joe')` to discover the real username, wasting a round-trip and making the experience feel broken.

## Solution

Add a `_resolve_principal()` helper — mirroring the existing `_resolve_group_uuid()` pattern for groups — that auto-resolves display names to usernames server-side:

1. Try exact username match in the DB
2. If not found, fall back to BOP display name search
3. If **exactly 1 match** → auto-resolve (return the principal)
4. If **multiple matches** → return candidates list for disambiguation
5. If **0 matches** → return the existing "not found" error with hint

Applied to 4 MCP tools: `check_user_permission`, `get_user_state`, `investigate_user_access`, `guide_user_access_delegation`.

### Single match — auto-resolved (first try success)

**Prompt:** *"Joe can't edit compliance policies. What's blocking him?"*

Before (2 round-trips):
```
1. investigate_user_access(username='Joe') → ❌ "User 'Joe' not found"
2. list_principals(name='Joe') → finds jsmith
3. investigate_user_access(username='jsmith') → ✅ result
```

After (1 round-trip):
```
1. investigate_user_access(username='Joe') → resolves to 'jsmith' → ✅ result
```

### Multiple matches — returns candidates for disambiguation

**Prompt:** *"What permissions does John have?"*

```json
{
  "error": "Multiple users match 'John'. Please specify the exact username.",
  "candidates": [
    {"username": "jsmith", "name": "John Smith"},
    {"username": "jdoe", "name": "John Doe"}
  ]
}
```

### Exact username — no change (still works as before)

**Prompt:** *"Check if jsmith has cost-management:cost_model:write"*

```
check_user_permission(username='jsmith') → ✅ direct DB match, no BOP call
```

## Test plan
- [x] All 19 `MCPInvestigateUserAccessTests` pass
- [x] All 4 `MCPCheckUserPermissionTests` pass
- [x] All 16 `MCPGetUserStateTests` pass
- [x] All 13 `MCPGuideUserAccessDelegationTests` pass
- [x] Pre-commit hooks pass (flake8, black, trailing whitespace, django-upgrade)
- [ ] Manual test: call `investigate_user_access(username='<first_name>')` and verify auto-resolution


lpichler:
## Code Review Findings

### 1. [Low] Single BOP match with no local Principal falls through silently in V1 path (Info)

**Location**: `mcp_views.py`, `check_user_permission()` V1 path (lines 2654-2667)

When `_lookup_principal()` returns `(None, None, None)` -- which happens when BOP returns exactly 1 match but that user has no `Principal` record in the local DB -- the V1 path proceeds to call `_check_user_permission_v1()` with the original unresolved display name. This is functionally correct (V1 will just fail to find the user), but it means the improved "not found" hint from the V2 path is not shown. The V1 path effectively swallows the resolution attempt silently. This is a minor UX inconsistency rather than a bug.

---

### 2. [Low] `_lookup_principal` uses `limit=5` which could miss disambiguation candidates (Info)

**Location**: `mcp_views.py`, `_lookup_principal()` (line 416)

The BOP search in `_lookup_principal` uses `limit=5`, which means if more than 5 users match a display name substring, only the first 5 are returned as candidates. This is a reasonable practical limit, but the disambiguation response doesn't indicate that there may be more matches. The user could be confused if their intended user isn't in the candidate list.

**Suggestion**: Consider adding a note like `"note": "Showing first 5 matches. Refine your search if your user is not listed."` when candidates are returned.

---

### 3. [Low] Missing test for `guide_user_access_delegation` display name resolution (Info)

**Location**: `tests/management/test_mcp_views.py`

Tests are added for `check_user_permission`, `get_user_state`, and `investigate_user_access`, but not for `guide_user_access_delegation`. The `guide_user_access_delegation` implementation differs from the others (it doesn't early-return on multiple matches, instead stores them in `result["user_info"]`), so it would benefit from dedicated test coverage.

EvanCasey13 (approved):
LGTM!

PR #2928, opened 2026-05-25
feat(mcp): append honest caveats to tool results

lpichler:
## Summary

- Llama Stack's MCP provider discards the `initialize` response, so caveats in `instructions` never reach the LLM
- Adds a `caveats` field to `ToolConfig` and `register_tool()` to carry per-tool caveat text
- When a tool has caveats, `_handle_tools_call` appends them as a **separate MCP content block** (`content[1]`), preserving JSON parseability in `content[0]`
- 14 tools now carry result-time caveats covering the 4 global honest caveats (permission-to-UI mapping, ResourceDefinition opacity, org admin implicit access, V1/V2 model) plus per-tool specifics

## Root Cause

Traced through the full chain: RBAC MCP server → Llama Stack `remote::model-context-protocol` provider → Lightspeed Stack → LLM. Llama Stack calls `session.initialize()` but discards the return value ([`mcp.py:182`](https://github.com/meta-llama/llama-stack/blob/main/llama_stack/providers/utils/tools/mcp.py#L182), [`mcp.py:324`](https://github.com/meta-llama/llama-stack/blob/main/llama_stack/providers/utils/tools/mcp.py#L324)). The `instructions` field — where honest caveats lived — is never forwarded to the LLM.

## Solution

The only data paths reliably reaching the LLM are **tool descriptions** (`tools/list`) and **tool results** (`tools/call`). Descriptions already contain `Caveats:` blocks for tool selection guidance. This PR adds the same caveats to tool results so they're in the LLM's immediate context when formulating its answer.

## Test plan

- [x] `test_tool_result_includes_caveats_in_separate_content_block` — verifies tools with caveats produce 2 content blocks, with caveat text in `content[1]`
- [x] `test_tool_result_without_caveats_has_single_content_block` — verifies tools without caveats (e.g., `hello`) produce only 1 content block
- [x] Updated `test_tools_call_list_principals_success` to validate caveat content block
- [x] All 362 MCP tests pass
- [x] Lint and black pass

RHCLOUD-47930

## Summary by Sourcery

Add per-tool caveats to MCP tool configuration and include them alongside tool results so the LLM receives honest usage limitations with each call.

New Features:
- Introduce a caveats field on MCP ToolConfig and register_tool to carry per-tool caveat text that is returned with tool outputs.

Enhancements:
- Append caveat text as a separate MCP content block in tools/call responses while preserving JSON-only output in the primary content block.
- Annotate multiple RBAC MCP tools with detailed, tool-specific caveats about permissions, org admin behavior, resource definitions, and model versions.

Tests:
- Add tests verifying that tools with caveats return an additional caveat content block and tools without caveats return a single content block, and update existing list_principals tools/call test to assert the new behavior.

### DIFF
--- a/rbac/management/mcp_views.py
+++ b/rbac/management/mcp_views.py
@@ -217,6 +217,7 @@ class ToolConfig:
     required_relation: str | None = None
     required_resource_type: str = "tenant"
     v1_permission: tuple[str, str] | None = None
+    caveats: str = ""
 
 
 _TOOL_CONFIG: dict[str, ToolConfig] = {}
@@ -231,6 +232,7 @@ def register_tool(
     required_relation: str | None = None,
     required_resource_type: str = "tenant",
     v1_permission: tuple[str, str] | None = None,
+    caveats: str = "",
 ) -> Callable[[Callable[..., Any]], Callable[..., Any]]:
     """Register a tool with both FastMCP and _TOOL_CONFIG.
 
@@ -256,6 +258,7 @@ def register_tool(
             required_relation=required_relation,
             required_resource_type=required_resource_type,
             v1_permission=v1_permission,
+            caveats=caveats,
         )
 
         if passes_request:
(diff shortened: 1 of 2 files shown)


dsova06 (commented):
Nice root-cause analysis on the Llama Stack `initialize` discard -- tracing through the full chain to find that `session.initialize()` return value is thrown away is good detective work. The auto-extraction approach in the second commit is a clean evolution that avoids the duplication concern entirely.

The implementation is solid: separate content block preserves JSON parseability in `content[0]`, the `caveats` parameter on `register_tool()` provides an escape hatch if a tool ever needs to diverge from its description caveats, and the test coverage hits both paths.

### Would be good to clean up

1. **Redundant assertion in `test_tools_call_list_principals_success`** -- `assertGreaterEqual(len(result["content"]), 1)` on line 381 is made redundant by `assertEqual(len(result["content"]), 2)` four lines later. The `>=1` reads like the old assertion that got left behind when the `==2` check was added. Dropping it would keep the test's intent focused (Sourcery flagged this too).

lpichler:
Thanks @dsova06 , done!

dsova06 (commented):
LGTM

dsova06 (approved):

PR #2932, opened 2026-05-26
[RHCLOUD-47926] V2 system role update/delete returns 400 instead of 404

dsova06:
## Link(s) to Jira
- [RHCLOUD-47926](https://issues.redhat.com/browse/RHCLOUD-47926)

The V2 role API returned 404 when attempting to update or delete a system (seeded) role. This was misleading -- the role exists and is visible in list/retrieve, but `get_queryset()` filtered to `type=CUSTOM` for all write actions, making seeded roles invisible to DRF's `get_object()` before validation could run.

The V1 API handles this correctly: the queryset includes system roles, and the serializer rejects updates with 400 ("System roles may not be updated").

**Changes:**
- **`get_queryset()`**: Only restrict to `type=CUSTOM` for `create` action. Update/delete actions now see all assignable roles (CUSTOM + SEEDED), so DRF finds the role and validation can provide the correct error.
- **`perform_atomic_update`**: After `get_object()`, validates that the role is CUSTOM. Non-custom roles get a 400 with "System roles may not be updated."
- **`_perform_bulk_destroy`**: Pre-checks requested IDs for non-custom assignable roles before calling the service. Returns 400 with "System roles may not be deleted." Also changed the `CustomRoleRequiredError` fallback handler from 500 to 400 with proper Problem JSON format.

**Before:** PUT on seeded role -> 404. Bulk delete of seeded role -> 404.
**After:** PUT on seeded role -> 400 "System roles may not be updated." Bulk delete of seeded role -> 400 "System roles may not be deleted."

Platform roles remain 404 (excluded by `assignable()` filter, so they are truly invisible).

## Local Testing

```bash
# Run targeted tests
pipenv run tox -e py312-fast -- tests.management.role.test_v2_view

# Key tests to verify:
# - test_update_seeded_role_returns_400: PUT on a seeded role returns 400
# - test_delete_seeded: bulk delete of seeded role returns 400
# - test_delete_mix_of_custom_and_seeded_returns_400: mixed delete returns 400, nothing deleted
# - test_update_platform_role_returns_404: platform roles still return 404
# - test_update_role_success: custom role update still works (200)
```

## Checklist

- [x] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [x] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

[RHCLOUD-47926]: https://redhat.atlassian.net/browse/RHCLOUD-47926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Align V2 role management behavior with V1 by returning proper validation errors when attempting to modify system roles instead of incorrectly reporting them as missing.

Bug Fixes:
- Ensure updating a seeded system role via the V2 API returns HTTP 400 with a clear error instead of HTTP 404.
- Ensure bulk deletion requests that include seeded system roles return HTTP 400 and do not delete any roles.
- Return HTTP 400 with Problem+JSON formatting when a non-custom role deletion triggers a CustomRoleRequiredError.

Tests:
- Add and update V2 role view tests to cover 400 responses for updating and deleting seeded roles and mixed custom/seeded bulk deletes, while preserving 404 behavior for platform roles.

### DIFF
--- a/rbac/management/role/v2_view.py
+++ b/rbac/management/role/v2_view.py
@@ -64,7 +64,7 @@ class RoleV2ViewSet(AtomicOperationsMixin, BaseV2ViewSet):
     DEFAULT_CREATE_UPDATE_FIELDS = {"id", "name", "description", "permissions", "last_modified"}
 
     def get_queryset(self):
-        """Return assignable roles for the requesting tenant. Restricts writes to custom roles."""
+        """Return assignable roles for the requesting tenant. Restricts creates to custom roles."""
         if self.action == "retrieve":
             fields = RoleV2Service.DEFAULT_RETRIEVE_FIELDS
         else:
@@ -73,6 +73,8 @@ class RoleV2ViewSet(AtomicOperationsMixin, BaseV2ViewSet):
 
         if self.action in ("list", "retrieve"):
             return base_qs.excluding_out_of_scope_v2_roles()
+        if self.action in ("update", "bulk_destroy"):
+            return base_qs
         return base_qs.filter(type=RoleV2.Types.CUSTOM)
 
     def get_serializer_context(self):
@@ -156,6 +158,9 @@ class RoleV2ViewSet(AtomicOperationsMixin, BaseV2ViewSet):
         with atomic_block():
             instance = self.get_object()
 
+            if instance.type != RoleV2.Types.CUSTOM:
+                raise ValidationError({"uuid": "System roles may not be updated."})
+
             audit_log = AuditLog()
             audit_log.log_edit(request=request, resource=AuditLog.ROLE_V2, object=instance)
 
@@ -193,6 +198,27 @@ class RoleV2ViewSet(AtomicOperationsMixin, BaseV2ViewSet):
 
         ids = set(serializer.validated_data["ids"])
 
+        if ids:
+            non_custom = (
+                RoleV2.objects.for_tenant(request.tenant)
+                .assignable()
+                .filter(uuid__in=ids)
+                .exclude(type=RoleV2.Types.CUSTOM)
+            )
+            if non_custom.exists():
+                return Response(
+                    v2response_error_from_errors(
+                        errors=[
+                            {
+                                "detail": "System roles may not be deleted.",
+                                "status": status.HTTP_400_BAD_REQUEST,
+                                "source": "ids",
+                            }
+                        ],
+                    ),
+                    status=status.HTTP_400_BAD_REQUEST,
+                )
+
         try:
             removed_roles = service.bulk_delete(ids, from_tenant=self.request.tenant)
 
@@ -207,12 +233,18 @@ class RoleV2ViewSet(AtomicOperationsMixin, BaseV2ViewSet):
                 status=status.HTTP_404_NOT_FOUND,
             )
         except CustomRoleRequiredError as e:
-            # This should be impossible. We constrain deletions to be from the user's tenant. Non-custom roles should
-            # only exist in the public tenant, and there shouldn't be any users in the public tenant. Thus,
-            # we will never find any non-custom roles to try to delete.
             return Response(
-                {"title": "An internal error occurred.", "detail": str(e)},
-                status=status.HTTP_500_INTERNAL_SERVER_ERROR,
+                v2response_error_from_errors(
+                    errors=[
+                        {
+                            "detail": str(e),
+                            "status": status.HTTP_400_BAD_REQUEST,
+                            "source": "ids",
+                        }
+                    ],
+                    exc=e,
+                ),
+                status=status.HTTP_400_BAD_REQUEST,
             )
 
         # We don't allow a revert here because sending a notification in the middle could fail. This would make it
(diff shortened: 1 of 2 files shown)


dsova06 (commented):
This fixes a real UX problem -- getting "not found" for a role you can see in the list is confusing. The approach is sound: widen the queryset so DRF can find the object, then validate at the action level. Platform roles correctly remain 404 since they're excluded by `assignable()`.

Tests cover the important cases well: seeded update (400), seeded delete (400), mixed custom+seeded delete (400 with nothing deleted), and platform update (still 404). The audit log placement after the type check is correct too -- no logging for rejected operations.

### Something to think about

1. **The `CustomRoleRequiredError` handler is now a fallback.** The view pre-check in `_perform_bulk_destroy` catches seeded roles first because it uses `for_tenant()` (which includes the public tenant), while the service's `bulk_delete` filters by `tenant=from_tenant` only. So seeded roles from the public tenant would never reach `CustomRoleRequiredError` through that path -- they'd hit `RolesNotFoundError` instead. Keeping the handler as defense-in-depth makes sense, but worth knowing it's effectively unreachable for the normal seeded role case.

Self-review -- LGTM

dsova06:
@sourcery-ai resolve

lpichler:
## Code Review Findings

### Issue 1: Inconsistent error response format between update and delete (Low severity) (Info)

**Location**: `rbac/management/role/v2_view.py`, `perform_atomic_update()` (line ~161 in PR)

The update path raises `ValidationError({"role": "System roles may not be updated."})`, which goes through DRF's exception handler pipeline (`custom_exception_handler_v2`). This produces a Problem JSON response where the error field is `"role"`.

The delete path manually constructs the response via `v2response_error_from_errors()` with `"source": "ids"`, producing an error with field `"ids"`.

While both return HTTP 400 and Problem JSON format, the error field attribution is inconsistent:
- **Update**: field = `"role"` (arbitrary key from the ValidationError dict)
- **Delete**: field = `"ids"` (the actual input field name)

This is a minor stylistic inconsistency. The `"role"` source in the update error doesn't correspond to an actual request field. Consider using `"uuid"` or omitting the field key, or constructing the response manually like the delete path does for consistency.

**Verdict**: Low priority, not a blocker. The test `test_update_seeded_role_returns_400` validates the response contains "System roles may not be updated" but does not check the error field name.

---

### Issue 2: Audit log is written before validation in perform_atomic_update (Very low severity) (Info)

**Location**: `rbac/management/role/v2_view.py`, `perform_atomic_update()` -- but this is a pre-existing issue, not introduced by this PR.

In the current master code, the audit log (`audit_log.log_edit`) is called after `get_object()` but before serializer validation. The PR inserts the system role check between `get_object()` and the audit log, which is correct placement. The `ValidationError` will cause the transaction to roll back, so no audit log entry persists. No issue with this PR's change.

---

### Issue 3: No explicit test for Problem JSON Content-Type on update 400 response (Info)

**Location**: `tests/management/role/test_v2_view.py`, `test_update_seeded_role_returns_400`

The test checks `response.status_code`, `response.data["status"]`, and string content, but does not verify `response["Content-Type"] == "application/problem+json"` -- unlike the existing `test_update_platform_role_returns_404` which does check Content-Type. This is a minor test coverage gap.

**Verdict**: Very low priority, not a blocker.

dsova06:
Thanks for the review, Libor.

**Issue 1** -- Fixed. Changed the error source from `"role"` to `"uuid"` to match the lookup field. Pushed in e9a999eb.

**Issue 2** -- Agreed, pre-existing and not introduced here. The type check is placed before the audit log intentionally so rejected operations are never logged.

**Issue 3** -- This assertion is already present in the test (line 1904): `self.assertEqual(response["Content-Type"], "application/problem+json")`. No change needed.

rbac/management/role/v2_view.py
```diff
@@ -73,7 +73,9 @@ def get_queryset(self):
 
         if self.action in ("list", "retrieve"):
             return base_qs.excluding_out_of_scope_v2_roles()
-        return base_qs.filter(type=RoleV2.Types.CUSTOM)
+        if self.action == "create":
```
astrozzc:
```
if self.action in ("update", "bulk_destroy"):
    return base_qs
return base_qs.filter(type=RoleV2.Types.CUSTOM)
```

let's make it more explicit, which is safer, in case there is partial_update, bulk_create, etc

astrozzc (approved):
LGTM

rbac/management/role/v2_view.py
dsova06: Fixed: flipped the logic so CUSTOM is the default and only update/bulk_destroy explicitly get the wider queryset.

dsova06:
Closing in favor of #3012. This PR was approved by @astrozzc but the branch could not be updated -- rebasing, merging master, and GitHub's "Update branch" button all fail due to the repository's verified-signature rule conflicting with unsigned commits on master. Recreated from a fresh branch with the same cherry-picked commits.

PR #2933, opened 2026-05-26
RHCLOUD-47246 - MCP - Add username filter to v1 search roles and add ignore filter to…

EvanCasey13:
… v2 search roles

## Link(s) to Jira
https://redhat.atlassian.net/browse/RHCLOUD-47246

Added the username parameter to the search_roles MCP tool

For V1 returns roles held by user jdoe via group membership and for V2 returns all roles + ignored_filters with guidance to use list_role_bindings

## Local Testing

```
pipenv run tox -e py312-fast --
  tests.management.test_mcp_views.MCPSearchRolesTests.test_search_roles_v1_filter_by_username
  tests.management.test_mcp_views.MCPUnifiedSearchRolesV2Tests.test_search_roles_v2_username_ignored_with_guidance
```

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

## Summary by Sourcery

Add username filtering support to the unified MCP search_roles tool, wiring it through to V1 role search and surfacing guidance for V2 orgs where the filter is not supported.

New Features:
- Support a username parameter on the unified search_roles MCP tool to filter roles for a specific user in V1 orgs and expose this capability in the tool description.

Enhancements:
- For V2 orgs, return username as an entry in ignored_filters with guidance to use list_role_bindings to discover user role bindings.
- Extend the search_roles MCP tool documentation string to describe the new username behavior and V1/V2 limitations.

Tests:
- Add MCP tests to verify V1 username-based role filtering and V2 ignored_filters behavior and guidance when username is provided.


dsova06 (commented):
### Would be good to clean up

1. **Use `setdefault` for `ignored_filters` in the V2 path.** This is the first use of the `ignored_filters` pattern in the MCP layer, so it sets the precedent. Right now `result["ignored_filters"] = {...}` works fine since the V2 role list response never includes that key. But if we add more MCP-layer ignored filters later (e.g., `display_name` is also V1-only), each one would overwrite the previous. Using `result.setdefault("ignored_filters", {})["username"] = {...}` would make the pattern safe for composition from the start. Sourcery flagged this too -- their reasoning is right even though the immediate risk is low.

2. **Manual cleanup in `test_search_roles_v1_filter_by_username` is unnecessary.** The `policy.delete()` / `group.delete()` / `principal.delete()` calls at the end of the test body can be dropped -- `django.test.TestCase` wraps each test in a transaction that gets rolled back automatically. The `tearDown` already handles `Role` cleanup, and even that is technically redundant with the transaction rollback.

### Something to think about

1. **The `ignored_filters` structure looks good as a convention.** Having `value`, `reason`, and `alternative` with a concrete tool call example is exactly what an AI client needs to self-correct. Worth keeping this shape consistent if other unified tools need similar V1/V2 divergence handling down the road.

Looks good otherwise. The V1 wiring through the existing `get_role_queryset` username handling is the right approach, and the V2 guidance pointing to `list_role_bindings` is clear and actionable.

lpichler:
## Code Review Findings

### Low Severity (Info)

**1. V1 test cleanup relies on manual deletion at end of test body rather than tearDown**
- **File**: `tests/management/test_mcp_views.py`, `test_search_roles_v1_filter_by_username`
- **Detail**: The test creates `Principal`, `Group`, and `Policy` objects and manually deletes them at the end of the test body. The class `tearDown` only deletes `Role.objects.all()`. If the test fails before reaching the cleanup lines (e.g., an assertion fails), the `Principal`, `Group`, and `Policy` objects will leak, potentially affecting parallel test runs. The existing pattern in other test classes (e.g., `MCPViewNonAdminTests`) places all cleanup in `tearDown`. Consider moving cleanup to `tearDown` or using `self.addCleanup()` / `self.enterContext()`.

**2. Inconsistent handling of V1-only filters in V2 path**
- **File**: `rbac/management/mcp_views.py`, `_search_roles_v2`
- **Detail**: The `display_name`, `application`, and `system` parameters are also V1-only and silently ignored by V2, but they do not get an `ignored_filters` entry. The new `username` filter gets special `ignored_filters` treatment. While this is intentional for the most confusing case (username), it creates inconsistency -- an MCP consumer might wonder why `display_name` is silently dropped but `username` gets guidance. This is a pre-existing design choice and acceptable, but worth noting for future alignment.

**3. V2 test does not clean up the created RoleV2 object**
- **File**: `tests/management/test_mcp_views.py`, `test_search_roles_v2_username_ignored_with_guidance`
- **Detail**: The test creates a `RoleV2` object but does not explicitly delete it. This is handled by the class `tearDown` which calls `RoleV2.objects.all().delete()`, so this is fine -- just noting the pattern is correct.

lpichler (approved):

PR #2934, opened 2026-05-26
[RHCLOUD-47745] Enforce workspace-level role scoping in V2 role list and binding

dsova06:
## Summary

[RHCLOUD-47745](https://redhat.atlassian.net/browse/RHCLOUD-47745)

Standard (child) workspaces should only expose **workspace-granular** roles -- those whose permissions fall through to `Scope.DEFAULT` as a fallback. Roles with **explicit-DEFAULT** permissions (configured via `DEFAULT_SCOPE_PERMISSIONS`) and **permissionless** roles are restricted to the default workspace only, since they represent grants that no service checks at the standard workspace level.

### What changed

**Scope service infrastructure** (`scope_service.py`):
- Extract `_wildcard_candidates()` to deduplicate wildcard logic between `scope_for_permission` and the new `is_explicitly_scoped`
- Add `ImplicitResourceService.is_explicitly_scoped()` -- returns `True` when a permission matches any configured scope entry (root, tenant, or default), `False` for fallback-DEFAULT
- Add `PermissionScopeCache.explicit_default_ids` -- lazily-built frozenset of Permission IDs that match `DEFAULT_SCOPE_PERMISSIONS` entries, for efficient DB-level filtering

**Role list filtering** (`v2_service.py`):
- `_filter_by_resource_type()` detects standard workspaces and excludes roles containing explicit-DEFAULT permissions (via cached IDs) and permissionless roles (`permissions__isnull=False`)

**Binding validation** (`role_binding/service.py`):
- `_validate_role_scopes()` rejects explicit-DEFAULT and permissionless roles on standard workspaces
- Add `skip_scope_validation` flag to `RoleBindingService` for callers that need to bypass scope checks (e.g. workspace destroy tests)

### Visibility matrix

| Role type | Default WS | Standard WS | Root WS |
|-----------|-----------|-------------|---------|
| Workspace-granular (fallback-DEFAULT) | Yes | Yes | No |
| Explicit-DEFAULT | Yes | No | No |
| Permissionless | Yes | No | No |
| ROOT-scoped | No | No | Yes |
| TENANT-scoped | No | No | No |

## Test plan

- [x] 8 unit tests for `is_explicitly_scoped` (exact, wildcard, fallback, empty, invalid)
- [x] 4 unit tests for `PermissionScopeCache.explicit_default_ids` (with/without config, excludes root/tenant, invalidation)
- [x] 9 integration tests for role list filtering across standard/default/root workspaces with workspace-granular, explicit-DEFAULT, mixed, and permissionless roles
- [x] 4 binding validation tests (reject explicit-DEFAULT on standard, reject permissionless on standard, allow workspace-granular on standard, allow explicit-DEFAULT on default)
- [x] Workspace destroy test updated with `skip_scope_validation=True`
- [x] Full test suite passes (3581 tests, 0 failures)
- [x] Linter passes

[RHCLOUD-47745]: https://redhat.atlassian.net/browse/RHCLOUD-47745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Enforce workspace-level role scoping for role listing and role bindings, distinguishing workspace-granular permissions from explicitly DEFAULT-scoped and permissionless roles.

Enhancements:
- Refine workspace role filtering so standard workspaces only expose workspace-granular (fallback-DEFAULT) roles while default and root workspaces retain their respective visibility rules.
- Extend the scope service and permission scope cache to identify explicitly DEFAULT-scoped permissions for efficient DB-level filtering and validation.
- Tighten role-binding validation to reject explicitly DEFAULT-scoped and permissionless roles on standard workspaces, while allowing them where appropriate (e.g. default workspace).
- Allow role binding operations to bypass scope validation in specific scenarios such as workspace destruction flows.

Tests:
- Add comprehensive unit and integration tests covering explicit scope detection, explicit-DEFAULT permission caching, workspace-specific role listing behavior, and binding validation across default, standard, and root workspaces.

### DIFF
--- a/rbac/management/permission/scope_service.py
+++ b/rbac/management/permission/scope_service.py
@@ -232,6 +232,36 @@ class ImplicitResourceService:
             default_scope_permissions=parse_setting(settings.DEFAULT_SCOPE_PERMISSIONS),
         )
 
+    @staticmethod
+    def _wildcard_candidates(parsed: PermissionValue) -> list[PermissionValue]:
+        """Return the parsed value plus progressively broader wildcard variants.
+
+        Precedence order: exact, unconstrained verb, unconstrained resource type,
+        application-only.  Duplicates (when the input is already a wildcard) are
+        excluded.
+        """
+        return [parsed] + [
+            wildcard
+            for wildcard in [
+                parsed.with_unconstrained_verb(),
+                parsed.with_unconstrained_resource_type(),
+                parsed.with_application_only(),
+            ]
+            if wildcard != parsed
+        ]
+
+    def is_explicitly_scoped(self, permission: str) -> bool:
+        """Return True if the permission matches any configured scope entry.
+
+        A permission is explicitly scoped when it (or a wildcard that covers it)
+        appears in root_scope_permissions, tenant_scope_permissions, or
+        default_scope_permissions.  Permissions that fall through to the
+        ``Scope.DEFAULT`` fallback in ``scope_for_permission`` are NOT
+        explicitly scoped -- they are "workspace-granular".
+        """
+        candidates = self._wildcard_candidates(PermissionValue.parse_v1(permission))
+        return any(candidate in self._permissions_map for candidate in candidates)
+
     def scope_for_permission(self, permission: str) -> Scope:
         """
         Return the scope that a permission binds to using this object's configured permissions.
@@ -250,18 +280,7 @@ class ImplicitResourceService:
         Note that, if the permission is a wildcard, some of these steps will be redundant. For instance,
         if the permission is app:*:verb, there are only two possible matches: app:*:verb and app:*:*.
         """
-        parsed = PermissionValue.parse_v1(permission)
-
-        # If we are passed a wildcard, some wildcard checks will be redundant.
-        candidates = [parsed] + [
-            wildcard
-            for wildcard in [
-                parsed.with_unconstrained_verb(),
-                parsed.with_unconstrained_resource_type(),
-                parsed.with_application_only(),
-            ]
-            if wildcard != parsed
-        ]
+        candidates = self._wildcard_candidates(PermissionValue.parse_v1(permission))
 
         for candidate in candidates:
             scope = self._permissions_map.get(candidate)
@@ -375,14 +394,19 @@ class PermissionScopeCache:
         """Create a cache backed by the given scope service."""
         self._scope_service = scope_service
         self._ids_by_scope: dict[Scope, frozenset[int]] | None = None
+        self._explicit_default_ids: frozenset[int] | None = None
 
     def _build(self) -> dict[Scope, frozenset[int]]:
         from management.permission.model import Permission
 
         result: dict[Scope, set[int]] = {scope: set() for scope in Scope}
+        explicit_default: set[int] = set()
         for row in Permission.objects.values_list("id", "permission", named=True):
             scope = self._scope_service.scope_for_permission(row.permission)
             result[scope].add(row.id)
+            if scope == Scope.DEFAULT and self._scope_service.is_explicitly_scoped(row.permission):
+                explicit_default.add(row.id)
+        self._explicit_default_ids = frozenset(explicit_default)
         return {s: frozenset(ids) for s, ids in result.items()}
 
     @property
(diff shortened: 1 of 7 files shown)


dsova06 (commented):
Clean PR -- the abstraction split between "explicitly scoped" vs "fallback-DEFAULT" is well-chosen and the test coverage is thorough across all three workspace types.

### Would be good to clean up

1. **Two `values_list` calls per role in `_validate_role_scopes`** -- the method calls `role.permissions.values_list("permission", flat=True)` and then `role.permissions.values_list("id", flat=True)`, hitting the DB twice per role when both fields could come from a single `values_list("id", "permission")` call. Not a big deal for typical binding updates (few roles), but easy to consolidate.

### Something to think about

1. **Duplicate workspace resolution** -- both `_filter_by_resource_type` and `_validate_role_scopes` follow the same pattern: call `scope_for_resource` (which does `Workspace.objects.get()` internally), then immediately run another `Workspace.objects.filter(..., type=STANDARD).exists()` query on the same row. It's fine for now since these are cheap PK lookups, but if more call sites adopt this pattern, a `scope_for_resource` variant that returns the workspace type alongside the scope would avoid the repeated fetch. Worth keeping in mind for later.

Looks good otherwise.

dsova06:
Addressed Sourcery feedback:

1. **`_build` calling `is_explicitly_scoped` per row** -- keeping as-is. The cache builds once per process lifetime (lazy singleton). The permission table is small (hundreds of rows), and the `is_explicitly_scoped` call is a dict lookup after parsing. Not worth optimizing.

2. **Duplicate workspace type detection** -- fixed. Extracted `resolve_workspace_scope()` in `scope_service.py` that resolves scope and standard-workspace flag in a single query. Both `_filter_by_resource_type` and `_validate_role_scopes` now use it instead of `scope_for_resource` + `.exists()`.

3. **Two `values_list` calls per role** -- fixed. Consolidated into a single `values_list("id", "permission")` call in `_validate_role_scopes`.

@sourcery-ai resolve

rbac/management/permission/scope_service.py
```diff
@@ -354,6 +373,21 @@ def scope_for_resource(resource_type: str, resource_id: str, tenant: Tenant) ->
...
+    Returns ``(scope, is_standard_workspace)`` or ``None`` if the workspace
+    does not exist for the given tenant.
+    """
+    try:
+        ws_type = Workspace.objects.values_list("type", flat=True).get(id=resource_id, tenant=tenant)
+    except Workspace.DoesNotExist:
+        return None
+    if ws_type == Workspace.Types.ROOT:
+        return Scope.ROOT, False
+    return Scope.DEFAULT, ws_type == Workspace.Types.STANDARD
```
coderbydesign, line 388: In the case where `ws_type == Workspace.Types.STANDARD` evaluates `True`, this would be misleading, because the error message would be: "not scoped for this resource (Default Workspace)" here: https://github.com/RedHatInsights/insights-rbac/pull/2934/changes#diff-407d09403acc30698e7edfb36045647d88f1906a88cef7a53cf5d94e803c5410R1167

A potential solution might be to update that `scope_label` when it's a standard workspace?

tests/management/role_binding/test_service.py
```diff
@@ -2713,6 +2713,121 @@ def test_update_allows_correctly_scoped_role_on_root_workspace(self):
...
+                self.service.update_role_bindings_for_subject(
+                    resource_type="workspace",
+                    resource_id=str(self.workspace.id),
+                    subject_type="group",
+                    subject_id=str(self.group.uuid),
+                    role_ids=[str(explicit_role.uuid)],
+                )
+            self.assertIn("not scoped", str(ctx.exception))
+
+    def test_update_rejects_permissionless_role_on_standard_workspace(self):
```
coderbydesign, line 2748: Can we test the case where a role has both default + standard permissions?

rbac/management/permission/scope_service.py
dsova06: Fixed: scope_label now uses "Standard Workspace" when the target is a standard workspace instead of the generic "Default Workspace". Also strengthened existing test assertions to verify the error message label.

tests/management/role_binding/test_service.py
dsova06: Added: test_update_rejects_mixed_default_and_granular_role_on_standard_workspace -- covers a role with both explicit-DEFAULT and workspace-granular permissions being rejected on a standard workspace.

coderbydesign (approved):
Approving based on functional verification from testing locally; this appears to be behaving correctly per the AC.

PR #2942, opened 2026-05-28
[RHCLOUD-48138] Split the mixed scope roles that contains tenant scope

astrozzc:
## Link(s) to Jira
- https://redhat.atlassian.net/browse/RHCLOUD-48138

Split the mixed scope roles that contains tenant scope into two roles

## Local Testing
Unit test

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

## Summary by Sourcery

Split mixed-scope roles that include tenant-level permissions into separate tenant and workspace/root bindings and ensure system role seeding, relation replication, and group-role dual-write mappings respect these per-scope bindings.

Bug Fixes:
- Ensure roles with mixed TENANT and workspace scopes create bindings at each appropriate scope instead of collapsing to a single workspace binding, preventing loss of tenant- or workspace-scoped access.

Enhancements:
- Introduce binding scope calculations and permission splitting utilities to derive per-scope bindings from a role's permissions, including special handling for mixed TENANT and ROOT/DEFAULT scopes.
- Update role migration, dual-write handlers, and system role seeding to support multiple binding scopes per role and route permissions to the correct V2 resources and platform parent roles.

Tests:
- Add unit tests covering binding scope derivation, permission splitting logic, mixed-scope custom and system roles, and seeding behavior to verify correct creation of tenant and workspace/root bindings.

### DIFF
--- a/rbac/management/permission/scope_service.py
+++ b/rbac/management/permission/scope_service.py
@@ -304,19 +304,13 @@ class ImplicitResourceService:
         primary_scope = self.scope_for_permission(permission)
         scopes = {primary_scope}
 
-        is_wildcard = parsed.resource_type == "*" or parsed.verb == "*"
-        if not is_wildcard:
+        if not parsed.is_wildcard:
             return scopes
 
         for configured_perm, configured_scope in self._permissions_map.items():
-            if configured_perm.application != parsed.application:
-                continue
             if configured_scope == primary_scope:
                 continue
-            # Check if the configured pattern is narrower than (subsumed by) our wildcard.
-            resource_match = parsed.resource_type == "*" or parsed.resource_type == configured_perm.resource_type
-            verb_match = parsed.verb == "*" or parsed.verb == configured_perm.verb
-            if resource_match and verb_match:
+            if parsed.subsumes(configured_perm):
                 scopes.add(configured_scope)
 
         return scopes
@@ -336,15 +330,6 @@ class ImplicitResourceService:
         """Return the implicit scope for a role based on its permissions."""
         # TODO: this may need to eventually take into account resource definitions for custom roles.
         return self.highest_scope_for_permissions(a.permission.permission for a in role.access.all())
-
-    def permissions_by_scope_for_role(self, role: Role) -> dict["Scope", list[str]]:
-        """Group a role's permissions by their individual scope."""
-        result: dict[Scope, list[str]] = {}
-        for access in role.access.all():
-            perm = access.permission.permission
-            scope = self.scope_for_permission(perm)
-            result.setdefault(scope, []).append(perm)
-        return result
 
     def binding_scopes_for_role(self, role: Role) -> list["Scope"]:
         """Return the scopes at which bindings should be created for this role.
@@ -505,40 +490,6 @@ class PermissionScopeCache:
 permission_scope_cache = PermissionScopeCache(default_implicit_resource_service)
 
 
-def binding_scopes_for_permissions(
-    permissions: Iterable[str],
-    resource_service: ImplicitResourceService | None = None,
-) -> list[Scope]:
-    """Return the distinct binding scopes needed for a set of permissions.
-
-    If the permissions span multiple scopes and one of them is TENANT,
-    TENANT is kept separate and the remaining workspace-level scopes are
-    merged to the highest among them (ROOT > DEFAULT).
-
-    If only ROOT and DEFAULT are mixed, a single ROOT scope is returned
-    because workspace parent inheritance covers DEFAULT.
-
-    Wildcard permissions (e.g. ``subscriptions:*:*``) that subsume patterns
-    configured in other scopes are treated as spanning those scopes too.
-    """
-    if resource_service is None:
-        resource_service = default_implicit_resource_service
-
-    scopes: set[Scope] = set()
-    for p in permissions:
-        scopes |= resource_service.all_scopes_for_permission(p)
-
-    if not scopes:
-        return [Scope.DEFAULT]
-
-    if Scope.TENANT in scopes and len(scopes) > 1:
-        workspace_scopes = scopes - {Scope.TENANT}
-        workspace_max = max(workspace_scopes)
-        return sorted({Scope.TENANT, workspace_max}, key=lambda s: s.value)
-
-    return [max(scopes)]
-
-
 def split_permissions_by_binding_scope(
     permissions: Iterable[str],
     resource_service: ImplicitResourceService | None = None,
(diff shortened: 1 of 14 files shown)


lpichler (approved):

rbac/management/permission/scope_service.py
```diff
@@ -290,6 +290,37 @@ def scope_for_permission(self, permission: str) -> Scope:
...
+            return scopes
+
+        for configured_perm, configured_scope in self._permissions_map.items():
+            if configured_perm.application != parsed.application:
+                continue
+            if configured_scope == primary_scope:
+                continue
+            # Check if the configured pattern is narrower than (subsumed by) our wildcard.
+            resource_match = parsed.resource_type == "*" or parsed.resource_type == configured_perm.resource_type
+            verb_match = parsed.verb == "*" or parsed.verb == configured_perm.verb
```
randomnetcat: Probably worth just adding a `subsumes` helper to `PermissionValue`.

rbac/management/permission/scope_service.py
```diff
@@ -290,6 +290,37 @@ def scope_for_permission(self, permission: str) -> Scope:
...
+
+        For a wildcard like ``subscriptions:*:*``, this additionally checks whether any
+        more-specific patterns in the scope map (e.g. ``subscriptions:reports:*``) belong to
+        a different scope. If so, the wildcard effectively spans both scopes.
+        """
+        parsed = PermissionValue.parse_v1(permission)
+        primary_scope = self.scope_for_permission(permission)
+        scopes = {primary_scope}
+
+        is_wildcard = parsed.resource_type == "*" or parsed.verb == "*"
```
randomnetcat: It's possibly also worth adding a helper for this as well.

rbac/management/permission/scope_service.py
```diff
@@ -306,6 +337,24 @@ def scope_for_role(self, role: Role) -> Scope:
...
         return self.highest_scope_for_permissions(a.permission.permission for a in role.access.all())
 
+    def permissions_by_scope_for_role(self, role: Role) -> dict["Scope", list[str]]:
+        """Group a role's permissions by their individual scope."""
+        result: dict[Scope, list[str]] = {}
+        for access in role.access.all():
+            perm = access.permission.permission
+            scope = self.scope_for_permission(perm)
+            result.setdefault(scope, []).append(perm)
+        return result
```
randomnetcat: This appears to be unused?

rbac/management/permission/scope_service.py
```diff
@@ -454,3 +503,82 @@ def invalidate(self):
...
+        resource_service = default_implicit_resource_service
+
+    permissions = list(permissions)
+    if not permissions:
+        return {}
+
+    scopes_needed = binding_scopes_for_permissions(permissions, resource_service)
+
+    if len(scopes_needed) == 1:
+        return {scopes_needed[0]: permissions}
```
randomnetcat: It seems like we're duplicating most of the work `binding_scopes_for_permissions` will do anyway? I think `split_permissions_by_binding_scope` is the fundamental operation here and `binding_scopes_for_permissions` could be implemented in terms of it.

rbac/management/permission/scope_service.py
```diff
@@ -454,3 +503,82 @@ def invalidate(self):
...
+        perm_scopes = resource_service.all_scopes_for_permission(perm)
+        placed = False
+        if Scope.TENANT in perm_scopes:
+            result.setdefault(Scope.TENANT, []).append(perm)
+            placed = True
+        if perm_scopes & {Scope.ROOT, Scope.DEFAULT}:
+            result.setdefault(workspace_scope, []).append(perm)
+            placed = True
+        if not placed:
+            result.setdefault(workspace_scope, []).append(perm)
```
randomnetcat: It seems like we could assert here. This could only occur if `perm_scopes` is empty (or contains only things of the wrong type).

rbac/management/role/definer.py
```diff
@@ -354,22 +354,26 @@ def _seed_v2_role_from_v1(v1_role, display_name, description, public_tenant, pla
...
 
-        admin_scope = admin_platform_parent_scope_for_seeded_system_role(
-            v1_role.name, v1_role.admin_default, scope, apply_override=True
-        )
-        admin_platform_role = platform_roles[(DefaultAccessType.ADMIN, admin_scope)]
         if v1_role.admin_default:
-            admin_platform_role.children.add(v2_role)
-            logger.info("Added %s as child of admin platform role %s", display_name, admin_platform_role.name)
+            for scope in binding_scopes:
+                admin_scope = admin_platform_parent_scope_for_seeded_system_role(
```
randomnetcat, line 371: `v1_role.admin_default` will always be `True` here (and it looks like this is the case at other call sites as well`?

rbac/migration_tool/sharedSystemRolesReplicatedRoleBindings.py
```diff
@@ -110,16 +111,46 @@ def __post_init__(self):
 
 def v1_role_to_v2_bindings(
     v1_role: Role,
-    default_resource: V2boundresource,
+    default_resource: "V2boundresource | dict[Scope, V2boundresource]",
```
randomnetcat: This could be a `Callable[[Scope], V2boundresource]`?

rbac/migration_tool/sharedSystemRolesReplicatedRoleBindings.py
```diff
@@ -110,16 +111,46 @@ def __post_init__(self):
...
+            scope = _scope_service.scope_for_permission(permission.permission)
+            # Fall back to the highest-scope resource if scope not explicitly mapped.
+            if scope in _resource_map:
+                return _resource_map[scope]
+            # Merge ROOT/DEFAULT to highest available workspace scope.
+            for fallback in sorted(_resource_map.keys(), reverse=True):
+                if fallback != Scope.TENANT:
+                    return _resource_map[fallback]
+            # Final fallback: any available resource.
+            return next(iter(_resource_map.values()))
```
randomnetcat: When would we actually want this? We should give up if we need a scope we can't satisfy.

rbac/management/permission/scope_service.py
```diff
@@ -454,3 +488,70 @@ def invalidate(self):
...
+    if not (Scope.TENANT in union_scopes and len(union_scopes) > 1):
+        return {max(union_scopes): permissions}
+
+    workspace_scope = max(union_scopes - {Scope.TENANT})
+    result: dict[Scope, list[str]] = {}
+    for perm, perm_scopes in perm_scopes_list:
+        if Scope.TENANT in perm_scopes:
+            result.setdefault(Scope.TENANT, []).append(perm)
+        if perm_scopes & {Scope.ROOT, Scope.DEFAULT}:
+            result.setdefault(workspace_scope, []).append(perm)
```
randomnetcat: This seems equivalent to:

```python
permissions_by_scope: dict[Scope, set[str]] = {}

for perm in permissions:
    for scope in resources_service.all_scopes_for_permission(perm):
        permissions_by_scope.setdefault(scope, set()).add(perm)

if (Scope.DEFAULT in permissions_by_scope) and (Scope.ROOT in permissions_by_scope):
    permissions_by_scope[Scope.ROOT].update(permissions_by_scope[Scope.DEFAULT])
    del permissions_by_scope[Scope.DEFAULT]
```
which is probably a bit clearer? It doesn't handle future workspace scopes, but I don't think we expect to have any new ones regardless.

rbac/management/role/definer.py
randomnetcat, line 371: This can result in adding the same scope twice. (I don't think this is a problem? Just noting it.)

tests/management/role/test_definer.py
```diff
@@ -844,45 +846,38 @@ def test_force_update_relationships(self, replicate):
         )
 
         self.assertEqual(
-            initial_count,
+            count_after_first_update + 2,
             len(tuples),
-            "Expected overall number of tuples not to change.",
+            "Approval platform-default roles with mixed TENANT and workspace permissions "
+            "gain an additional platform parent binding per role.",
         )
```
randomnetcat, line 863: We should add assertions for the parent–child relationship.

tests/management/role/test_definer.py
```diff
@@ -1174,15 +1169,15 @@ def test_v2_role_parents_cleared_when_scope_changes_to_tenant(self):
         tenant_platform_role = PlatformRoleV2.objects.get(uuid=tenant_platform_role_uuid)
 
         parents = list(v2_role.parents.all())
```
randomnetcat, line 1176: We should probably assert its length here (or use `assertCountEqual`).

rbac/management/permission/scope_service.py
astrozzc: Good idea!

rbac/management/role/definer.py
astrozzc: right

tests/management/role/test_definer.py
astrozzc: good idea

rbac/management/permission/scope_service.py
```diff
@@ -560,25 +511,41 @@ def split_permissions_by_binding_scope(
...
-            placed = True
-        if perm_scopes & {Scope.ROOT, Scope.DEFAULT}:
-            result.setdefault(workspace_scope, []).append(perm)
-            placed = True
-        if not placed:
-            result.setdefault(workspace_scope, []).append(perm)
-
-    return result
+    return {
+        scope: [perm for perm in permissions if perm in scoped_perms]
```
randomnetcat: Is this not just `scope: list(scoped_perms)`?

rbac/migration_tool/sharedSystemRolesReplicatedRoleBindings.py
```diff
@@ -40,6 +41,27 @@
...
+def bound_resource_resolver_from_map(resource_map: dict[Scope, V2boundresource]) -> ScopeBoundResourceResolver:
+    """Return a resolver that maps scope to resource, with workspace-scope fallback."""
+
+    def resolve(scope: Scope) -> V2boundresource:
+        if scope in resource_map:
+            return resource_map[scope]
+        for fallback in sorted(resource_map.keys(), reverse=True):
+            if fallback != Scope.TENANT:
+                return resource_map[fallback]
+        return next(iter(resource_map.values()))
```
randomnetcat: Rather than doing this, we should probably just make it return `resource_map[scope]`, then, in providing the map, we can set `resource_map[Scope.DEFAULT] = resource_map[Scope.ROOT]` if necessary (i.e. if `ROOT` is a used scope but `DEFAULT` isn't)?

I really think picking a random resource is going to come back to bite us.

tests/management/role/test_definer.py
randomnetcat: I see you've added one; what's the second relation that's added?

rbac/management/permission/scope_service.py
astrozzc: right, order and duplication does not matter here, good catch!

rbac/migration_tool/sharedSystemRolesReplicatedRoleBindings.py
astrozzc: That's true, agree

tests/management/role/test_definer.py
astrozzc: There are two seeded approval roles, I added a comment there to make it more clear

randomnetcat (approved):

PR #2948, opened 2026-05-28
feat(mcp): pre-validate tool arguments against JSON schema [RHCLOUD-47084]

lpichler:
## Summary

Pre-validate MCP tool arguments against the tool's JSON schema (generated by FastMCP) before invoking the tool function. This catches invalid arguments at the API boundary with sanitized error messages, instead of leaking Python internals via `TypeError`.

- Validate arguments using `jsonschema.validate()` against each tool's `inputSchema`
- Inject `additionalProperties: false` to reject unknown arguments
- Return JSON-RPC `-32602 Invalid params` with human-readable messages (e.g., "Expected 'message' to be string, got int")
- No Python exception details, parameter names, or tracebacks leak to clients
- The existing `TypeError` catch remains as a safety net

## Changes

| File | What |
|------|------|
| `rbac/management/mcp_views.py` | Add `_get_tool_schemas()`, `_sanitize_validation_error()`, `_validate_tool_arguments()` and call validation in `_handle_tools_call` |
| `tests/management/test_mcp_views.py` | 13 new tests: wrong type, extra args, missing required, valid pass-through, unit tests for helpers |

## Commits

1. **feat(mcp): pre-validate tool arguments against JSON schema** — Core implementation: schema cache, sanitizer, validator, integration point
2. **test(mcp): add schema pre-validation tests** — 13 tests covering all acceptance criteria
3. **refactor: address Sourcery AI review suggestion** — Extract field name from required error message for cleaner output

## Test plan

- [x] Wrong type arguments rejected with sanitized message
- [x] Extra unknown arguments rejected
- [x] Missing required arguments rejected
- [x] Valid arguments pass through to tool execution
- [x] Optional arguments can be omitted
- [x] No internal names leaked in error messages
- [x] Existing MCP tests still pass (399 tests)
- [x] Full test suite passes
- [x] Linter passes

Resolves: [RHCLOUD-47084](https://redhat.atlassian.net/browse/RHCLOUD-47084)

[RHCLOUD-47084]: https://redhat.atlassian.net/browse/RHCLOUD-47084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lpichler:
@sourcery-ai review

dsova06 (commented):
Clean, well-scoped addition. The three-function decomposition (schema cache, sanitizer, validator) is easy to follow and the test coverage is thorough. A few things I'd want to tighten up.

1. **The `additionalProperties` sanitizer leaks jsonschema's internal phrasing.** The branch does `f"Unknown argument provided: {error.message}"`, and jsonschema's message for this case is `"Additional properties are not allowed ('foo' were unexpected)"`. The resulting client output is `"Unknown argument provided: Additional properties are not allowed ('foo' were unexpected)"` -- redundant and exposes the library's exact wording, which is what the PR is specifically trying to avoid. Would be great to parse out just the property name(s) the same way the `required` branch does, e.g., `"Unknown argument: foo"`.

2. **`additionalProperties` override is unconditional.** `{**schema, "additionalProperties": False}` will silently overwrite any existing value. FastMCP schemas don't set this today, but using `setdefault` is a one-line change that's future-proof:

   ```python
   strict_schema = dict(schema)
   strict_schema.setdefault("additionalProperties", False)
   ```

   If a tool ever legitimately accepted arbitrary kwargs, the current code would silently break it. Sourcery flagged this too -- I agree with them here.

### Would be good to clean up

3. **The regex in the `required` branch is fragile.** `re.match(r"'(.+)' is a required property", error.message)` couples the sanitizer to jsonschema's current English phrasing. `error.validator_value` gives you the required field names directly without parsing strings. The fallback handles mismatches, so this isn't urgent, but it would be more robust.

4. **`test_multiple_extra_arguments_rejected` is weaker than its neighbors.** It only checks the error code, not the message content. Adding `assertIn("Unknown argument", data["error"]["message"])` (and the negative assertions for jsonschema internals) would match the thoroughness of the other tests in this class.

Solid work otherwise -- the placement after `confirmation_token` pop and after auth checks is exactly right, and the test suite covers the important paths well.

lpichler:
Thanks @dsova06! suggestions were addressed!

dsova06 (approved):
All three substantive findings addressed cleanly. The additionalProperties sanitizer now produces focused messages, setdefault protects against future schema conflicts, and the required-field extraction uses structured attributes as the primary path. LGTM.

PR #2951, opened 2026-05-28
fix(mcp): sanitize exception messages in JSON-RPC error responses

lpichler:
## Summary
- Sanitize raw Python exception messages (`str(e)`) that were being included in JSON-RPC error responses returned to MCP clients
- These messages can leak internal file paths, database connection details, function signatures, and schema information
- Each fix follows the pattern: log the raw exception server-side for operators, return a generic message to the client

## Changes

| Location | Risk | Fix |
|----------|------|-----|
| `_check_user_permission_v1` (line 2729) | `f"Failed to check permissions: {e}"` leaked DB errors, connection strings | Generic message + `logger.warning` |
| `investigate_user_access` (line 3595) | `str(e)` stored and returned as `effective_access_error` | Generic message, existing logger retained |
| `_handle_tools_call` TypeError (line 5468) | `f"...{exc}"` leaked function signatures and param names | Strip `exc` from response + `logger.warning` |
| `_call_view_json` (line 360) | Raw HTTP response body (possibly debug HTML) as `detail` | Suppress body for 5xx; parse JSON for 4xx |
| `_check_user_permission_v1` detail (line 2740) | Unfiltered `data["detail"]` pass-through from upstream | Validate type and length before returning |

## Test plan
- [x] `tox -e lint` passes
- [x] All 386 MCP tests pass (`tests.management.test_mcp_views`)
- [x] No regressions — only `mcp_views.py` modified

## Summary by Sourcery

Sanitize MCP JSON-RPC and HTTP error responses to avoid leaking internal exception details while preserving diagnostics in server logs.

Bug Fixes:
- Prevent internal HTTP response bodies and debug content from being returned directly to MCP clients, especially for 5xx errors.
- Avoid exposing raw upstream error details and arbitrary content in permission-check responses returned to clients.
- Stop surfacing raw exception messages in effective access investigations and tool call parameter validation errors, limiting client-visible information to generic messages.

Enhancements:
- Log full exception and response details server-side when sanitizing client-facing error messages to aid operators in debugging.

### DIFF
--- a/rbac/management/mcp_views.py
+++ b/rbac/management/mcp_views.py
@@ -356,8 +356,15 @@ def _call_view_json(
     if response.status_code == 204:
         return json.dumps({"status": "no_content"})
     content = response.content.decode()
+    if response.status_code >= 500:
+        logger.warning("mcp: _call_view_json got HTTP %d from %s: %s", response.status_code, path, content[:500])
+        return json.dumps({"error": f"HTTP {response.status_code}", "detail": "Internal server error"})
     if response.status_code >= 400:
-        return json.dumps({"error": f"HTTP {response.status_code}", "detail": content})
+        try:
+            detail = json.loads(content)
+        except (json.JSONDecodeError, ValueError):
+            detail = f"HTTP {response.status_code}"
+        return json.dumps({"error": f"HTTP {response.status_code}", "detail": detail})
     return content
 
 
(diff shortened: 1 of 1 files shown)


lpichler:
Superseded by new PR with clean branch based on current upstream/master.

PR #2952, opened 2026-05-28
[RHCLOUD-47886] Add disaster recovery reconciliation for Kessel Relations

dsova06:
## Summary

After a PostgreSQL database restore, SpiceDB retains relation tuples that were created or deleted during the data loss interval. This creates orphaned tuples (resources that no longer exist in RBAC) and missing tuples (resources that were restored but whose deletions are still in SpiceDB).

This PR adds a PSK-authenticated internal endpoint that triggers a Celery task to read Kafka replication events from the data loss window, validate each against current RBAC database state, and write corrective events through the outbox table so the normal Debezium pipeline reconciles SpiceDB.

### How it works

1. Operator POSTs `restore_timestamp` and optional `buffer_seconds` to `/_private/api/disaster_recovery/reconcile/`
2. Endpoint validates input and enqueues a Celery task (returns 202 with task_id)
3. Task reads Kafka events from `[restore_timestamp - buffer, restore_timestamp]` using `offsets_for_times()` with a DR-specific consumer group
4. For each event, bulk-checks resource existence against RBAC models (one query per resource type)
5. Applies the corrective truth table:
   - `relations_to_add` + resource exists = SKIP (correct state)
   - `relations_to_add` + resource gone = corrective DELETE (orphaned tuple)
   - `relations_to_remove` + resource exists = corrective ADD (restored resource needs tuple)
   - `relations_to_remove` + resource gone = SKIP (correct state)
6. Writes corrective events to outbox with `DR_CORRECTIVE_ADD` / `DR_CORRECTIVE_REMOVE` event types
7. Returns structured summary with counts and timing

### New files

- `rbac/management/disaster_recovery/kafka_reader.py` -- Kafka consumer with timestamp-based seeking
- `rbac/management/disaster_recovery/resource_checker.py` -- Bulk resource existence checker for 6 model types
- `rbac/management/disaster_recovery/corrective_writer.py` -- Truth table logic and outbox writes
- `rbac/management/disaster_recovery/service.py` -- Orchestrator

### Configuration

- `DR_RECONCILE_ENABLED` (default: false) -- feature flag gate
- `DR_KAFKA_CONSUMER_GROUP_ID` (default: rbac-dr-consumer-group) -- isolated consumer group
- `DR_MAX_EVENTS_PER_RECONCILE` (default: 10000) -- safety cap

## JIRA

[RHCLOUD-47886](https://issues.redhat.com/browse/RHCLOUD-47886)

## Test plan

- [x] 51 unit tests covering all components (kafka reader, resource checker, corrective writer, service orchestrator, views, feature flags)
- [x] Full test suite passes (3,667 tests, no regressions)
- [x] Black formatting verified
- [ ] Manual verification in Stage after merge (QE)

## How to test manually

```bash
# With httpie (PSK auth handled by internal middleware)
http POST localhost:8000/_private/api/disaster_recovery/reconcile/ \
  restore_timestamp="2026-05-28T10:00:00Z" \
  buffer_seconds:=300
```

[RHCLOUD-47886]: https://redhat.atlassian.net/browse/RHCLOUD-47886?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Add a disaster recovery reconciliation workflow for Kessel relation data using Kafka history to generate corrective replication events after a database restore.

New Features:
- Introduce an internal POST endpoint to trigger disaster recovery reconciliation over a specified restore window.
- Add a Celery task that orchestrates disaster recovery reconciliation based on a restore timestamp and buffer window.
- Add a Kafka reader that scans replication events within a time window for disaster recovery.
- Add a resource existence checker that bulk-validates referenced RBAC resources across multiple models.
- Add a corrective event writer that derives corrective add/remove actions and emits them via the existing outbox replication pipeline.
- Define new replication event types for disaster recovery corrective add and remove actions.

Enhancements:
- Add configuration flags and limits to control disaster recovery reconciliation behavior, Kafka consumer group, and max processed events.

Tests:
- Add unit tests for the disaster recovery Kafka reader, resource checker, corrective writer, orchestrator service, internal endpoint, and feature flag behavior of the Celery task.


dsova06 (commented):

1. **Gate the view on `DR_RECONCILE_ENABLED`** -- right now, when the feature is disabled, an operator POSTing to the endpoint gets a 202 (task enqueued successfully) but the task immediately returns a "disabled" message. It would be a better experience to check the flag in the view and return a clear error response immediately. That way the operator knows the feature is off without having to look up the task result. Other feature-flagged endpoints in the codebase follow this pattern, and it avoids unnecessary Celery overhead.

2. **Validate `buffer_seconds` in `reconcile()`, not just the view** -- the view enforces `buffer_seconds >= 0`, but `reconcile()` itself accepts any integer. If someone calls the service function directly in the future (management command, Celery beat, another task), a negative value would silently invert the time window so `start_timestamp_ms > end_timestamp_ms`. A quick guard (`if buffer_seconds < 0: raise ValueError(...)` or clamp to 0) would make the service self-contained and safe for non-HTTP callers.

### Something to think about

3. **Idempotency on retry** -- each call to `replicator.replicate()` in `write_corrective_events` persists independently. If the task crashes partway through, some corrective events will have been written and on retry the full set gets written again, producing duplicates in the outbox. SpiceDB relation writes are idempotent so this won't cause incorrect state, but it might be worth a brief note in the docstring so future maintainers know re-running the reconciliation for the same window is safe.

4. **Subject-side existence** -- the resource checker validates whether the resource (object) in each tuple still exists, but doesn't check the subject side. If a subject was deleted during the data loss window and not restored, a corrective add would create a relation pointing to a non-existent subject in SpiceDB. This is a narrow edge case and SpiceDB handles dangling subjects gracefully, but worth keeping in mind for a potential future improvement.

The architecture is clean -- nice separation between the Kafka reader, resource checker, corrective writer, and orchestrator. The truth table logic is well-documented and all four branches are tested. Test coverage is thorough with both unit and integration-style tests using InMemoryLog.

rbac/management/disaster_recovery/service.py
```diff
@@ -0,0 +1,98 @@
...
+from management.disaster_recovery.kafka_reader import read_events_in_window
+from management.disaster_recovery.resource_checker import check_resources_exist
+from management.relation_replicator.outbox_replicator import OutboxReplicator
+
+logger = logging.getLogger(__name__)
+
+
+def reconcile(
+    restore_timestamp_ms: int,
+    buffer_seconds: int = 300,
```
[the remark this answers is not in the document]
dsova06: Fixed: added `if buffer_seconds < 0: raise ValueError(...)` guard at the top of `reconcile()`. Also gated the view itself on `DR_RECONCILE_ENABLED` so callers get an immediate 403 instead of a no-op task.

tests/management/disaster_recovery/test_kafka_reader.py
```diff
@@ -0,0 +1,248 @@
...
+            read_events_in_window(1000, 2000)
+        self.assertIn("not enabled", str(ctx.exception))
+
+    @override_settings(KAFKA_ENABLED=True, RBAC_KAFKA_CONSUMER_TOPIC=None)
+    def test_no_topic_raises(self):
+        with self.assertRaises(DisasterRecoveryError) as ctx:
+            read_events_in_window(1000, 2000)
+        self.assertIn("topic", str(ctx.exception))
+
+    @override_settings(
```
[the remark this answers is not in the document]
dsova06: Added: `test_empty_offsets_dict` that verifies the `if not offsets:` early-exit branch returns an empty list without calling `poll`.

dsova06:
@sourcery-ai resolve

rbac/internal/views.py
```diff
@@ -2624,3 +2624,50 @@ def recompute_tenant_role_bindings(request, org_id):
...
+
+    from datetime import datetime
+
+    from management.tasks import run_disaster_recovery_reconcile
+
+    body = load_request_body(request)
+
+    restore_timestamp_str = body.get("restore_timestamp")
+    if not restore_timestamp_str:
+        return JsonResponse({"error": "restore_timestamp is required"}, status=400)
```
lpichler, line 2648: **low**: [ISSUE-2] No concurrency guard on the reconciliation endpoint

Multiple simultaneous POST requests will each enqueue separate Celery tasks. Overlapping reconciliations waste resources and produce unreliable summary counts. While SpiceDB writes are idempotent, duplicate runs are undesirable for an operational DR tool.

**Suggestion:** Consider adding a Redis-based distributed lock or documenting that only one reconciliation should run at a time.

rbac/internal/views.py
```diff
@@ -2624,3 +2624,50 @@ def recompute_tenant_role_bindings(request, org_id):
...
+        dt = datetime.fromisoformat(restore_timestamp_str.replace("Z", "+00:00"))
+        restore_timestamp_ms = int(dt.timestamp() * 1000)
+    except (ValueError, TypeError) as e:
+        return JsonResponse({"error": f"Invalid restore_timestamp format: {e}"}, status=400)
+
+    buffer_seconds = body.get("buffer_seconds", 300)
+    if not isinstance(buffer_seconds, int) or buffer_seconds < 0:
+        return JsonResponse({"error": "buffer_seconds must be a non-negative integer"}, status=400)
+
+    task = run_disaster_recovery_reconcile.delay(
```
lpichler, line 2664: **low**: [ISSUE-3] buffer_seconds accepts boolean values

The validation 'not isinstance(buffer_seconds, int)' accepts True (=1) and False (=0) because bool is a subclass of int in Python. JSON true/false deserialize as booleans.

**Suggestion:** Add 'or isinstance(buffer_seconds, bool)' to the check, or use 'type(buffer_seconds) is not int'.

lpichler:
## Code Review Findings

**File:** `rbac/management/disaster_recovery/corrective_writer.py` (line 199)

**medium**: [ISSUE-1] Missing org_id in corrective event info causes warning spam

The corrective writer creates ReplicationEvent objects with info={'source_offset': ..., 'reason': ...} but does not include org_id. When OutboxReplicator calls event.resource_context(), it reads org_id from event_info and logs a warning when missing: 'Missing required org_id for dr_corrective_add event.' This will generate one warning per corrective event, potentially thousands during a DR reconciliation.

**Suggestion:** Include org_id in the info dict if derivable from the original Kafka event's resource context, or add DR event types to resource_context() as a special case that skips the org_id requirement.

lpichler:
**there is no per-message logging in the DR modules** -- it takes a bulk/batch approach instead of the per-message approach the RBAC Kafka consumer uses.

Here's the gap compared to the RBAC consumer:

| Aspect | RBAC Kafka Consumer | DR Reconciliation |
|--------|-------------------|-------------------|
| Per-message offset log | `partition: X, offset: Y` every message | Only on **parse errors** |
| Normal operation | 3+ log lines per message | ~2-3 lines for the **entire batch** |
| Granularity | Per-message (streaming) | Aggregated totals |

### What the DR code does log

- **Start**: `"Starting DR reconciliation: window [%d, %d] (buffer=%ds)"` -- time window in ms, no offsets
- **After Kafka read**: `"Read %d events with %d total tuples"` -- counts only
- **Completion**: One summary line with `events=%d tuples=%d adds=%d removes=%d skipped=%d errors=%d (%.3fs)`
- **On parse error only**: `"Failed to parse relation_to_add at offset=%d: %s"` -- this is the only place offsets appear

### What's missing

If a corrective write fails (`corrective_writer.py` lines 126/139), you get the relation tuple stringified in the exception log, but no offset or partition reference -- so you can't easily trace a failed correction back to the original Kafka message.

**Recommendation for the review**: The DR code should add at minimum a per-batch progress log per partition (e.g., `"Processed partition %d: offsets [%d, %d], %d events"`) so operators can correlate what was processed during a reconciliation run if something goes wrong. This is especially important for a DR procedure where traceability is critical.

dsova06:
Fixed: org_id is now extracted from the Kafka event's resource_context and included in the corrective event info dict. This flows through ParsedReplicationEvent -> CorrectiveAction -> ReplicationEvent.info, so resource_context() will find it and skip the warning.

dsova06:
Fixed: added per-partition progress logging to the Kafka reader. After consuming, it now logs partition number, event count, and offset range for each partition. This gives operators the traceability needed to correlate processed events back to Kafka during a DR run.

rbac/internal/views.py
dsova06: Acknowledged. This is an internal DR tool invoked manually during incident response, and SpiceDB writes are idempotent, so overlapping runs are wasteful but not harmful. Adding a Redis lock would add complexity for an edge case that's unlikely in practice (operators coordinate during DR). Leaving as-is for now; if we see the need in production, we can add it as a follow-up.

rbac/internal/views.py
dsova06: Fixed: added `isinstance(buffer_seconds, bool)` check before the int check. Also added a test for this edge case (`buffer_seconds: true` in JSON now returns 400).

lpichler:
## Code Review Findings

**File:** `rbac/management/disaster_recovery/corrective_writer.py` (line 201)

**medium**: [ISSUE-1] Corrective events written outside a transaction boundary

Each corrective event is written via replicator.replicate() independently, without a wrapping transaction. The AGENTS.md convention states: 'Do not perform outbox writes or external calls outside a transaction boundary.' The OutboxWAL.log() calls transaction.on_commit() which may fire immediately without an active transaction, causing the Prometheus counter to increment at the wrong time.

**Suggestion:** Wrap corrective event writes (or batches of writes) in transaction.atomic(). Since there is no corresponding business data change, this is mainly for convention consistency and correct on_commit behavior.

lpichler:
## Code Review Findings

**File:** `rbac/management/disaster_recovery/kafka_reader.py` (line 288)

**low**: [ISSUE-2] Walrus operator in _parse_debezium_payload hides assignment

The walrus operator 'isinstance(payload_data := payload_field, dict)' is unnecessary and reduces readability. The assignment is unconditional since it always assigns payload_field to payload_data.

**Suggestion:** Replace with: 'elif isinstance(payload_field, dict): payload_data = payload_field'

dsova06:
Fixed: wrapped all corrective event writes in transaction.atomic() so on_commit handlers fire at the correct time. Both the adds and removes loops now execute within a single transaction boundary.

dsova06:
Fixed: replaced walrus operator with explicit assignment. Now reads `elif isinstance(payload_field, dict): payload_data = payload_field`.

lpichler:
@dsova06 Can you add dry-mode ? so we can check which events it pertains ...

rbac/management/disaster_recovery/corrective_writer.py
```diff
@@ -0,0 +1,162 @@
+"""Generate and write corrective events for disaster recovery reconciliation."""
```
lpichler, line 1: **medium**: [1] Error handling inside transaction.atomic() in corrective_writer.py is contradictory

Per-event try/except inside a single transaction.atomic() block creates misleading error counts. If one event fails but others succeed, the error count suggests partial success, but the transaction is all-or-nothing.

**Suggestion:** Either remove per-event try/except and let the transaction fail cleanly, or use individual transaction.atomic() blocks per event for true partial success.

lpichler:
## Code Review Findings

**File:** `rbac/internal/views.py` (line 1)

**nitpick**: [4] 403 response for disabled feature flag should be 501 or 503

Feature flag disabled returns 403 (Forbidden) but semantically it is not an authorization issue.

**Suggestion:** Use 501 Not Implemented or 503 Service Unavailable instead.

lpichler:
Can add dry-mode ? so we can check which events it pertains ...

dsova06:
Fixed: added dry_run mode. Pass `"dry_run": true` in the request body. In dry-run, the reconciliation reads Kafka events, checks resource existence, and generates the corrective action plan but skips writes. The response includes an `actions` array with each planned corrective action (tuple, action type, reason, source partition/offset). Logs also include per-action DRY RUN lines.

dsova06:
Fixed: moved transaction.atomic() from outer wrapper to per-event scope. Each corrective write now has its own transaction, so a DB error on one event rolls back only that event without affecting others. Error counts are now accurate.

dsova06:
Regarding the 403 for disabled feature flag: keeping 403 -- this is consistent with other internal endpoints in the codebase that use 403 for disabled features. 501/503 would imply the endpoint itself is not implemented or the service is down, which isn't the case.

PR #2953, opened 2026-05-28
[RHCLOUD-47887] Add workspace events recovery for disaster recovery

dsova06:
## Link(s) to Jira
- [RHCLOUD-47887](https://issues.redhat.com/browse/RHCLOUD-47887)

After an RBAC database restore, downstream services (HBI and 7 others) retain workspace state that diverges from the restored RBAC state. This PR adds a disaster recovery endpoint that generates corrective workspace events to bring downstream services back in sync.

**What**: An internal endpoint (`POST /_private/api/disaster_recovery/workspaces/`) that accepts a restore timestamp, reads workspace events from Kafka for the data loss window, compares them against current RBAC DB state, and writes corrective events to the outbox table for Debezium to publish.

**Why**: During a database restore, workspace create/update/delete events that were originally published to Kafka are lost from RBAC's perspective but already consumed by downstream services. Without corrective events, downstream services have orphaned workspaces, missing workspaces, or stale workspace data.

**How**: The endpoint enqueues a Celery task that:
1. Reads workspace events from Kafka within the time window (restore_timestamp - buffer to now) using `offsets_for_times`
2. Deduplicates by workspace ID (keeps latest event per workspace)
3. Compares each event against current RBAC DB state using a truth table:
   - `create` + not in DB -> write `delete` corrective (orphaned in downstream)
   - `create` + in DB -> skip (both agree)
   - `delete` + in DB -> write `create` corrective with current state (missing in downstream)
   - `delete` + not in DB -> skip (both agree)
   - `update` + in DB -> write `update` corrective with current state (stale in downstream)
   - `update` + not in DB -> write `delete` corrective (orphaned in downstream)
4. Writes corrective events through the standard outbox table so Debezium publishes them

The feature is gated behind `DR_RECOVERY_ENABLED` (defaults to `False`).

**New files**:
- `rbac/core/kafka_dr.py` -- Reusable Kafka timestamp reader for DR scenarios
- `rbac/management/workspace/dr_recovery.py` -- Corrective event generation logic

## Local Testing

The endpoint is behind internal middleware (requires `X-RH-Identity` header with Associate type) and the `DR_RECOVERY_ENABLED` feature flag. To test:

```bash
# Enable the feature flag
export DR_RECOVERY_ENABLED=True

# POST to the endpoint (requires internal identity header)
http POST localhost:8000/_private/api/disaster_recovery/workspaces/ \
  restore_timestamp="2026-05-28T10:00:00Z" \
  buffer_minutes:=5 \
  X-RH-Identity:<base64-encoded-associate-identity>
```

Unit tests cover all 6 truth table cases, event parsing/filtering/deduplication, endpoint validation, auth, and the Celery task:
```bash
pipenv run tox -e py312-fast -- tests.management.workspace.test_dr_recovery tests.internal.test_dr_recovery_views
```

## Checklist

No migration changes required. No API spec changes needed (internal endpoint, not public API). Feature is disabled by default and has no impact until explicitly enabled during a DR scenario.

- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [ ] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [ ] File Management
- [ ] Memory Management

[RHCLOUD-47887]: https://redhat.atlassian.net/browse/RHCLOUD-47887?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Add an internal disaster recovery workflow to reconcile workspace state after RBAC DB restores by replaying Kafka events and emitting corrective workspace events.

New Features:
- Introduce an internal POST /_private/api/disaster_recovery/workspaces/ endpoint to trigger workspace disaster recovery jobs.
- Add a Celery task that reads workspace events from Kafka over a restore window and generates corrective outbox events based on current DB state.
- Provide reusable Kafka utilities for reading events by timestamp range and workspace DR-specific reconciliation logic with statistics reporting.

Enhancements:
- Add configuration flags for enabling workspace disaster recovery and overriding the workspace Kafka topic.

Tests:
- Add unit tests for workspace DR reconciliation logic, covering event parsing, filtering, deduplication, and corrective event generation scenarios.
- Add endpoint and task tests validating auth, input validation, feature flag behavior, Kafka interaction, and task result structure.


dsova06 (commented):

1. **Partitions without messages in the window get read from offset 0.** In `kafka_dr.py`, when `offsets_for_times()` returns `None` for a partition (meaning no messages at or after `start_timestamp_ms`), the code skips the `seek` but the partition stays assigned. With `auto_offset_reset=earliest` and no committed offset for the `rbac-dr-recovery` group, the consumer reads from offset 0 on that partition -- potentially millions of old messages that all get filtered out by the `< start_timestamp_ms` check. Worth pausing those partitions or seeking them to end:

```python
for tp in topic_partitions:
    offset_info = start_offsets.get(tp)
    if offset_info is None:
        consumer.pause([tp])  # nothing to read from this partition
        continue
    consumer.seek(tp, offset_info.offset)
```

2. **Consumer reads all messages past the end of the recovery window.** The `continue` on `message.timestamp > end_timestamp_ms` means the consumer reads and discards every message from the window end to the topic's latest offset across all partitions. For a production workspace topic with days of activity after a restore point, this is a lot of unnecessary I/O. Looking up end offsets with a second `offsets_for_times(end_timestamp_ms)` call and tracking per-partition progress would bound the work to just the window. Alternatively, a simple approach: track how many consecutive messages have exceeded the end timestamp, and break after a threshold (since cross-partition interleaving means a single message past the window isn't definitive).

### Would be good to clean up

3. **`__import__("json")` in the deserializer lambda** (`kafka_dr.py:50`). `json` is a standard library module -- importing it at the top of the file and referencing it directly in the lambda is cleaner and avoids the repeated dynamic import:

```python
import json
# ...
"value_deserializer": lambda m: json.loads(m.decode("utf-8")),
```

### Something to think about

4. **No guard against concurrent or duplicate execution.** If someone hits the endpoint twice (accidentally or during a chaotic DR scenario), two tasks run in parallel and both write corrective events for the same workspaces. Similarly, if the task fails halfway and gets re-run, it writes duplicates for already-processed workspaces. For a DR tool this might be acceptable if the corrective events are idempotent from the downstream consumers' perspective (a second "create" for an already-existing workspace is a no-op), but worth confirming that assumption. If not, a simple Redis-based lock on the task or deduplication key on the outbox entries would help.

---

Sourcery flagged the `Z` suffix in `datetime.fromisoformat` as a bug, but that's a non-issue -- Python 3.11+ handles `Z` natively and this project targets 3.12.

The truth table logic, test coverage, and overall architecture are well done. The corrective event tests cover all 6 cases plus edge cases like deduplication and production format verification.

rbac/internal/views.py
```diff
@@ -2624,3 +2626,70 @@ def recompute_tenant_role_bindings(request, org_id):
...
+            "restore_timestamp": "2026-05-28T10:00:00Z",
+            "buffer_minutes": 5
+        }
+
+    Returns 202 with task_id on success.
+    """
+    if not getattr(settings, "DR_RECOVERY_ENABLED", False):
+        return JsonResponse({"detail": "DR recovery is disabled (DR_RECOVERY_ENABLED=False)"}, status=403)
+
+    try:
```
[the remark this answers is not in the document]
dsova06: Not a bug -- Python 3.11+ handles the Z suffix natively in datetime.fromisoformat, and this project targets Python 3.12. The tests confirm this works correctly.

rbac/core/kafka_dr.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+def _create_consumer(group_id: str = DR_CONSUMER_GROUP) -> KafkaConsumer:
+    """Create a KafkaConsumer with appropriate auth settings."""
+    kwargs: dict[str, Any] = {
+        "group_id": group_id,
+        "enable_auto_commit": False,
+        "auto_offset_reset": "earliest",
+        "consumer_timeout_ms": 5000,
+        "value_deserializer": lambda m: __import__("json").loads(m.decode("utf-8")),
```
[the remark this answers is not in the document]
dsova06: Fixed: imported json at the top of the module and referenced it directly in the lambda.

rbac/management/workspace/dr_recovery.py
```diff
@@ -0,0 +1,266 @@
...
+    outbox = Outbox(
+        aggregatetype=AggregateTypes.WORKSPACE.value,
+        aggregateid=str(PartitionKey.byEnvironment()),
+        event_type=event_type,
+        payload=payload,
+    )
+    outbox_log.log(outbox)
+
+
+def generate_corrective_workspace_events(
```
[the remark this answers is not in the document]
dsova06: Keeping the current approach. The TypedDict works well here because the stats dict is passed directly to Celery task results and logging -- dict(stats) is a natural conversion. The dict-overwrite dedup is a standard Python pattern and the docstring already documents the behavior. The two-entry operation mapping is small enough that either approach is fine.

dsova06:
@sourcery-ai resolve

dsova06:
Addressed review feedback:

1. **Partitions without messages in the window** -- Fixed. Partitions where `offsets_for_times` returns `None` are now paused so the consumer skips them entirely instead of reading from offset 0.

2. **Consumer reads past the window end** -- Fixed. Added a second `offsets_for_times` call for the end timestamp. The consumer now tracks per-partition end offsets and pauses partitions once they reach it. When all active partitions are done, the loop breaks immediately.

3. **`__import__("json")` cleanup** -- Fixed. Imported `json` at the top of the module.

4. **Concurrent execution guard** -- Noted. Corrective events are idempotent from the downstream perspective (duplicate create/delete/update for an already-synced workspace is a no-op), so this is acceptable for a rarely-used DR tool.

rbac/internal/views.py
```diff
@@ -2624,3 +2626,70 @@ def recompute_tenant_role_bindings(request, org_id):
...
+
+    POST /_private/api/disaster_recovery/workspaces/
+
+    Accepts JSON body:
+        {
+            "restore_timestamp": "2026-05-28T10:00:00Z",
+            "buffer_minutes": 5
+        }
+
+    Returns 202 with task_id on success.
```
lpichler, line 2643: **medium**: [ISSUE-1] No concurrency guard against duplicate DR recovery runs

The endpoint has no protection against concurrent or duplicate recovery task submissions. If an operator triggers the endpoint twice, two Celery tasks will run concurrently, both reading the same Kafka window and writing duplicate corrective events to the outbox, causing duplicate corrections to downstream services.

**Suggestion:** Add a distributed lock (e.g., Redis-based) or task-level idempotency check. Use a deterministic task_id derived from restore_timestamp, or check Redis for an in-progress flag before starting the task.

rbac/core/kafka_dr.py
```diff
@@ -0,0 +1,168 @@
...
+                    break
+                continue
+
+            if message.timestamp > end_timestamp_ms:
+                continue
+
+            if message.timestamp < start_timestamp_ms:
+                continue
+
+            try:
```
lpichler, line 150: **medium**: [ISSUE-2] Kafka consumer may iterate unnecessarily beyond the end timestamp on unbounded partitions

When offsets_for_times returns None for end_timestamp_ms on a partition, there is no end boundary. The consumer reads and discards all messages beyond the window one by one until the 5-second idle timeout, rather than pausing the partition on first out-of-range message.

**Suggestion:** When message.timestamp > end_timestamp_ms on an unbounded partition, pause and mark the partition as finished instead of just continuing.

rbac/management/workspace/dr_recovery.py
```diff
@@ -0,0 +1,266 @@
...
+    Args:
+        kafka_events: Raw Kafka events from the workspace topic within the lost window.
+        outbox_log: Optional OutboxLog implementation (defaults to OutboxWAL for production).
+
+    Returns:
+        CorrectiveEventStats with counts of each action taken.
+    """
+    if outbox_log is None:
+        outbox_log = OutboxWAL()
+
```
lpichler, line 172: **medium**: [ISSUE-3] Corrective events not wrapped in a database transaction

The generate_corrective_workspace_events function writes multiple outbox entries in a loop without an enclosing transaction. A crash mid-loop leaves partial corrections in the WAL, creating inconsistent recovery state. This violates the codebase guideline: 'Do not perform outbox writes outside a transaction boundary.'

**Suggestion:** Wrap the event generation loop in django.db.transaction.atomic() to ensure all corrective events are committed atomically.

rbac/internal/views.py
```diff
@@ -2624,3 +2626,70 @@ def recompute_tenant_role_bindings(request, org_id):
...
+    except (ValueError, TypeError):
+        return JsonResponse({"detail": "restore_timestamp must be a valid ISO 8601 datetime"}, status=400)
+
+    if parsed_ts.tzinfo is None:
+        parsed_ts = parsed_ts.replace(tzinfo=datetime.timezone.utc)
+
+    now = datetime.datetime.now(datetime.timezone.utc)
+    if parsed_ts > now:
+        return JsonResponse({"detail": "restore_timestamp must be in the past"}, status=400)
+
```
lpichler, line 2668: **low**: [ISSUE-4] Boolean values pass the buffer_minutes integer validation

In Python, bool is a subclass of int, so isinstance(True, int) returns True. JSON body with buffer_minutes: true would pass validation and be interpreted as 1 minute.

**Suggestion:** Add isinstance(buffer_minutes, bool) check before the int check: if isinstance(buffer_minutes, bool) or not isinstance(buffer_minutes, int) or buffer_minutes < 0.

rbac/management/workspace/dr_recovery.py
```diff
@@ -0,0 +1,266 @@
...
+    ReplicationEventType,
+)
+from management.workspace.model import Workspace
+from management.workspace.serializer import WorkspaceEventSerializer
+
+logger = logging.getLogger(__name__)
+
+PROCESSABLE_WORKSPACE_TYPES = {Workspace.Types.STANDARD, Workspace.Types.DEFAULT}
+_PROCESSABLE_TYPE_VALUES = {str(t) for t in PROCESSABLE_WORKSPACE_TYPES}
+
```
lpichler, line 52: **low**: [ISSUE-5] _PROCESSABLE_TYPE_VALUES uses str() conversion which is fragile

Using str(t) on TextChoices members rather than t.value is fragile across Django versions where __str__ behavior has changed.

**Suggestion:** Use .value explicitly: _PROCESSABLE_TYPE_VALUES = {t.value for t in PROCESSABLE_WORKSPACE_TYPES}

rbac/internal/views.py
dsova06: Fixed: added Redis-based distributed lock via cache.add() with 1-hour timeout. Lock is released in a finally block to ensure cleanup even on failure.

rbac/core/kafka_dr.py
dsova06: Fixed: partitions without a known end offset (offsets_for_times returned None) are now paused and marked finished on the first message that exceeds end_timestamp_ms.

rbac/management/workspace/dr_recovery.py
dsova06: Fixed: wrapped the corrective event generation loop in transaction.atomic() so all outbox writes are committed atomically.

rbac/internal/views.py
dsova06: Fixed: added isinstance(buffer_minutes, bool) guard before the int check, since bool is a subclass of int in Python.

rbac/management/workspace/dr_recovery.py
dsova06: Fixed: reverted to .value with a type: ignore[attr-defined] comment for mypy. The str() approach was fragile across Django versions as you noted.

rbac/core/kafka_dr.py
```diff
@@ -0,0 +1,189 @@
...
+        "value_deserializer": lambda m: json.loads(m.decode("utf-8")),
+    }
+
+    if settings.KAFKA_AUTH:
+        kwargs.update(settings.KAFKA_AUTH)
+    elif settings.KAFKA_SERVERS:
+        kwargs["bootstrap_servers"] = settings.KAFKA_SERVERS
+    else:
+        raise RuntimeError("Kafka is not configured (no KAFKA_AUTH or KAFKA_SERVERS)")
+
```
lpichler, line 62: **low**: KAFKA_AUTH AttributeError when Kafka is disabled

settings.KAFKA_AUTH is only defined when KAFKA_ENABLED=True. If DR_RECOVERY_ENABLED=True but KAFKA_ENABLED=False, accessing settings.KAFKA_AUTH raises AttributeError.

rbac/management/workspace/dr_recovery.py
```diff
@@ -0,0 +1,268 @@
+#
```
lpichler, line 1: **medium**: Exception inside transaction.atomic() may silently roll back all corrective events

All outbox writes are in a single transaction.atomic() block with per-event try/except. If a DB-level error occurs (e.g., IntegrityError from save), PostgreSQL marks the transaction as aborted. All subsequent writes fail silently, and the entire transaction rolls back on exit, losing all corrective events while stats suggest success.

rbac/core/kafka_dr.py
```diff
@@ -0,0 +1,189 @@
...
+        "enable_auto_commit": False,
+        "auto_offset_reset": "earliest",
+        "consumer_timeout_ms": 5000,
+        "value_deserializer": lambda m: json.loads(m.decode("utf-8")),
+    }
+
+    if settings.KAFKA_AUTH:
+        kwargs.update(settings.KAFKA_AUTH)
+    elif settings.KAFKA_SERVERS:
+        kwargs["bootstrap_servers"] = settings.KAFKA_SERVERS
```
lpichler, line 59: **low**: consumer_timeout_ms=5000 may be too aggressive for DR scenarios

5-second consumer timeout may not be sufficient when Kafka is under load during DR. The for-message iterator exits when no messages arrive within this window.

rbac/core/kafka_dr.py
```diff
@@ -0,0 +1,189 @@
...
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+"""Kafka utilities for disaster recovery -- read events by timestamp range."""
+
+import json
+import logging
+from dataclasses import dataclass
+from typing import Any, Callable, Optional
+
```
lpichler, line 24: **cosmetic**: typing.Optional import deprecated for Python 3.12

Uses typing.Optional instead of X | None syntax preferred in Python 3.12+.

rbac/core/kafka_dr.py
dsova06: Fixed: settings.KAFKA_AUTH is now accessed via getattr() with a None default, so it won't raise AttributeError when KAFKA_ENABLED=False. Same for KAFKA_SERVERS.

rbac/management/workspace/dr_recovery.py
dsova06: Fixed: moved transaction.atomic() inside the per-event loop so each corrective event gets its own savepoint. A DB-level error on one event no longer aborts the entire batch -- subsequent events continue to commit independently.

rbac/core/kafka_dr.py
dsova06: Fixed: consumer_timeout_ms is now configurable via DR_KAFKA_CONSUMER_TIMEOUT_MS setting (default 30000ms, up from 5000ms). Operators can tune this for high-load DR scenarios.

rbac/core/kafka_dr.py
dsova06: Fixed: replaced typing.Optional and typing.Callable with X | None and collections.abc.Callable throughout both files.

rbac/internal/views.py
dsova06: Already implemented: the task uses a Redis-based distributed lock (cache.add with 1-hour timeout, line 582). Duplicate submissions return early with a message.

rbac/core/kafka_dr.py
dsova06: Already fixed: when message.timestamp > end_timestamp_ms on an unbounded partition, the partition is paused and marked finished (lines 139-145).

rbac/management/workspace/dr_recovery.py
dsova06: Already fixed: each corrective event write is wrapped in its own transaction.atomic() (per-event scope, not outer wrapper). A DB error on one event only rolls back that event.

rbac/internal/views.py
dsova06: Already fixed: isinstance(buffer_minutes, bool) check is present at line 2670.

rbac/management/workspace/dr_recovery.py
dsova06: Already fixed: uses t.value explicitly (line 51).

rbac/core/kafka_dr.py
dsova06: Already fixed: uses getattr(settings, 'KAFKA_AUTH', None) with None default (line 55).

rbac/management/workspace/dr_recovery.py
dsova06: Already fixed: per-event transaction.atomic() scope with try/except outside the atomic block (lines 197-228). A DB error rolls back only that event's transaction.

rbac/core/kafka_dr.py
dsova06: Already fixed: consumer_timeout_ms is configurable via DR_KAFKA_CONSUMER_TIMEOUT_MS setting, defaults to 30000ms (line 51).

rbac/core/kafka_dr.py
dsova06: The current code uses collections.abc.Callable and typing.Any. Optional was removed in favor of X | None syntax where applicable.

dsova06:
Fixed: added dry_run mode to the workspace recovery endpoint and task. Pass `"dry_run": true` in the request body. In dry-run, events are read from Kafka, parsed, and compared against DB state, but no corrective events are written to the outbox. Stats (creates, deletes, updates, skipped) are still reported so you can see exactly what would happen. Each planned action is logged with DRY RUN prefix.

lpichler (approved):

PR #2955, opened 2026-05-28
[RHCLOUD-42232] Migrating code from UMB to Kafka

shannon-donahue:
## Link(s) to Jira
- [RHCLOUD-42232](https://redhat.atlassian.net/browse/RHCLOUD-42232)
-  [HCMSEC-3031](https://redhat.atlassian.net/browse/HCMSEC-3031)

Principal Cleanup Refactor:
  - Refactored rbac/management/principal/cleaner.py to use Kafka for principal cleanup events instead of UMB depending on unleash flag
  - Integrated with existing Kafka consumer infrastructure (rbac/core/kafka_consumer.py)
  -  Toggle between UMB and Kafka in real-time via Unleash:
    - Set Unleash flag rbac.principal-cleanup.use-kafka.enabled to True → Uses Kafka
    - Set Unleash flag rbac.principal-cleanup.use-kafka.enabled to False → Uses UMB

  Configuration Updates:
  - Updated .env.example with new Kafka-specific environment variables
  - Modified rbac/rbac/settings.py to add Kafka consumer configuration settings
  - Updated deploy/rbac-clowdapp.yml with Kafka consumer deployment configuration
  - Enhanced docker-compose.yml to include Kafka and Zookeeper services for local development

  Celery Task Updates:
  - Updated rbac/rbac/celery.py and rbac/management/tasks.py to use Kafka-based principal cleanup
  - Removed UMB-specific Celery beat schedules

  Testing Infrastructure:
  - Converted test suite in tests/management/principal/test_cleaner.py  from UMB to KAFKA
  - Removed UMB-specific test mocking and fixtures
  - Added scripts/test_kafka_producer.py for manual integration testing

## Local Testing
 All existing tests pass with the new Kafka implementation:

#### Run all tests (should pass)
  `pipenv run tox -e py312-fast`

  #### Run only principal cleanup tests
  `pipenv run tox -e py312-fast -- tests.management.principal.test_cleaner`

  #### Run with coverage
  `pipenv run tox -e py312`

 Expected Results: All tests pass.

#### 1. Start services with Debezium
  `podman-compose -f docker-compose.debezium.yml up -d`

  #### 2. Run consumer
  `./scripts/debezium/run_kafka_consumer.sh`

  ##### 3. Send test relations message
  `./scripts/debezium/send_test_relations_message.sh`

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

[RHCLOUD-42232]: https://redhat.atlassian.net/browse/RHCLOUD-42232?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HCMSEC-3031]: https://redhat.atlassian.net/browse/HCMSEC-3031?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Migrate principal cleanup processing from UMB to Kafka and wire it into the scheduled cleanup job and configuration.

New Features:
- Introduce Kafka-based principal cleanup flow that consumes canonical user events and updates or deletes principals and tenants accordingly.
- Add local Kafka and Zookeeper services plus a helper Kafka producer script for manual/integration testing of principal cleanup messages.

Enhancements:
- Replace UMB/STOMP XML handling with Kafka JSON processing in the principal cleaner, including new success/failure metrics and lock-based concurrency control.
- Update Celery scheduling and settings to drive principal cleanup via Kafka feature flags and environment variables instead of UMB-specific ones.
- Expand and refactor principal cleanup tests to exercise Kafka-based behaviour, tenant bootstrap interactions, and failure handling paths.
- Adjust Docker and Debezium-related scripts and health checks to support running the Kafka consumer with metrics in local and test environments.
- Change dependency installation to skip Pipfile.lock resolution during image build to work around lock conflicts.

Build:
- Updated UMB-related Python dependencies (stompest, xmltodict) and rely on kafka-python for messaging.
- Update Dockerfile to install dependencies with pipenv using --skip-lock to avoid lock-induced build failures.

Deployment:
- Update ClowdApp deployment configuration and parameters to remove UMB host/secret usage and add Kafka-based principal cleanup configuration flags.

Tests:
- Rewrite principal cleanup integration tests to use Kafka consumers and JSON messages instead of UMB frames, covering multiple tenant scenarios, disable/update paths, and error handling.
- Add a standalone Kafka producer test script to send sample canonical user messages for exercising the cleanup pipeline end-to-end.

## Summary by CodeRabbit

* **New Features**
  * Added Kafka-based principal cleanup with configurable 3-mode behavior: UMB-only, Kafka shadow (dry-run), and Kafka active (writes).
  * Introduced new Kafka cleanup controls, topic/DLQ routing, and message-consumer health/enablement, plus automated scheduling/dispatch.
  * Updated local development to run Kafka with ZooKeeper via docker-compose (including health checks and consumer metrics).

* **Bug Fixes**
  * Improved principal reference extraction to capture both organization and account details during cleanup.

* **Tests**
  * Expanded end-to-end Kafka cleaner test coverage, including dry-run and DLQ behaviors; updated UMB failure-path mocks.

* **Chores**
  * Added Kafka local-testing defaults to `.env.example` and included a small Kafka producer test script.

### DIFF
--- a/deploy/rbac-clowdapp.yml
+++ b/deploy/rbac-clowdapp.yml
@@ -270,6 +270,12 @@ objects:
             value: ${EPH_ENV}
           - name: PRINCIPAL_CLEANUP_UPDATE_ENABLED_UMB
             value: ${PRINCIPAL_CLEANUP_UPDATE_ENABLED_UMB}
+          - name: PRINCIPAL_CLEANUP_UPDATE_ENABLED_KAFKA
+            value: ${PRINCIPAL_CLEANUP_UPDATE_ENABLED_KAFKA}
+          - name: KAFKA_PRINCIPAL_CLEANUP_TOPIC
+            value: ${KAFKA_PRINCIPAL_CLEANUP_TOPIC}
+          - name: KAFKA_PRINCIPAL_CLEANUP_DLQ_TOPIC
+            value: ${KAFKA_PRINCIPAL_CLEANUP_DLQ_TOPIC}
           - name: V2_MIGRATION_APP_EXCLUDE_LIST
             value: ${V2_MIGRATION_APP_EXCLUDE_LIST}
           - name: V2_STRICT_ACCESS_CHECK_FLAG_APPLICATION_NAMES
@@ -319,6 +325,10 @@ objects:
             value: ${PARITY_CHECK_ORG_IDS}
           - name: PARITY_CHECK_SCHEDULE
             value: ${PARITY_CHECK_SCHEDULE}
+          - name: PRINCIPAL_CLEANUP_DELETION_ENABLED_KAFKA
+            value: ${PRINCIPAL_CLEANUP_DELETION_ENABLED_KAFKA}
+          - name: KAFKA_PRINCIPAL_CLEANUP_JOB_ENABLED
+            value: ${KAFKA_PRINCIPAL_CLEANUP_JOB_ENABLED}
           - name: READ_ONLY_API_MODE
             value: ${READ_ONLY_API_MODE}
           - name: PARITY_CHECK_ENABLED
@@ -616,6 +626,8 @@ objects:
             value: ${EPH_ENV}
           - name: PRINCIPAL_CLEANUP_UPDATE_ENABLED_UMB
             value: ${PRINCIPAL_CLEANUP_UPDATE_ENABLED_UMB}
+          - name: PRINCIPAL_CLEANUP_UPDATE_ENABLED_KAFKA
+            value: ${PRINCIPAL_CLEANUP_UPDATE_ENABLED_KAFKA}
           - name: V2_MIGRATION_APP_EXCLUDE_LIST
             value: ${V2_MIGRATION_APP_EXCLUDE_LIST}
           - name: V2_STRICT_ACCESS_CHECK_FLAG_APPLICATION_NAMES
@@ -851,6 +863,10 @@ objects:
             value: ${RBAC_KAFKA_CONSUMER_GROUP_ID}
           - name: RBAC_KAFKA_CUSTOM_CONSUMER_BROKER
             value: ${RBAC_KAFKA_CUSTOM_CONSUMER_BROKER}
+          - name: KAFKA_PRINCIPAL_CLEANUP_TOPIC
+            value: ${KAFKA_PRINCIPAL_CLEANUP_TOPIC}
+          - name: KAFKA_PRINCIPAL_CLEANUP_DLQ_TOPIC
+            value: ${KAFKA_PRINCIPAL_CLEANUP_DLQ_TOPIC}
           - name: REPLICATION_TO_RELATION_ENABLED
             value: ${REPLICATION_TO_RELATION_ENABLED}
           - name: EPH_ENV
@@ -1101,6 +1117,8 @@ objects:
               value: ${SA_NAME}
             - name: PRINCIPAL_CLEANUP_UPDATE_ENABLED_UMB
               value: ${PRINCIPAL_CLEANUP_UPDATE_ENABLED_UMB}
+            - name: PRINCIPAL_CLEANUP_UPDATE_ENABLED_KAFKA
+              value: ${PRINCIPAL_CLEANUP_UPDATE_ENABLED_KAFKA}
             - name: V2_MIGRATION_APP_EXCLUDE_LIST
               value: ${V2_MIGRATION_APP_EXCLUDE_LIST}
             - name: V2_STRICT_ACCESS_CHECK_FLAG_APPLICATION_NAMES
@@ -1324,6 +1342,8 @@ objects:
               value: ${SA_NAME}
             - name: PRINCIPAL_CLEANUP_UPDATE_ENABLED_UMB
               value: ${PRINCIPAL_CLEANUP_UPDATE_ENABLED_UMB}
+            - name: PRINCIPAL_CLEANUP_UPDATE_ENABLED_KAFKA
+              value: ${PRINCIPAL_CLEANUP_UPDATE_ENABLED_KAFKA}
             - name: V2_MIGRATION_APP_EXCLUDE_LIST
               value: ${V2_MIGRATION_APP_EXCLUDE_LIST}
             - name: V2_STRICT_ACCESS_CHECK_FLAG_APPLICATION_NAMES
@@ -1881,12 +1901,18 @@ parameters:
   value: 'redhat'
 - name: PRINCIPAL_CLEANUP_DELETION_ENABLED_UMB
   description: Allow cleanup job to delete principals via messages from UMB
+- name: PRINCIPAL_CLEANUP_DELETION_ENABLED_KAFKA
+  description: Allow cleanup job to delete principals via messages from Kafka
   value: 'False'
 - name: PRINCIPAL_CLEANUP_UPDATE_ENABLED_UMB
   description: Allow cleanup job to update principals via messages from UMB
+- name: PRINCIPAL_CLEANUP_UPDATE_ENABLED_KAFKA
+  description: Allow cleanup job to update principals via messages from Kafka
   value: 'False'
 - name: UMB_JOB_ENABLED
   description: Temp env to enable the UMB job
+- name: KAFKA_PRINCIPAL_CLEANUP_JOB_ENABLED
+  description: Enable the Kafka-based principal cleanup job
   value: 'True'
 - name: UMB_HOST
   description: Host of the UMB service
@@ -2040,6 +2066,12 @@ parameters:
   value: '1'
 - name: RBAC_KAFKA_CUSTOM_CONSUMER_BROKER
   description: Custom Kafka broker URL for the RBAC Kafka consumer (if empty, uses default Clowder/localhost configuration)
+  value: ''
+- name: KAFKA_PRINCIPAL_CLEANUP_TOPIC
+  description: Kafka topic name for principal cleanup events. REQUIRED when PRINCIPAL_CLEANUP_DELETION_ENABLED_KAFKA=true (deployment will fail if not set).
+  value: ''
+- name: KAFKA_PRINCIPAL_CLEANUP_DLQ_TOPIC
+  description: Kafka Dead Letter Queue topic for failed principal cleanup messages (optional, prevents infinite retry loops)
   value: ''
 - name: ROOT_SCOPE_PERMISSIONS
   description: Comma-separated list of permissions that bind to root workspace scope (supports wildcards like rbac:*:read)
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -39,6 +39,11 @@ services:
         - RBAC_DESTRUCTIVE_SEEDING_ENABLED_UNTIL=${RBAC_DESTRUCTIVE_SEEDING_ENABLED_UNTIL}
         - KAFKA_ENABLED=false
         - RBAC_KAFKA_CUSTOM_CONSUMER_BROKER=kafka:9092
+        - PRINCIPAL_CLEANUP_DELETION_ENABLED_KAFKA=False
+        - PRINCIPAL_CLEANUP_UPDATE_ENABLED_KAFKA=False
+        - KAFKA_PRINCIPAL_CLEANUP_JOB_ENABLED=True
+        - KAFKA_PRINCIPAL_CLEANUP_TOPIC=rbac.principal-cleanup
+        - KAFKA_PRINCIPAL_CLEANUP_DLQ_TOPIC=rbac.principal-cleanup.dlq
         - RBAC_KAFKA_CONSUMER_TOPIC=outbox.event.relations-replication-event
         - RBAC_KAFKA_CONSUMER_GROUP_ID=rbac-consumer-group
         - REPLICATION_TO_RELATION_ENABLED=true
@@ -64,6 +69,8 @@ services:
         rbac-worker:
           condition: service_healthy
         rbac-scheduler:
+          condition: service_healthy
+        kafka:
           condition: service_healthy
       healthcheck:
         test: ["CMD-SHELL", "python3 -c \"import urllib.request; urllib.request.urlopen('http://localhost:8080/metrics')\""]
@@ -134,6 +141,34 @@ services:
       timeout: 5s
       retries: 3
       start_period: 5s
+
+  zookeeper:
+    container_name: rbac_zookeeper
+    image: confluentinc/cp-zookeeper:7.5.0
+    environment:
+      ZOOKEEPER_CLIENT_PORT: 2181
+      ZOOKEEPER_TICK_TIME: 2000
+    ports:
+      - "2181:2181"
+
+  kafka:
+    container_name: rbac_kafka
+    image: confluentinc/cp-kafka:7.5.0
+    depends_on:
+      - zookeeper
+    ports:
+      - "9092:9092"
+    environment:
+      KAFKA_BROKER_ID: 1
+      KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
+      KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://kafka:9092
+      KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
+      KAFKA_AUTO_CREATE_TOPICS_ENABLE: "true"
+    healthcheck:
+      test: ["CMD-SHELL", "kafka-topics --bootstrap-server localhost:9092 --list"]
+      interval: 10s
+      timeout: 10s
+      retries: 5
 
   wait_for_app:
     container_name: wait_for_app
--- a/rbac/feature_flags.py
+++ b/rbac/feature_flags.py
@@ -43,6 +43,8 @@ class FeatureFlags:
     TOGGLE_USE_ROLE_BINDING_VIEW_PERMISSION = "rbac.use-role-binding-view-permission.enabled"
     # Per-org flag: when enabled, the org uses v2 APIs for write operations and v1 write APIs are blocked.
     TOGGLE_V2_EDIT_API_ENABLED = "platform.rbac.workspaces"
+    # When enabled, use Kafka for principal cleanup; when disabled, use UMB.
+    TOGGLE_USE_KAFKA_CLEANUP = "rbac.principal-cleanup.use-kafka.enabled"
     # Per-org flag: when enabled, the org uses only V2 access checks for HBI (and V1 access checks are blocked).
     TOGGLE_V2_ADDITIONAL_MANDATORY_ACCESS_CHECK_REQUIRED = "hbi.rbac-v2"
 
@@ -183,6 +185,87 @@ class FeatureFlags:
             context={"orgId": str(org_id)},
             fallback_function=lambda ignored_toggle_name, ignored_context: settings.V2_EDIT_API_ENABLED,
         )
+
+    def is_kafka_principal_cleanup_enabled(self):
+        """Check whether to use Kafka for principal cleanup.
+
+        DEPRECATED: Use get_principal_cleanup_mode() instead for 3-state control.
+
+        When enabled (True), use Kafka for principal cleanup.
+        When disabled (False), use UMB for principal cleanup.
+        Falls back to False (UMB) if Unleash is unavailable - UMB is the proven original method.
+        """
+        return self.is_enabled(
+            feature_name=self.TOGGLE_USE_KAFKA_CLEANUP,
+            fallback_function=lambda ignored_toggle_name, ignored_context: False,
+        )
+
+    def get_principal_cleanup_mode(self) -> str:
+        """
+        Get the principal cleanup mode using the feature flag.
+
+        This method supports 4 modes controlled by the Unleash flag value:
+        - 'umb_only' (flag disabled OR flag enabled with missing/unrecognized variant):
+          Only UMB consumer runs and writes to DB
+        - 'kafka_shadow' (flag enabled with kafka_shadow variant): Both UMB and Kafka run, only UMB writes
+          (Kafka dry-run)
+        - 'kafka_validation' (flag enabled with kafka_validation variant): Kafka tries to write first,
+          UMB writes only if Kafka fails (Phase 2 validation)
+        - 'kafka_active' (flag enabled with kafka_active variant): Only Kafka consumer runs and writes to DB
+
+        IMPORTANT: Enabling the flag WITHOUT setting a valid variant will keep you on UMB (umb_only mode).
+        This is a safety feature to prevent accidental Kafka activation. You must explicitly set the variant
+        to kafka_shadow, kafka_validation, or kafka_active to use Kafka.
+
+        Returns:
+            str: One of 'umb_only', 'kafka_shadow', 'kafka_validation', or 'kafka_active'
+        """
+        if self.client is None:
+            self.initialize()
+
+        if self.client is None:
+            logger.warning("FeatureFlags not initialized, defaulting to umb_only mode")
+            return "umb_only"
+
+        # First check if the toggle is enabled
+        is_enabled = self.is_enabled(
+            feature_name=self.TOGGLE_USE_KAFKA_CLEANUP,
+            fallback_function=lambda ignored_toggle_name, ignored_context: False,
+        )
+
+        if not is_enabled:
+            # Flag is disabled -> UMB only mode
+            return "umb_only"
+
+        # Flag is enabled -> check for variant to determine mode
+        try:
+            variant = self.client.get_variant(
+                self.TOGGLE_USE_KAFKA_CLEANUP,
+                fallback_variant={"name": "umb_only", "enabled": False},
+            )
+
+            mode = variant.get("name", "umb_only")
+
+            # Validate mode - only these variants are valid when flag is enabled
+            if mode in ["kafka_shadow", "kafka_validation", "kafka_active"]:
+                return mode
+            else:
+                # Unknown/missing variant when flag is enabled -> safe default is UMB
+                # This prevents accidental Kafka activation and requires explicit variant selection
+                logger.warning(
+                    f"Flag 'rbac.principal-cleanup.use-kafka.enabled' is ON but variant '{mode}' is not recognized. "
+                    f"Staying on UMB (umb_only mode) for safety. "
+                    f"To activate Kafka, set variant to: kafka_shadow, kafka_validation, or kafka_active"
+                )
+                return "umb_only"
+
+        except Exception as e:
+            logger.warning(f"Error getting variant for principal cleanup mode: {e}, falling back to umb_only")
+            return "umb_only"
+
+    def is_kafka_shadow_mode_enabled(self) -> bool:
+        """Check if Kafka is in shadow/dry-run mode."""
+        return self.get_principal_cleanup_mode() == "kafka_shadow"
 
     def is_v2_strict_access_check_enabled(self, org_id: str) -> bool:
         """Check whether strict V2 access checks are required in the given org.
--- a/rbac/management/principal/cleaner.py
+++ b/rbac/management/principal/cleaner.py
@@ -17,14 +17,20 @@
 
 """Handler for principal clean up."""
 
+import base64
+import json
 import logging
 import os
 import ssl
-from typing import Optional
+from typing import NamedTuple, Optional
+from xml.parsers.expat import ExpatError
 
 import xmltodict
+from core.kafka import RBACProducer
 from django.conf import settings
 from django.db import connection, transaction
+from kafka import KafkaConsumer
+from kafka.errors import KafkaError
 from management.principal.model import Principal
 from management.principal.proxy import PrincipalProxy, external_principal_to_user
 from management.relation_replicator.outbox_replicator import OutboxReplicator
@@ -50,8 +56,10 @@ CA_LOC = "/opt/rbac/rbac/management/principal/umb_certificates/ca.crt"
 CERT_LOC = "/opt/rbac/rbac/management/principal/umb_certificates/tls.crt"
 KEY_LOC = "/opt/rbac/rbac/management/principal/umb_certificates/tls.key"
 
+
 LOCK_ID = 42  # For Keith, with Love
 
+# UMB Metric Messages
 METRIC_STOMP_MESSAGES_ACK_TOTAL = "stomp_messages_ack_total"
 METRIC_STOMP_MESSAGES_NACK_TOTAL = "stomp_messages_nack_total"
 stomp_messages_ack_total = Counter(
@@ -61,6 +69,42 @@ stomp_messages_ack_total = Counter(
 stomp_messages_nack_total = Counter(
     METRIC_STOMP_MESSAGES_NACK_TOTAL,
     "Number of stomp UMB messages that failed to be processed",
+)
+
+# KAFKA Metric Messages
+METRIC_KAFKA_MESSAGES_SUCCESS_TOTAL = "kafka_messages_success_total"
+METRIC_KAFKA_MESSAGES_FAILURE_TOTAL = "kafka_messages_failure_total"
+kafka_messages_success_total = Counter(
+    METRIC_KAFKA_MESSAGES_SUCCESS_TOTAL,
+    "Number of Kafka messages processed successfully",
+)
+kafka_messages_failure_total = Counter(
+    METRIC_KAFKA_MESSAGES_FAILURE_TOTAL,
+    "Number of Kafka messages that failed to be processed",
+)
+
+# KAFKA Shadow Mode Metrics
+METRIC_KAFKA_DRY_RUN_MESSAGES_TOTAL = "kafka_dry_run_messages_total"
+METRIC_KAFKA_DRY_RUN_ERRORS_TOTAL = "kafka_dry_run_errors_total"
+kafka_dry_run_messages_total = Counter(
+    METRIC_KAFKA_DRY_RUN_MESSAGES_TOTAL,
+    "Number of Kafka messages processed in dry-run/shadow mode",
+)
+kafka_dry_run_errors_total = Counter(
+    METRIC_KAFKA_DRY_RUN_ERRORS_TOTAL,
+    "Number of Kafka messages that would have failed if not in dry-run mode",
+)
+
+# KAFKA Validation Mode Metrics
+METRIC_KAFKA_VALIDATION_SUCCESS_TOTAL = "kafka_validation_success_total"
+METRIC_KAFKA_VALIDATION_FALLBACK_TOTAL = "kafka_validation_fallback_total"
+kafka_validation_success_total = Counter(
+    METRIC_KAFKA_VALIDATION_SUCCESS_TOTAL,
+    "Number of times Kafka succeeded in validation mode (no UMB fallback needed)",
+)
+kafka_validation_fallback_total = Counter(
+    METRIC_KAFKA_VALIDATION_FALLBACK_TOTAL,
+    "Number of times UMB fallback was triggered in validation mode",
 )
 
 
@@ -148,7 +192,7 @@ QUEUE = f"/queue/Consumer.{settings.SA_NAME}.users-subscription.VirtualTopic.can
 UMB_CLIENT = Stomp(CONFIG)
 
 
-def retrieve_user_info(message) -> User:
+def retrieve_user_info_umb(message) -> User:
     """
     Retrieve user info from the message.
 
@@ -161,7 +205,7 @@ def retrieve_user_info(message) -> User:
         if (id := header.get("InstanceId")) is not None:
             instance_id = id
 
-    logger.debug("retrieve_user_info: Processing message with instance_id=%s", instance_id)
+    logger.debug("retrieve_user_info_UMB: Processing message with instance_id=%s", instance_id)
 
     message_user = message["Payload"]["Sync"]["User"]
     identifiers = message_user["Identifiers"]
@@ -189,12 +233,96 @@ def retrieve_user_info(message) -> User:
         # identifiers["Reference"] might be a dict
         if not isinstance((refs := identifiers["Reference"]), list):
             refs = [identifiers["Reference"]]
+        # Preserve original UMB behavior: use first matching reference and stop
+        # This maintains production behavior where messages typically have one of each reference type
+        # If multiple WEB/Customer or EBS/Account references exist, use the first one
         for ref in refs:
             if ref["@system"] == "WEB" and ref["@entity-name"] == "Customer" and ref["@qualifier"] == "id":
                 user.org_id = ref["#text"]
                 break
             if ref["@system"] == "EBS" and ref["@entity-name"] == "Account" and ref["@qualifier"] == "number":
                 user.account = ref["#text"]
+                break
+
+        return user
+
+    user_data = bop_resp["data"][0]
+    return external_principal_to_user(user_data)
+
+
+def retrieve_user_info_kafka(message) -> User:
+    """
+    Retrieve user info from the Kafka message.
+
+    Args:
+        message: JSON message from Kafka containing user event data
+
+    returns:
+        user: User object as of latest known state.
+    """
+    instance_id: Optional[str] = None
+
+    # Extract instance ID from header if present
+    if (header := message.get("Header")) is not None:
+        if (id := header.get("InstanceId")) is not None:
+            instance_id = id
+
+    logger.debug("retrieve_user_info_kafka: Processing message with instance_id=%s", instance_id)
+
+    # Navigate through JSON structure (similar to XML but without @ and # prefixes)
+    message_user = message["Payload"]["Sync"]["User"]
+    identifiers = message_user["Identifiers"]
+    user_id: Optional[str] = None
+
+    # Handle both list and single identifier cases
+    identifier_list = identifiers.get("Identifier", [])
+    if not isinstance(identifier_list, list):
+        identifier_list = [identifier_list]
+
+    # Find the user ID from identifiers
+    for identifier in identifier_list:
+        is_web_user_id = (
+            identifier.get("system") == "WEB"
+            and identifier.get("entity-name") == "User"
+            and identifier.get("qualifier") == "id"
+        )
+        if is_web_user_id:
+            user_id = identifier.get("text") or identifier.get("value")
+            break
+
+    if user_id is None:
+        raise ValueError(f"User id not found in message. instance_id={instance_id}")
+
+    # Query BOP for user information
+    bop_resp = PROXY.request_filtered_principals([user_id], options={"query_by": "user_id", "return_id": True})
+
+    if not bop_resp["data"]:  # User has been deleted
+        # Get data from message instead
+        user = User()
+        user.user_id = user_id
+        user.is_active = False
+        user.username = message_user["Person"]["Credentials"]["Login"]
+
+        # Handle references (might be a dict or list)
+        references = identifiers.get("Reference", [])
+        if not isinstance(references, list):
+            references = [references]
+
+        # Match UMB behavior: use first matching reference and stop (preserve production semantics)
+        # Kafka receives the same messages as UMB (just different transport), so behavior should be identical
+        # Messages typically have one WEB/Customer reference (org_id) and one EBS/Account reference (account)
+        for ref in references:
+            is_web_customer = (
+                ref.get("system") == "WEB" and ref.get("entity-name") == "Customer" and ref.get("qualifier") == "id"
+            )
+            is_ebs_account = (
+                ref.get("system") == "EBS" and ref.get("entity-name") == "Account" and ref.get("qualifier") == "number"
+            )
+            if is_web_customer:
+                user.org_id = ref.get("text") or ref.get("value")
+                break
+            if is_ebs_account:
+                user.account = ref.get("text") or ref.get("value")
                 break
 
         return user
@@ -221,7 +349,7 @@ def process_umb_event(frame, umb_client: Stomp, bootstrap_service: TenantBootstr
             data_dict = xmltodict.parse(body)
             canonical_message = data_dict.get("CanonicalMessage")
 
-            user = retrieve_user_info(canonical_message)
+            user = retrieve_user_info_umb(canonical_message)
             # By default, only process disabled users.
             # If the setting is enabled, process all users.
             if not user.is_active or settings.PRINCIPAL_CLEANUP_UPDATE_ENABLED_UMB:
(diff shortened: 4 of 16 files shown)


rbac/management/principal/cleaner.py
```diff
@@ -17,14 +17,14 @@
```
lpichler, line 17: **medium**: [CR-001] Behavioral change in reference extraction (break removed)

The original UMB code had break statements in the reference loop, setting only org_id OR account. The new code sets both. This is likely a bug fix but is an undocumented behavioral change.

**Suggestion:** Add a comment noting this is an intentional fix from the original UMB behavior.

lpichler, line 17: **medium**: [CR-003] Hardcoded Kafka topic name

Topic 'VirtualTopic.canonical.user' is hardcoded. The existing Kafka infrastructure uses configurable topic names from settings and ClowdApp topic configuration.

**Suggestion:** Add a setting like KAFKA_PRINCIPAL_CLEANUP_TOPIC and use it instead of the hardcoded value.

lpichler, line 17: **medium**: [CR-004] Offset commit semantics do not provide at-least-once guarantees for failed messages

consumer.commit() without arguments commits the latest consumed offset, which implicitly commits failed message offsets when a later message succeeds. This contradicts the at-least-once semantics described in comments.

**Suggestion:** Either break on failure (stop and retry from failed offset) or use partition-specific offset commits.

lpichler, line 17: **low**: [CR-006] 15-second consumer timeout creates processing gaps

consumer_timeout_ms=15000 combined with 60s Celery beat creates a 45-second gap where messages are not consumed. Matches UMB behavior but not ideal for Kafka.

**Suggestion:** Document as intentional. Consider persistent consumer in future.

lpichler:
## Code Review Findings

**File:** `rbac/rbac/celery.py` (line 1)

**low**: [CR-005] Confusing naming: DELETION_ENABLED controls scheduling, not deletion

PRINCIPAL_CLEANUP_DELETION_ENABLED_KAFKA controls whether the Celery beat task is scheduled, not whether deletion occurs. The actual cleanup always processes disabled users regardless of this flag.

**Suggestion:** Consider renaming or adding clarifying comments.

lpichler:
@shannon-donahue @astrozzc should be here way how to switch from umb for kafka by some env variable or feature flag ? (and keep existing funcionality ? )

astrozzc:
> @shannon-donahue @astrozzc should be here way how to switch from umb for kafka by some env variable or feature flag ? (and keep existing funcionality ? )

That's good idea. So we can switch back and forth much faster, if anything goes wrong.
@shannon-donahue Could you update with feature flag? there are some existing examples for that

shannon-donahue:
> > @shannon-donahue @astrozzc should be here way how to switch from umb for kafka by some env variable or feature flag ? (and keep existing funcionality ? )
> 
> That's good idea. So we can switch back and forth much faster, if anything goes wrong. @shannon-donahue Could you update with feature flag? there are some existing examples for that

yeah of course great idea! Let me add back in the UMB functionality and lay out a feature flag for that 👍 I can do another PR once we are sure things are functioning properly to clean up the UMB code

shannon-donahue:
> > @shannon-donahue @astrozzc should be here way how to switch from umb for kafka by some env variable or feature flag ? (and keep existing funcionality ? )
> 
> That's good idea. So we can switch back and forth much faster, if anything goes wrong. @shannon-donahue Could you update with feature flag? there are some existing examples for that

This functionality is now controlled by the following unleash flag
 -  Toggle between UMB and Kafka in real-time via Unleash:
    - Set Unleash flag rbac.principal-cleanup.use-kafka.enabled to True → Uses Kafka
    - Set Unleash flag rbac.principal-cleanup.use-kafka.enabled to False → Uses UMB

rbac/rbac/celery.py
```diff
@@ -56,13 +57,38 @@
     },
 }
 
-if settings.PRINCIPAL_CLEANUP_DELETION_ENABLED_UMB:
+# Determine which principal cleanup method to use
+# Priority: Unleash flag (if both enabled) > UMB > Kafka > BOP fallback
+if settings.PRINCIPAL_CLEANUP_DELETION_ENABLED_UMB and settings.PRINCIPAL_CLEANUP_DELETION_ENABLED_KAFKA:
+    # Both are enabled - use Unleash flag to decide
+    if FEATURE_FLAGS.is_kafka_principal_cleanup_enabled():
```
astrozzc: This runs during the celery start. So the worker has to be restarted to become effective.
Have to have a single principal_cleanup_via_message_bus task that checks the Unleash flag at runtime and dispatches to the appropriate consumer.

rbac/rbac/settings.py
```diff
@@ -484,6 +484,8 @@
 
 RBAC_KAFKA_CUSTOM_CONSUMER_BROKER = ENVIRONMENT.get_value("RBAC_KAFKA_CUSTOM_CONSUMER_BROKER", default="")
 
+KAFKA_PRINCIPAL_CLEANUP_TOPIC = ENVIRONMENT.get_value("KAFKA_PRINCIPAL_CLEANUP_TOPIC", default="")
```
astrozzc, line 533: What happens if this is not set and have empty value? Would it keep throwing errors?
And it should be added into the rbac-clowdapp.yml so that we can configure it through app-interface

rbac/management/principal/cleaner.py
```diff
@@ -245,6 +340,49 @@ def process_umb_event(frame, umb_client: Stomp, bootstrap_service: TenantBootstr
...
+
+            kafka_messages_success_total.inc()
+            return (True, True)  # Continue processing, message succeeded
+        except Exception as e:
+            logger.error("process_kafka_message: Error processing Kafka message: %s", str(e))
+            capture_exception(e)
+            kafka_messages_failure_total.inc()
+            # Continue processing next message, but mark this one as failed
+            # Failed message offset will not be committed, so it can be retried on restart
+            return (True, False)  # Continue processing, message failed
```
astrozzc: If the message is not processable, like it is malformed. It will be kept in infinite loop. The umb has nack method to put into DLQ. Here we would like similar strategy

docker-compose.yml
```diff
@@ -37,8 +37,11 @@ services:
         - PRINCIPAL_PROXY_SERVICE_SSL_VERIFY=${PRINCIPAL_PROXY_SERVICE_SSL_VERIFY-False}
         - RBAC_DESTRUCTIVE_API_ENABLED_UNTIL=${RBAC_DESTRUCTIVE_API_ENABLED_UNTIL}
         - RBAC_DESTRUCTIVE_SEEDING_ENABLED_UNTIL=${RBAC_DESTRUCTIVE_SEEDING_ENABLED_UNTIL}
-        - KAFKA_ENABLED=false
+        - KAFKA_ENABLED=true
```
astrozzc: Is it intentional?

rbac/management/principal/cleaner.py
```diff
@@ -245,6 +340,49 @@ def process_umb_event(frame, umb_client: Stomp, bootstrap_service: TenantBootstr
     return True
 
 
+def process_kafka_message(message, bootstrap_service: TenantBootstrapService) -> tuple[bool, bool]:
```
astrozzc: Is there a better way to describe the result? tuple itself is not that useful, I know it is described in the description, but maybe better to have a defined subject for it.

astrozzc (commented):
Nit: PR description "Removed stompest and xmltodict from Pipfile xxx" have to be updated

Great progress! Thank you!

docker-compose.yml
shannon-donahue: oops left over from testing set back to false.

rbac/management/principal/cleaner.py
shannon-donahue: I changed it to a namedTuple. However I can change this to a dataclass if that is preferred.

rbac/management/principal/cleaner.py
shannon-donahue: I added in a DLQ topic, and configured that so it should follow the UMB pattern. Happy to follow up with more changes if needed.

rbac/rbac/settings.py
shannon-donahue: Added to rbac-clowdapp and added in validation. Though currently UMB is not validated at configuration I can add that if we want with the addition of the unleash flag to make sure both are set

rbac/rbac/celery.py
shannon-donahue: Dispatcher task created to check unleash flag at runtime rather than startup. Task name: principal_cleanup_via_message_bus()

rbac/management/principal/cleaner.py
```diff
@@ -245,6 +353,138 @@ def process_umb_event(frame, umb_client: Stomp, bootstrap_service: TenantBootstr
...
+        dry_run: If True, validate message but don't write to database (shadow mode)
+
+    Returns:
+        MessageProcessingResult with:
+        - should_continue: False if another listener is running (lock contention), True otherwise
+        - success: True if message was processed successfully, False if it failed
+    """
+    with transaction.atomic():
+        # This is locked per transaction to ensure another listener process does not run concurrently.
+        if not _lock_listener():
```
astrozzc, line 408: Consider acquiring the advisory lock once at the process_principal_events_from_kafka level (outside the message loop) using pg_advisory_lock (session-level) instead of per-transaction. The existing UMB code had the same pattern, but Kafka may process messages faster. Not a blocker for initial merge, but worth a follow-up optimization.

rbac/management/principal/cleaner.py
```diff
@@ -245,6 +353,138 @@ def process_umb_event(frame, umb_client: Stomp, bootstrap_service: TenantBootstr
...
+            # Parse JSON message
+            message_data = json.loads(message.value)
+            canonical_message = message_data.get("CanonicalMessage", message_data)
+
+            user = retrieve_user_info_kafka(canonical_message)
+
+            if dry_run:
+                # DRY RUN MODE: Validate message structure and log what would happen
+                logger.info(
+                    "🔍 DRY RUN: Would process user_id=%s org_id=%s is_active=%s",
```
astrozzc: Remove emoji from message

rbac/management/tasks.py
```diff
@@ -26,26 +26,92 @@
...
+
+    if mode == "umb_only":
+        # UMB-only mode: Only UMB processes and writes to DB
+        if settings.UMB_JOB_ENABLED:
+            logger.info("UMB-only mode: processing via UMB")
+            process_principal_events_from_umb()
+        else:
+            logger.warning("UMB mode selected but UMB_JOB_ENABLED is False")
+
+    elif mode == "kafka_shadow":
```
astrozzc, line 124: Consider spawning the Kafka shadow as a separate sub-task so they run in parallel. Not a blocker. Because each lock for 15s, would lock for 30s in total

astrozzc (commented):
Image

Based on the Messaging Bridge user guide, U2K copies msg.Data[:] (raw UMB message body) into msg.Value[:]. UMB canonical user messages are XML. But process_kafka_message() does json.loads(message.value) and retrieve_user_info_kafka() expects JSON keys like "system", "entity-name", "text" (without XML @/# prefixes).

cleaner.py
Lines 395-396
            message_data = json.loads(message.value)
            canonical_message = message_data.get("CanonicalMessage", message_data)
vs UMB version:


cleaner.py
Lines 328-330
            body = frame.body.decode("utf-8", errors="ignore")
            data_dict = xmltodict.parse(body)
            canonical_message = data_dict.get("CanonicalMessage")
If you're relying on the Messaging Bridge to sync messages during the transition, Kafka messages would contain raw XML and json.loads would fail. 

maybe we can try json load, if failed, parse XML? or parse xml, if failed, json load?
Parse XML in the Kafka consumer too — detect format (starts with <?xml or <) and use xmltodict.parse() just like the UMB code does, falling back to json.loads() for when the producer eventually migrates to JSON.

astrozzc:
1. I think it is better to allow dry_run to throw to dlq because
Shadow mode's entire purpose is validation — you want to discover what would fail in production before switching over. If errors are just logged and swallowed, you might miss them in noisy logs.
DLQ gives you a concrete queue of problematic messages you can inspect, count, and alert on (e.g., "DLQ is growing → something's wrong with the Kafka message format").
UMB is still the active writer — so there's zero risk. The dry-run DLQ is purely observational.
It would immediately surface the XML format issue — every message would land in the DLQ with a json.JSONDecodeError, making the problem obvious from day one of shadow mode.

2. For the error handling
There's no distinction between transient errors (BOP timeout, DB connection lost) and permanent errors (malformed message, missing fields). Both get sent to the DLQ equally.

For a proper DLQ pattern, you'd typically:

DLQ: Only for permanently unprocessable messages (parse errors, schema violations)
Retry: For transient failures (network timeouts, temporary DB issues)

rbac/management/principal/cleaner.py
shannon-donahue: Okay perfect I think since we want to get this in and tested we can save this optimization for a next PR. It would require significant refactoring so just to get us in and started makes sense to hold off.

rbac/management/principal/cleaner.py
shannon-donahue: Removed 👍

rbac/management/tasks.py
shannon-donahue: I think this we leave for a separate PR if we need it. Shadow mode is only going to be a temporary add in anyway for testing and should be removed promptly. This may be something we do not need to add. But good to keep in mind if we see issues on the initial testing phase.

astrozzc (approved):
LGTM. Will leave for Libor to approve. I think he has some comments about setting the the offset
@lpichler

rbac/management/principal/cleaner.py
```diff
@@ -129,145 +119,170 @@ def clean_tenants_principals():
...
+    Returns:
+        bool: True if processing was successful, False otherwise
     """
     with transaction.atomic():
         # This is locked per transaction to ensure another listener process does not run concurrently.
         if not _lock_listener():
             # If there is another listener, let it run and abort this one.
-            logger.info("process_umb_event: Another listener is running. Aborting.")
+            logger.info("process_kafka_message: Another listener is running. Aborting.")
             return False
```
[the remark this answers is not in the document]
shannon-donahue: Fixed in 234158904fb94e132c76c4bd498cd65a8c753385. The consumer loop now distinguishes between lock contention (breaks loop) and per-message processing errors (logs, increments metrics, and
  continues to next message).

rbac/management/principal/cleaner.py
```diff
@@ -129,145 +119,170 @@ def clean_tenants_principals():
...
-            raise e
+
+    # Build Kafka consumer configuration
+    kafka_config = {
+        "bootstrap_servers": settings.KAFKA_SERVERS,
+        "group_id": f"{settings.SA_NAME}-principal-cleanup",
+        "auto_offset_reset": "earliest",
+        "enable_auto_commit": True,
+        "value_deserializer": lambda m: m.decode("utf-8"),
+        "consumer_timeout_ms": 15000,  # 15 second timeout, similar to UMB
```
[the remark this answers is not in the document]
shannon-donahue: Addressed in 5775085989d28335117d6edbf29af516883dc03f. Disabled auto-commit and implemented manual offset commits after successful message processing to ensure at-least-once semantics.

rbac/management/principal/cleaner.py
```diff
@@ -129,145 +119,170 @@ def clean_tenants_principals():
...
+            # Parse JSON message
+            message_data = json.loads(message.value)
+            canonical_message = message_data.get("CanonicalMessage", message_data)
 
             user = retrieve_user_info(canonical_message)
             # By default, only process disabled users.
             # If the setting is enabled, process all users.
-            if not user.is_active or settings.PRINCIPAL_CLEANUP_UPDATE_ENABLED_UMB:
+            if not user.is_active or settings.PRINCIPAL_CLEANUP_UPDATE_ENABLED_KAFKA:
                 # If Tenant is not already ready, don't ready it
```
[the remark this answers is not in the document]
shannon-donahue: Fixed in 6e912cab18ad77df12e2d58aaa8f8c51c13982d9. Initialized consumer = None before the try block and added conditional check in finally block to only close if consumer was successfully created.

Dockerfile
```diff
@@ -71,7 +71,8 @@ COPY Pipfile .
 COPY Pipfile.lock .
 RUN \
     # install the dependencies into the working dir (i.e. ${APP_ROOT}/.venv)
-    pipenv install --deploy && \
+    # NOTE: Using --skip-lock temporarily due to Pipfile.lock dependency conflicts
+    pipenv install --skip-lock && \
```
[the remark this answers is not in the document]
shannon-donahue: accidental change while working through local testing errors. Reverted back to original no change here

rbac/management/principal/cleaner.py
shannon-donahue: Fixed in 246f1b3dfa04d139b1ca4ee3b3fd057e2cc5d2fa. Added KAFKA_PRINCIPAL_CLEANUP_TOPIC setting and configured it in rbac-clowdapp.yml.

shannon-donahue: Addressed in 5775085989d28335117d6edbf29af516883dc03f and b408d0e6706fb2771a9b3e010324ed2a600352c0. Implemented manual offset commits with distinction between permanent errors (sent to DLQ, offset
  committed) and transient errors (loop breaks, offset not committed for retry).

shannon-donahue:
> ## Code Review Findings
> **File:** `rbac/rbac/celery.py` (line 1)
> 
> **low**: [CR-005] Confusing naming: DELETION_ENABLED controls scheduling, not deletion
> 
> PRINCIPAL_CLEANUP_DELETION_ENABLED_KAFKA controls whether the Celery beat task is scheduled, not whether deletion occurs. The actual cleanup always processes disabled users regardless of this flag.
> 
> **Suggestion:** Consider renaming or adding clarifying comments.

This naming matches the existing UMB implementation pattern for consistency. The flag controls whether the scheduled task runs, which indirectly controls whether deletions occur.  I believe this change should go in a separate PR since it would require refactoring both the UMB and Kafka paths. I will add a comment though detailing this 👍

rbac/management/principal/cleaner.py
shannon-donahue: On my final passthrough I noticed this fix seems to have gotten removed to match UMB pattern. However this fix was intentional so I added it back in with a comment 👍

shannon-donahue: comment was left on a previous commit for this just adding replies for history 👍

deploy/rbac-clowdapp.yml
```diff
@@ -793,6 +795,10 @@ objects:
             value: ${RBAC_KAFKA_CONSUMER_GROUP_ID}
           - name: RBAC_KAFKA_CUSTOM_CONSUMER_BROKER
             value: ${RBAC_KAFKA_CUSTOM_CONSUMER_BROKER}
+          - name: KAFKA_PRINCIPAL_CLEANUP_TOPIC
+            value: ${KAFKA_PRINCIPAL_CLEANUP_TOPIC}
+          - name: KAFKA_PRINCIPAL_CLEANUP_DLQ_TOPIC
+            value: ${KAFKA_PRINCIPAL_CLEANUP_DLQ_TOPIC}
```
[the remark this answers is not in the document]
shannon-donahue: I added the variables to common_env_vars. However I did not add to the docker-compose.yml since UMB does not follow that design pattern. @lpichler @astrozzc What do you think with this comment?

rbac/feature_flags.py
```diff
@@ -182,5 +184,72 @@ def is_v2_edit_api_enabled(self, org_id: str) -> bool:
...
+            # Validate mode - only kafka_shadow and kafka_active are valid when flag is enabled
+            if mode == "kafka_shadow":
+                return "kafka_shadow"
+            else:
+                # Default to kafka_active for any other variant
+                return "kafka_active"
+
+        except Exception as e:
+            logger.warning(f"Error getting variant for principal cleanup mode: {e}, defaulting to kafka_active")
+            return "kafka_active"
```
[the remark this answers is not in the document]
shannon-donahue: Defaulting now to running umb only rather than kafka in terms of potential variant failures.

rbac/management/principal/cleaner.py
```diff
@@ -245,6 +355,179 @@ def process_umb_event(frame, umb_client: Stomp, bootstrap_service: TenantBootstr
...
+            # Transient errors: Network issues, DB connection problems, temporary service unavailability
+            is_permanent_error = isinstance(
+                e,
+                (
+                    json.JSONDecodeError,  # Malformed JSON
+                    KeyError,  # Missing required field in message
+                    ValueError,  # Invalid data format
+                    UnicodeDecodeError,  # Invalid message encoding
+                ),
+            )
```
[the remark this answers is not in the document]
shannon-donahue: Added code change to treat malformed xml structures as permanent errors.

rbac/management/principal/cleaner.py
```diff
@@ -245,6 +355,179 @@ def process_umb_event(frame, umb_client: Stomp, bootstrap_service: TenantBootstr
...
+
+            # For production mode, only send permanent errors to DLQ
+            # Transient errors will not commit offset and will retry
+            if not is_permanent_error:
+                logger.warning(
+                    "process_kafka_message: Transient error at offset %d. "
+                    "Will not commit offset - message will be retried on next consumer run.",
+                    message.offset,
+                )
+                return MessageProcessingResult(should_continue=True, success=False)
```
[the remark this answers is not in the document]
shannon-donahue: Fixed dry-run/shadow mode to always return success=True for both transient AND permanent errors. Previously, permanent errors would fall through to DLQ logic and could return success=False if DLQ
  wasn't configured or DLQ send failed, blocking shadow mode validation. Now dry-run mode always continues processing regardless of error type, allowing it to expose all problematic messages without
  blocking

rbac/management/principal/cleaner.py
```diff
@@ -245,6 +355,179 @@ def process_umb_event(frame, umb_client: Stomp, bootstrap_service: TenantBootstr
...
+                        return MessageProcessingResult(should_continue=True, success=True)
+                    except Exception as dlq_error:
+                        logger.error(
+                            "process_kafka_message: Failed to send message to DLQ: %s. "
+                            "Message will be retried on restart.",
+                            str(dlq_error),
+                        )
+                        capture_exception(dlq_error)
+                        # DLQ send failed, so don't commit offset (will retry message)
+                        return MessageProcessingResult(should_continue=True, success=False)
```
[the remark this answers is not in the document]
shannon-donahue: Moved DLQ send outside the transaction.atomic() block to avoid holding the advisory lock and DB connection during network I/O. The code now sets flags (send_to_dlq, error_for_dlq) inside the
  transaction, then performs the actual Kafka send after the transaction commits and locks are released. This prevents slow or hanging DLQ sends from blocking other consumers. Also added a comment
  noting that original_message contains PII and requires appropriate access controls on the DLQ topic.

rbac/management/principal/cleaner.py
```diff
@@ -270,6 +553,128 @@ def process_principal_events_from_umb(bootstrap_service: Optional[TenantBootstra
...
+    # consumer (like launch-rbac-kafka-consumer) would be more appropriate, but this
+    # approach maintains compatibility with the existing UMB-based architecture.
+    kafka_config = {
+        "bootstrap_servers": settings.KAFKA_SERVERS,
+        "group_id": f"{settings.SA_NAME}-principal-cleanup",
+        "auto_offset_reset": "earliest",
+        "enable_auto_commit": False,  # Manual commit for at-least-once semantics
+        "value_deserializer": lambda m: m.decode("utf-8"),
+        "consumer_timeout_ms": 15000,  # 15 second timeout per run, matches UMB behavior
+    }
```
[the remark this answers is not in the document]
shannon-donahue: Removed value_deserializer from Kafka consumer config and moved UTF-8 decoding inside process_kafka_message's try block. This allows UnicodeDecodeError to be caught by existing error handling and
  classified as a permanent error for DLQ. Also added tombstone detection (null value check) to handle Kafka delete messages gracefully. Invalid UTF-8 and tombstones now reach the error handling/DLQ
  logic instead of crashing the consumer before process_kafka_message is called.

rbac/management/tasks.py
```diff
@@ -26,26 +26,92 @@
...
+    mode = FEATURE_FLAGS.get_principal_cleanup_mode()
+    logger.info(f"Principal cleanup mode: {mode}")
+
+    if mode == "umb_only":
+        # UMB-only mode: Only UMB processes and writes to DB
+        if settings.UMB_JOB_ENABLED:
+            logger.info("UMB-only mode: processing via UMB")
+            process_principal_events_from_umb()
+        else:
+            logger.warning("UMB mode selected but UMB_JOB_ENABLED is False")
```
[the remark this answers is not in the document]
shannon-donahue: Added UMB fallback for kafka_active mode when KAFKA_PRINCIPAL_CLEANUP_JOB_ENABLED is disabled. This ensures principal cleanup continues running via UMB if Kafka is temporarily unavailable, matching
  the fallback pattern already present in the unknown-mode branch.

  Note: Did not add fallback for umb_only mode because UMB is already the baseline method - if UMB_JOB_ENABLED is false in umb_only mode, there's no sensible fallback (both the primary and fallback
  methods would be disabled). The warning-only behavior for umb_only mode is correct.

tests/test_feature_flags.py
```diff
@@ -125,3 +125,130 @@ def test_is_use_role_binding_view_permission_enabled_defaults_to_true(self):
...
+            FEATURE_FLAGS.client.get_variant = mock_get_variant
+
+            mode = FEATURE_FLAGS.get_principal_cleanup_mode()
+            self.assertEqual(mode, "kafka_active")
+
+            # Restore original methods
+            FEATURE_FLAGS.client.get_variant = original_get_variant
+
+        FEATURE_FLAGS.is_enabled = original_is_enabled
+
```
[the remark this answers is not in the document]
shannon-donahue: Moved the get_principal_cleanup_mode() calls and assertions outside the if FEATURE_FLAGS.client: conditional blocks in both tests. Now:
  - Both tests explicitly fail with self.assertIsNotNone(FEATURE_FLAGS.client, "FEATURE_FLAGS.client is None - cannot run test") if the client is unavailable after initialization
  - The actual test assertions always run (not skipped when client is None)
  - Tests properly validate their intended behavior in all scenarios

  This ensures the tests cannot silently pass without performing their intended checks.

tests/test_feature_flags.py
```diff
@@ -125,3 +125,130 @@ def test_is_use_role_binding_view_permission_enabled_defaults_to_true(self):
         # which defaults to True
         self.assertTrue(settings.USE_ROLE_BINDING_VIEW_PERMISSION)
         self.assertTrue(FEATURE_FLAGS.is_use_role_binding_view_permission_enabled())
+
+    def test_get_principal_cleanup_mode_defaults_to_umb_only_without_client(self):
+        """Test that get_principal_cleanup_mode returns 'umb_only' when client is not initialized."""
+        FEATURE_FLAGS.client = None
+        mode = FEATURE_FLAGS.get_principal_cleanup_mode()
+        self.assertEqual(mode, "umb_only")
```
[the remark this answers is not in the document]
shannon-donahue: @patch("feature_flags.FEATURE_FLAGS.initialize") to prevent re-initialization during the test. This ensures the test properly exercises the client is None fallback path in
  get_principal_cleanup_mode() (which returns "umb_only").

tests/management/principal/test_kafka_cleaner.py
```diff
@@ -0,0 +1,1188 @@
...
+        consumer_instance = MagicMock()
+        consumer_instance.__iter__.return_value = iter([mock_message])
+        consumer_mock.return_value = consumer_instance
+
+        process_principal_events_from_kafka()
+
+        # Principal should be deleted when inactive and not found in proxy
+        self.assertFalse(Principal.objects.filter(username=principal_name).exists())
+        # No replication events should be produced for non-V2 tenants
+        consumer_instance.close.assert_called_once()
```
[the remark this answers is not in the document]
shannon-donahue: Added @patch for OutboxReplicator.replicate and assertion to verify it was not called. The test now properly validates its purpose - ensuring non-V2 tenants (with V2_BOOTSTRAP_TENANT=False) don't
  produce replication events when processing disabled users. Retained all existing assertions for principal deletion and consumer cleanup.

tests/management/principal/test_kafka_cleaner.py
```diff
@@ -0,0 +1,1188 @@
...
+        mock_message = create_mock_kafka_message(KAFKA_MESSAGE_BODY)
+        consumer_instance = MagicMock()
+        consumer_instance.__iter__.return_value = iter([mock_message])
+        consumer_mock.return_value = consumer_instance
+
+        process_principal_events_from_kafka()
+        self.assertTrue(Principal.objects.filter(username=principal_name).exists())
+
+        consumer_instance.__iter__.return_value = iter([create_mock_kafka_message(KAFKA_MESSAGE_SPECIAL)])
+        process_principal_events_from_kafka()
```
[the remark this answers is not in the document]
shannon-donahue: Added assertion after the second process_principal_events_from_kafka() call to verify that processing KAFKA_MESSAGE_SPECIAL (another non-existent user message) doesn't delete the existing
  "principal-keep" principal. This ensures the test validates behavior for both message processing cycles instead of leaving the second call unverified.

rbac/management/tasks.py
```diff
@@ -26,26 +26,92 @@
...
+            logger.info("Shadow mode: Running UMB consumer (active - writes to DB)")
+            process_principal_events_from_umb()
+        else:
+            logger.warning("Shadow mode requires UMB but UMB_JOB_ENABLED is False")
+
+        if settings.KAFKA_PRINCIPAL_CLEANUP_JOB_ENABLED:
+            logger.info("Shadow mode: Running Kafka consumer (dry-run - no DB writes)")
+            process_principal_events_from_kafka(dry_run=True)
+        else:
+            logger.warning("Shadow mode requires Kafka but KAFKA_PRINCIPAL_CLEANUP_JOB_ENABLED is False")
```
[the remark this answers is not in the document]
shannon-donahue: Added separate try/except blocks around UMB and Kafka consumer calls in shadow mode. Each consumer now runs independently - if UMB fails, it's logged and sent to Sentry, but Kafka dry-run still
  executes. This ensures shadow mode can validate Kafka behavior even when UMB has issues, which is critical for safe migration testing. Preserved all existing feature flag checks and execution order.

rbac/management/principal/cleaner.py
```diff
@@ -245,6 +356,210 @@ def process_umb_event(frame, umb_client: Stomp, bootstrap_service: TenantBootstr
...
+                    message.value.decode("utf-8") if isinstance(message.value, bytes) else message.value
+                ),
+                "error": str(error_for_dlq),
+                "error_type": type(error_for_dlq).__name__,
+                "partition": message.partition,
+                "offset": message.offset,
+                "timestamp": message.timestamp,
+                "dry_run": dry_run,
+            }
+            dlq_producer.send_kafka_message(dlq_topic, dlq_message)
```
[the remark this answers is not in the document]
shannon-donahue: Skipping - current design is intentional and correct

  The DLQ send happens outside the transaction intentionally to avoid holding the database advisory lock during network I/O. This is appropriate for DLQ messages because:

  1. DLQ is best-effort diagnostics - These are debugging messages, not business data that requires exactly-once delivery guarantees
  2. Offset commit is atomic - The critical operation (committing the Kafka offset) happens via the function return value, independent of DLQ send success
  3. Outbox adds unnecessary complexity - The outbox pattern is designed for business events that need transactional guarantees. DLQ messages don't need this.
  4. Failure is acceptable - If the DLQ send fails (network issue, Kafka unavailable), it's logged but doesn't block message processing. The message won't be reprocessed.

  The current design follows the principle of "don't hold locks during I/O" and treats DLQ as what it is: a best-effort debugging tool, not a critical data path

rbac/management/principal/cleaner.py
```diff
@@ -245,6 +356,210 @@ def process_umb_event(frame, umb_client: Stomp, bootstrap_service: TenantBootstr
...
+    if send_to_dlq and error_for_dlq:
+        dlq_topic = settings.KAFKA_PRINCIPAL_CLEANUP_DLQ_TOPIC
+        try:
+            # Build DLQ message with error context
+            # NOTE: original_message may contain PII (usernames, org/account IDs)
+            # Ensure DLQ topic has appropriate access controls and retention policy
+            dlq_message = {
+                "original_message": (
+                    message.value.decode("utf-8") if isinstance(message.value, bytes) else message.value
+                ),
```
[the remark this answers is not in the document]
shannon-donahue: Updated DLQ message construction to preserve invalid UTF-8 payloads by base64-encoding them when UTF-8 decode fails. This prevents the catch-22 where poison messages (e.g., UnicodeDecodeError) 
  couldn't be delivered to the DLQ for inspection.

  Changes:
  1. Try UTF-8 decode first - if successful, use decoded string
  2. On UnicodeDecodeError, base64-encode the raw bytes to preserve them
  3. Include message_encoding metadata field to indicate encoding used ("utf-8", "base64", or "string")
  4. Fixed dry-run mode to test DLQ functionality by sending permanent errors to DLQ

  This ensures all poison messages can be captured in the DLQ regardless of encoding issues, while preserving the original bytes for later analysis.

rbac/feature_flags.py
```diff
@@ -182,5 +184,75 @@ def is_v2_edit_api_enabled(self, org_id: str) -> bool:
...
+        )
+
+    def get_principal_cleanup_mode(self) -> str:
+        """
+        Get the principal cleanup mode using the feature flag.
+
+        This method supports 3 modes controlled by the Unleash flag value:
+        - 'umb_only' (flag disabled): Only UMB consumer runs and writes to DB
+        - 'kafka_shadow' (flag enabled with variant): Both UMB and Kafka run, only UMB writes (Kafka dry-run)
+        - 'kafka_active' (flag enabled, default): Only Kafka consumer runs and writes to DB
```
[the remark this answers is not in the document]
shannon-donahue: Fixed - Updated the mode descriptions in both rbac/feature_flags.py (lines 205-208) and rbac/management/tasks.py (lines 101-104) to clarify that:
  - kafka_active requires an explicitly configured variant (removed misleading "default" label)
  - kafka_shadow requires the kafka_shadow variant specifically
  - umb_only is used when the flag is disabled OR when an unknown/missing variant is encountered (the actual safe fallback)

  The documentation now accurately reflects the runtime resolution logic where unknown variants fall back to umb_only mode.

rbac/management/principal/cleaner.py
```diff
@@ -245,6 +357,231 @@ def process_umb_event(frame, umb_client: Stomp, bootstrap_service: TenantBootstr
...
+                    user.org_id,
+                    user.is_active,
+                )
+
+                # Validate message structure is correct but DON'T call update_user (no DB writes)
+                if not user.is_active or settings.PRINCIPAL_CLEANUP_UPDATE_ENABLED_KAFKA:
+                    logger.info(
+                        "DRY RUN: Would call bootstrap_service.update_user() for user %s",
+                        user.username,
+                    )
```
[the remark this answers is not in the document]
shannon-donahue: Fixed - Changed the dry-run logging in rbac/management/principal/cleaner.py to:
  - Downgraded from logger.info to logger.debug (both calls at lines 434 and 441)
  - Removed data fields left only needed DRY RUN messages for debugging purposes

rbac/management/principal/cleaner.py
```diff
@@ -245,6 +357,231 @@ def process_umb_event(frame, umb_client: Stomp, bootstrap_service: TenantBootstr
...
+        # Return success=True so offset gets committed (message moved to DLQ)
+        return MessageProcessingResult(should_continue=True, success=True)
+    except Exception as dlq_error:
+        logger.error(
+            "process_kafka_message: Failed to send message to DLQ: %s. " "Message will be retried on restart.",
+            str(dlq_error),
+        )
+        capture_exception(dlq_error)
+        # DLQ send failed, so don't commit offset (will retry message)
+        return MessageProcessingResult(should_continue=True, success=False)
```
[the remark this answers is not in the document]
shannon-donahue: Fixed - Updated the DLQ exception handler in rbac/management/principal/cleaner.py to:
  - Check if dry_run mode is active when DLQ send fails
  - Return success=True (commit offset) in dry-run mode to allow processing to continue
  - Preserve success=False (retry message) behavior for normal production processing
  - Added warning log to indicate dry-run DLQ failure handling

lpichler (commented):
Review findings from code-review-2 — 3 actionable items posted inline. Note: ISSUE-1 (bytes-branch dead code) has been addressed; the consumer no longer has a value_deserializer, making the `isinstance(message.value, bytes)` check intentional and correct.

rbac/rbac/settings.py
```diff
@@ -530,6 +530,9 @@ def _parse_logging_handlers(raw_value):
...
+KAFKA_PRINCIPAL_CLEANUP_DLQ_TOPIC = ENVIRONMENT.get_value("KAFKA_PRINCIPAL_CLEANUP_DLQ_TOPIC", default="")
```
lpichler, line 534: **[medium]**: New Kafka topics bypass Clowder topic resolution

These settings read topic names directly from env vars, but other Kafka topics in this file go through Clowder's `KafkaTopics.get()` mapping (see nearby lines). In deployed environments managed by Clowder, the actual topic names may differ from the env var values, causing consumers to read from the wrong topic.

**Suggestion**: Resolve through Clowder mapping like the other topics:
```python
clowder_cleanup_topic = KafkaTopics.get(KAFKA_PRINCIPAL_CLEANUP_TOPIC)
if clowder_cleanup_topic:
    KAFKA_PRINCIPAL_CLEANUP_TOPIC = clowder_cleanup_topic.name
```

rbac/rbac/celery.py
```diff
@@ -56,13 +56,30 @@
...
+    }
+elif settings.PRINCIPAL_CLEANUP_DELETION_ENABLED_UMB:
     if settings.UMB_JOB_ENABLED:  # TODO: This is temp flag, remove it after populating user_id
         app.conf.beat_schedule["principal-cleanup-every-minute"] = {
             "task": "management.tasks.principal_cleanup_via_umb",
             "schedule": 60,  # Every 60 second
             "args": [],
         }
+elif settings.PRINCIPAL_CLEANUP_DELETION_ENABLED_KAFKA:
+    if settings.KAFKA_PRINCIPAL_CLEANUP_JOB_ENABLED:
```
lpichler, line 77: **[medium]**: Scheduling gap — Kafka-only path silently drops all cleanup when job is disabled

When `PRINCIPAL_CLEANUP_DELETION_ENABLED_KAFKA=True` and `KAFKA_PRINCIPAL_CLEANUP_JOB_ENABLED=False`, this `elif` block is entered but the inner `if` fails, so nothing is added to `beat_schedule`. The outer `else` (7-day fallback) is unreachable because the `elif` already matched. Principal cleanup will not run at all in this configuration.

**Suggestion**: Add a fallback inside this block so the 7-day cleanup still runs when the Kafka job flag is off:
```python
elif settings.PRINCIPAL_CLEANUP_DELETION_ENABLED_KAFKA:
    if settings.KAFKA_PRINCIPAL_CLEANUP_JOB_ENABLED:
        app.conf.beat_schedule["principal-cleanup-every-minute"] = { ... }
    else:
        app.conf.beat_schedule["principal-cleanup-every-sevenish-days"] = { ... }
```

scripts/test_kafka_producer.py
```diff
@@ -0,0 +1,76 @@
...
+    )
+
+    print(f"Sending test message to topic: {topic}")
+    future = producer.send(topic, value=SAMPLE_MESSAGE)
+
+    # Wait for the message to be sent
+    try:
+        record_metadata = future.get(timeout=10)
+        print(f"✓ Message sent successfully!")
+        print(f"  Topic: {record_metadata.topic}")
```
lpichler, line 58: **[info]**: Emoji characters in print statements

The `✓` and `✗` symbols can cause encoding issues in some terminal environments and CI log parsers.

**Suggestion**: Replace with plain-text markers:
```python
print("[OK] Message sent successfully!")
print(f"[FAIL] Failed to send message: {e}")
```

lpichler:
**[medium]**: PR description incorrectly states dependencies were removed

The PR description says _"Removed stompest and xmltodict from Pipfile"_ but `Pipfile` is not changed in this PR — both packages remain as dependencies (they are still required by the UMB code path that coexists behind the feature flag).

**Suggestion**: Update the PR description to accurately reflect that `stompest` and `xmltodict` are retained as long as the UMB path remains active behind the Unleash flag.

lpichler (commented):
Dimensional review finding 1/10

rbac/management/principal/cleaner.py
```diff
@@ -245,6 +369,227 @@ def process_umb_event(frame, umb_client: Stomp, bootstrap_service: TenantBootstr
...
+                return MessageProcessingResult(should_continue=True, success=True)
+
+            # Parse message - handle both XML (from UMB bridge) and JSON (from native Kafka producer)
+            # The messaging bridge copies raw UMB message bodies (XML) to Kafka during migration
+            # Decode here (not in consumer config) so UTF-8 errors reach our error handling/DLQ logic
+            message_value = message.value.decode("utf-8") if isinstance(message.value, bytes) else message.value
+
+            # Detect format: XML starts with '<' or '<?xml', JSON starts with '{' or '['
+            message_value_stripped = message_value.strip()
+            if message_value_stripped.startswith("<"):
```
lpichler: **[HIGH]**: XML/JSON format detection is fragile

Detecting format by `startswith("<")` breaks when the payload has leading whitespace (e.g. `" <CanonicalMessage>"` is misrouted to the JSON path). There is also no validation that a non-XML message is actually valid JSON before calling `json.loads`.

**Suggestion**: Try JSON parse first and fall back to XML on `JSONDecodeError`; if both fail, send to DLQ with a clear error message.

rbac/management/principal/cleaner.py
```diff
@@ -245,6 +369,227 @@ def process_umb_event(frame, umb_client: Stomp, bootstrap_service: TenantBootstr
...
+    Returns:
+        MessageProcessingResult with:
+        - should_continue: False if another listener is running (lock contention), True otherwise
+        - success: True if message was processed successfully, False if it failed
+    """
+    # Track error for DLQ send outside transaction (to avoid holding DB lock during network I/O)
+    send_to_dlq = False
+    error_for_dlq = None
+
+    with transaction.atomic():
```
lpichler, line 406: **[HIGH]**: Exception handling inside `transaction.atomic()` can leave DB in inconsistent state

When a permanent error is caught and `send_to_dlq` is set, the `transaction.atomic()` block exits normally (no rollback), so any partial DB writes from earlier in the same transaction are committed even though the message is going to the DLQ.

**Suggestion**: Call `transaction.set_rollback(True)` before setting `send_to_dlq = True`, or move `update_user` into a nested savepoint so its effects can be rolled back independently while the outer transaction (holding the advisory lock) commits cleanly.

lpichler (commented):
Dimensional review finding 2/10

rbac/management/principal/cleaner.py
```diff
@@ -270,6 +615,128 @@ def process_principal_events_from_umb(bootstrap_service: Optional[TenantBootstra
...
+
+    # Build Kafka consumer configuration
+    # NOTE: This consumer runs periodically via Celery beat (every 60s) and consumes for 15s,
+    # creating a 45-second gap between consumption periods. This matches the UMB behavior
+    # where the consumer also ran periodically. For continuous consumption, a persistent
+    # consumer (like launch-rbac-kafka-consumer) would be more appropriate, but this
+    # approach maintains compatibility with the existing UMB-based architecture.
+    kafka_config = {
+        "bootstrap_servers": settings.KAFKA_SERVERS,
+        "group_id": f"{settings.SA_NAME}-principal-cleanup",
```
lpichler: **[MEDIUM]**: Consumer group ID collides across environments sharing the same Kafka cluster

`group_id` is derived solely from `SA_NAME` (e.g. `nonprod-hcc-rbac-principal-cleanup`). In multi-env setups (staging, ephemeral, CI) that share a Kafka cluster, all environments join the same consumer group, causing offset interference and message loss.

**Suggestion**: Include an environment discriminator: `f"{settings.SA_NAME}-{settings.ENV_NAME}-principal-cleanup"`. `EPH_ENV` is already propagated via the ClowdApp manifest and could serve this purpose.

lpichler (commented):
Dimensional review finding 3/10

rbac/management/principal/cleaner.py
```diff
@@ -245,6 +369,227 @@ def process_umb_event(frame, umb_client: Stomp, bootstrap_service: TenantBootstr
...
+            else:
+                logger.warning(
+                    "process_kafka_message: No DLQ producer configured. "
+                    "Failed message at offset %d will be retried on restart.",
+                    message.offset,
+                )
+                return MessageProcessingResult(should_continue=True, success=False if not dry_run else True)
+
+    # Transaction and advisory lock are now released - safe to do network I/O
+    # At this point, send_to_dlq must be True (only code path that doesn't return in except block above)
```
lpichler: **[HIGH]**: Race window between transaction commit and DLQ send

After `transaction.atomic()` exits, the advisory lock is released. A second consumer can pick up the same message before the DLQ send completes. If the DLQ send then fails, the offset is not committed and the message is retried, potentially processing it twice.

**Suggestion**: Either include the DLQ send within the transaction boundary, or document this as a known race condition with a monitoring alert on DLQ-send failures so operators can correlate duplicate processing events.

lpichler (commented):
Dimensional review finding 4/10

rbac/management/principal/cleaner.py
```diff
@@ -189,13 +233,93 @@ def retrieve_user_info(message) -> User:
         # identifiers["Reference"] might be a dict
         if not isinstance((refs := identifiers["Reference"]), list):
             refs = [identifiers["Reference"]]
+        # BUG FIX: Original UMB code had break statements that would only extract org_id OR account,
+        # not both. Removing breaks to extract both values when present in the message.
```
lpichler: **[HIGH]**: `break` removal in `retrieve_user_info_umb()` changes existing production UMB behavior

Removing the `break` statements changes semantics: previously only the first matching `WEB/Customer` or `EBS/Account` reference was used; now all references are iterated and later ones overwrite earlier ones. The comment calls this a bug fix, but the UMB path serves live production traffic and there are no UMB-specific tests covering messages with multiple `WEB/Customer` or `EBS/Account` references.

**Suggestion**: Add UMB tests covering multi-reference messages to confirm the new behavior is correct, or separate this behavioral change into its own PR.

lpichler (commented):
Dimensional review finding 5/10

rbac/feature_flags.py
```diff
@@ -184,6 +186,77 @@ def is_v2_edit_api_enabled(self, org_id: str) -> bool:
...
+                self.TOGGLE_USE_KAFKA_CLEANUP,
+                fallback_variant={"name": "umb_only", "enabled": False},
+            )
+
+            mode = variant.get("name", "umb_only")
+
+            # Validate mode - only these variants are valid when flag is enabled
+            if mode in ["kafka_shadow", "kafka_validation", "kafka_active"]:
+                return mode
+            else:
```
lpichler, line 252: **[HIGH]**: `get_principal_cleanup_mode()` returns `umb_only` when flag is enabled but has no recognized variant

When Unleash enables the flag but returns its default variant (name=`disabled` or any unrecognized string), the `else` branch returns `"umb_only"`. The docstring says enabling the flag defaults to `kafka_active`, but the implementation contradicts this: an operator who enables the flag without setting a variant silently stays on UMB.

**Suggestion**: Change the `else` branch to `return "kafka_active"` when `is_enabled` is `True` and no recognized variant is set, or add `"disabled"` as an explicit alias for `"kafka_active"` in the valid-mode list.

lpichler (commented):
Dimensional review finding 6/10

rbac/management/tasks.py
```diff
@@ -75,11 +80,145 @@ def principal_cleanup():
...
+        else:
+            # Kafka succeeded - log for comparison purposes
+            logger.info(
+                "Validation mode: Kafka succeeded, UMB fallback not needed. "
+                "Monitor metrics to compare Kafka vs historical UMB performance."
+            )
+
+    elif mode == "kafka_active":
+        # Kafka-active mode: Only Kafka processes and writes to DB
+        if settings.KAFKA_PRINCIPAL_CLEANUP_JOB_ENABLED:
```
lpichler, line 201: **[HIGH]**: `kafka_active` mode silently falls back to UMB when `KAFKA_PRINCIPAL_CLEANUP_JOB_ENABLED=False`

When an operator selects `kafka_active` mode via the Unleash flag but `KAFKA_PRINCIPAL_CLEANUP_JOB_ENABLED` is `False`, the task silently processes via UMB with only a warning log. Operators monitoring Kafka metrics will see no activity and incorrectly assume cleanup is not running.

**Suggestion**: Raise an exception (or emit an error metric) when `kafka_active` mode is selected but Kafka is disabled. Operators should make an explicit flag change to `umb_only` rather than relying on this silent fallback.

lpichler (commented):
Dimensional review finding 7/10

rbac/management/tasks.py
```diff
@@ -26,6 +26,12 @@
 from celery import shared_task
 from django.conf import settings
 from django.core.management import call_command
+from management.principal.cleaner import (
```
lpichler: **[MEDIUM]**: Module-level imports break the established lazy-import convention

All existing tasks import from `management.principal.cleaner` inside the task function body. Moving these to module-level deviates from the established pattern and introduces circular-import risk at Django startup time.

**Suggestion**: Move `from management.principal.cleaner import ...` and `from sentry_sdk import capture_exception` back inside the task function bodies to match the existing convention.

lpichler (commented):
Dimensional review finding 8/10

deploy/rbac-clowdapp.yml
```diff
@@ -2036,6 +2062,12 @@ parameters:
 - name: RBAC_KAFKA_CUSTOM_CONSUMER_BROKER
   description: Custom Kafka broker URL for the RBAC Kafka consumer (if empty, uses default Clowder/localhost configuration)
   value: ''
+- name: KAFKA_PRINCIPAL_CLEANUP_TOPIC
```
lpichler, line 2070: **[HIGH]**: `KAFKA_PRINCIPAL_CLEANUP_TOPIC` has an empty default but is required when Kafka cleanup is enabled

`value: ''` means the topic silently defaults to an empty string. When `PRINCIPAL_CLEANUP_DELETION_ENABLED_KAFKA=true` is set but the topic parameter is omitted, the consumer logs an error and returns early rather than failing fast at startup.

**Suggestion**: Add deployment-time validation (init container or Clowder validation hook) that fails the deployment when `KAFKA_PRINCIPAL_CLEANUP_DELETION_ENABLED_KAFKA=true` but `KAFKA_PRINCIPAL_CLEANUP_TOPIC` is empty.

lpichler (commented):
Dimensional review finding 9/10

rbac/management/tasks.py
```diff
@@ -75,11 +80,145 @@ def principal_cleanup():
...
 
 
+@shared_task
+def principal_cleanup_via_kafka():
+    """Celery task to clean up principals via Kafka messages."""
+    process_principal_events_from_kafka()
+
+
+@shared_task
+def principal_cleanup_via_message_bus():
```
lpichler, line 93: **[HIGH]**: No tests for `principal_cleanup_via_message_bus` dispatcher task

The dispatcher task contains critical routing logic with 4 modes (`umb_only`, `kafka_shadow`, `kafka_validation`, `kafka_active`) and several fallback paths. This logic has zero test coverage: the new `test_kafka_cleaner.py` tests only `process_principal_events_from_kafka`, not the dispatcher task itself.

**Suggestion**: Add tests covering all 4 modes and key edge cases: Kafka disabled in `kafka_active` mode, Kafka failure triggering UMB fallback in `kafka_validation` mode, and both backends disabled.

lpichler (commented):
Dimensional review finding 10/10

lpichler (approved):

rbac/feature_flags.py
shannon-donahue: do not want to change from defaulting to UMB only for this change. We have not tested kafka stability so we do not want to default to kafka active. I have updated the docs however to note this.

rbac/rbac/settings.py
shannon-donahue: Added Clowder topic resolution

scripts/test_kafka_producer.py
shannon-donahue: removed these good catch

rbac/rbac/celery.py
shannon-donahue: added an else clause inside the Kafka block that schedules the 7-day fallback cleanup

rbac/management/principal/cleaner.py
shannon-donahue: Fixed "group_id": f"{settings.SA_NAME}-{env_name}-principal-cleanup",

rbac/management/tasks.py
shannon-donahue: Lazy import architecture reinstated.

deploy/rbac-clowdapp.yml
shannon-donahue: Implemented deployment-time validation that fails fast when PRINCIPAL_CLEANUP_DELETION_ENABLED_KAFKA=true but
  KAFKA_PRINCIPAL_CLEANUP_TOPIC is empty.

rbac/management/tasks.py
shannon-donahue: Added additional test coverage for this

rbac/management/principal/cleaner.py
shannon-donahue: Added the breaks back in and followed the same architecture for Kafka so that we have matching behavior when testing.

rbac/management/tasks.py
shannon-donahue: Now raising RuntimeError instead of silent fallback when there's a configuration mismatch

lpichler (approved):

PR #2958, opened 2026-05-29
feat(mcp): redact PII from MCP tool responses

lpichler:
## Summary

Prevent LLM leakage of personal data by redacting sensitive PII fields from MCP tool responses. The LLM does not need emails or last names to perform RBAC investigation tasks — UUIDs and first names are sufficient for identifying and correlating users.

### 1. Per-tool PII redaction via `redacted_fields`

Added `redacted_fields: tuple[str, ...]` to the `ToolConfig` dataclass and `@register_tool()` decorator. Each principal-related tool declares which fields to strip:

```python
_PRINCIPAL_REDACTED_FIELDS = ("email", "last_name")

@register_tool(
    description="...",
    redacted_fields=_PRINCIPAL_REDACTED_FIELDS,
)
def list_principals(...):
```

A new `_strip_pii_from_principal_list()` helper strips declared fields from serialized responses and enriches entries with `Principal.uuid` from the database.

### 2. Tools affected (6)

| Tool | What changes in the response |
|------|------------------------------|
| `list_principals` | `email`, `last_name` stripped; `uuid` added |
| `list_group_principals` | Same |
| `investigate_tam_access` | `requester_info` returns only `{user_id}` instead of full BOP profile |
| `audit_redhat_access` | `user_info` returns only `{user_id}` |
| `audit_group_for_dissolution` | Members identified by `uuid`; `username` removed from member dicts and stranded lists |
| `get_user_state` | Top-level key is `uuid` instead of `username` |

**Unchanged tools:** `check_user_permission`, `investigate_user_access`, `list_audit_logs`, `investigate_group_changes`, `get_rbac_recent_changes` — these use `principal_username` (system identifier from audit logs), not display PII.

### 3. Feature flag: `MCP_PII_REDACTION_ENABLED`

- **Default: `True`** (redaction on)
- Set `MCP_PII_REDACTION_ENABLED=False` to restore full PII in responses (e.g., for internal debugging)
- Read via `settings.MCP_PII_REDACTION_ENABLED` with `_is_pii_redaction_enabled()` helper
- Added to `tox.ini` `[testenv]` for explicitness

### 4. Tool descriptions updated

All affected tools' `@register_tool(description=...)` strings updated to document:
- Which fields are redacted
- That UUIDs are used as identifiers
- Response schema changes (e.g., `{uuid, username, first_name, ...}` instead of `{username, email, first_name, last_name, ...}`)

## Test plan

- [x] All 405 MCP view tests pass (`tox -e py312-fast -- tests.management.test_mcp_views`)
- [x] `test_principal_tools_have_redacted_fields` — verifies all principal-related tools declare `redacted_fields`
- [x] `test_list_principals_pii_visible_when_redaction_disabled` — flag off, PII present
- [x] `test_get_user_state_pii_visible_when_redaction_disabled` — flag off, `username` key instead of `uuid`
- [x] `test_audit_redhat_access_pii_visible_when_redaction_disabled` — flag off, `name`/`email` in `user_info`
- [x] `test_audit_group_for_dissolution_pii_visible_when_redaction_disabled` — flag off, `username` in members
- [x] Existing redaction-on tests assert `email`/`last_name` absent from responses
- [x] `tox -e lint` passes (flake8 + black)


rbac/management/mcp_views.py
```diff
@@ -650,6 +653,37 @@ def _list_principals_by_name(
         },
         default=str,
     )
+    return _strip_pii_from_principal_list(raw_json, tenant=getattr(request, "tenant", None))
+
+
+_PII_FIELDS = ("username", "email", "first_name", "last_name")
+
+
+def _strip_pii_from_principal_list(raw_json: str, tenant=None) -> str:
```
EvanCasey13: Should we add case for checking if tenant is none or principal is not present, this could lead to an empty uuid field?

EvanCasey13 (approved):

lpichler:
## Code Review Findings

### Issue 1 (Low severity): `_strip_pii_from_principal_list` performs DB query per call without caching (Info)

In `_strip_pii_from_principal_list`, when redaction is enabled, the function queries `Principal.objects.filter(username__in=usernames, tenant=tenant)` to resolve UUIDs. This is called once per `list_principals`/`_list_principals_by_name`/`list_group_principals` invocation, so typically just one query per tool call. This is acceptable for current usage patterns but worth noting -- if the function were ever called in a loop, it could cause N+1 issues.

**Verdict**: Acceptable -- not a bug, just a note for future maintainers.

---

### Issue 2 (Low severity): `_list_principals_by_name` with `username_only=True` produces empty objects after redaction (Info)

When `username_only=True`, the upstream API returns `{"username": "jsmith"}` for each entry (single key). After `_strip_pii_from_principal_list` removes `username` and finds no matching Principal in DB (since the proxy data may not match local principals), each entry in `data` becomes `{}` -- an empty dict. The test at line 601-603 now asserts `assertNotIn("username", user)` and `assertNotIn("email", user)` but doesn't check that the resulting objects contain anything useful.

**Verdict**: Minor UX concern -- the LLM consuming these results would see `data: [{}, {}]` which is not very useful. However, `username_only` is a niche parameter and the redaction behavior is consistent with the overall approach. Could be improved by omitting empty entries or adding a note.

---

### Issue 3 (Informational): `investigate_tam_access` filtering still works with PII internally (Info)

The filtering by `requester_name`/`requester_email` correctly uses BOP data internally before redacting the output. The comment `"(uses BOP data internally, not exposed)"` documents this well.

**No bugs found.** The implementation is clean, consistent across all 6 tools, and properly gated behind the feature flag. The feature flag follows the established pattern (Unleash toggle + env var fallback). Tests cover both the redacted (default) and non-redacted paths.

---

### Issue 1 (Low, Observation): `guide_user_access_delegation` reads `username` from redacted response (Info)

**Location**: `rbac/management/mcp_views.py`, line ~4291 (in PR branch)

The `guide_user_access_delegation` tool calls `list_principals()` internally and then reads `user_data.get("username")` from the response. Since `list_principals` now strips the `username` field when PII redaction is enabled, this will produce `{"username": None}` in the `user_info` dict rather than the actual username or being omitted entirely.

**Impact**: Low. The field will contain `None` rather than being absent, which is inconsistent with the redaction pattern used elsewhere (where PII fields are removed entirely). The tool still functions correctly since it uses the `username` input parameter for subsequent operations, not the returned value. However, an LLM reading `"username": null` might be confused.

**Verdict**: Minor inconsistency, not a crash bug. The tool is outside the PR's stated scope (6 tools), but the PR's changes to `list_principals` have this downstream effect.

---

### Issue 2 (Info, Design Note): PII still present in tools outside PR scope (Info)

Several other tools continue to return PII when redaction is enabled:
- `check_user_permission` / `investigate_user_access` — returns `username` in `user_info`
- `list_audit_logs` — returns `principal_username`
- `guide_user_access_delegation` — returns `username` in result
- Error messages in `get_user_state`, `investigate_tam_access` echo back input usernames/emails

These are all outside the PR's stated scope of 6 tools and are documented as such. The error message cases are defensible since they echo caller-provided input. Future PRs could address the remaining tools.

lpichler:
## Code Review Findings

### [4] Scope simplification changes bundled with PII redaction PR (informational)

**File:** `multiple files`

Approximately half the diff relates to scope service simplification (removing binding_scopes_for_role, all_scopes_for_permission, etc.) brought in via merge. PR description does not mention these changes.

---

**File:** `rbac/management/mcp_views.py` (line 1)

**low**: [1] String formatting bug in audit_redhat_access (non-redacted path)

Due to f-string concatenation and operator precedence, .strip() applies only to the last_name f-string, not the full name. Results in trailing space when last_name is empty. Only affects MCP_PII_REDACTION_ENABLED=False path.

**Suggestion:** Wrap the entire name construction in parentheses before calling .strip(): (f"...first_name..." + " " + f"...last_name...").strip()



---

**File:** `rbac/management/mcp_views.py` (line 1)

**informational**: [2] Dead code: _validate_tool_arguments and _sanitize_validation_error

The call to _validate_tool_arguments() was removed from _handle_tools_call(), and its tests were removed, but the function definitions remain as dead code.

**Suggestion:** Either remove the dead functions or restore the validation call.



---

**File:** `rbac/management/mcp_views.py` (line 1)

**informational**: [3] redacted_fields on ToolConfig is metadata-only, not consumed by framework

redacted_fields is set on each tool but never consumed automatically. Redaction is done manually in each tool function. This is acceptable as a foundation for future automation.

**Suggestion:** Consider adding a decorator or middleware that auto-strips redacted_fields, reducing per-tool boilerplate.

PR #2963, opened 2026-06-01
feat(mcp): add lookup_person tool for contractor/vendor terminology

lpichler:
## Summary
- Adds a `lookup_person` MCP tool that forwards to `get_user_state`, with contractor/vendor/consultant/temp/intern keywords prominent in its name and description.
- Fixes the issue where LLMs connected via Lightspeed cannot resolve "Our contractor Joe..." queries because Lightspeed does not read the MCP `instructions` field from initialize responses — LLMs only see tool names and descriptions when selecting which tool to call.

## Test plan
- [x] Existing `MCPGetUserStateTests` pass (16 tests) — `lookup_person` delegates to the same function
- [ ] Verify with Lightspeed that "Our contractor Joe is being offboarded" now routes to `lookup_person`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

New Features:
- Introduce a lookup_person MCP tool that delegates to the existing get_user_state endpoint for RBAC user state retrieval.

### DIFF
--- a/rbac/management/mcp_views.py
+++ b/rbac/management/mcp_views.py
@@ -2995,6 +2995,11 @@ def _check_user_permission_v1(request: HttpRequest, username: str, permission: s
     )
 
 
+_DEFAULT_INCLUDE_GROUP_ROLES = True
+_DEFAULT_INCLUDE_PERMISSIONS = True
+_DEFAULT_AUDIT_LOG_LIMIT = 10
+
+
 @register_tool(
     description=(
         "Get comprehensive state for a specific user, returning all RBAC information in one call. "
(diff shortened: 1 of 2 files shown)


dsova06 (commented):
Clean, well-scoped PR. The approach makes sense given the Lightspeed limitation where LLMs only see tool names and descriptions, not the initialize instructions field.

One note on the Sourcery suggestion to use `**kwargs`: that won't work here. `register_tool` runs `inspect.signature(fn)` to generate the MCP JSON schema that gets sent to clients via `tools/list`. If `lookup_person` used `**kwargs`, the tool would advertise zero parameters and the LLM wouldn't know what arguments to pass. The explicit signature duplication is the right call.

### Something to think about

1. **Parameter default drift** -- since `lookup_person` duplicates the exact signature and defaults from `get_user_state` (e.g., `audit_log_limit=10`), a future change to `get_user_state`'s defaults wouldn't automatically propagate. Might be worth pulling the shared defaults into module-level constants (`_DEFAULT_AUDIT_LOG_LIMIT = 10`) that both functions reference, so they stay in sync. Not urgent since these defaults have been stable, but worth keeping in mind.

dsova06 (approved):
Looks good -- the shared constants address the default drift concern nicely. Clean PR overall.

lpichler:
thanks @dsova06 !

PR #2980, opened 2026-06-05
RHCLOUD-35482: Check resource tenants before allowing binding

randomnetcat:
## Link(s) to Jira
[RHCLOUD-35482](https://redhat.atlassian.net/browse/RHCLOUD-35482)

Add the capability to check resources requested in resource definitions before creating role bindings for them.

This required a considerable refactor of the code for parsing attribute filters; everything is now unified rather than having piecemeal functions that disagreed on how they should be parsed(!). (More cleanup is likely still needed, but this should be good enough for all non-internal APIs.)

This also does not handle checks for internal APIs/migrations; that will need to be done later. Since I believe we've already validated that all existing workspace resource bindings refer to acceptable workspaces, there shouldn't be any current issues. The risk would come when we add a new type of resource and then run a migration on an existing role with an attribute filter that previously meant nothing but now references an out-of-tenant resource.

[RHCLOUD-35482]: https://redhat.atlassian.net/browse/RHCLOUD-35482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Enforce tenant-aware validation of role binding resources by integrating centralized attribute filter parsing, external inventory checks, and binding authorization policies into role creation, update, and migration flows.

New Features:
- Introduce binding authorization policies and inventory-backed checks to ensure role bindings are only created for resources within the requesting tenant.
- Add support for checking resource tenancy via the Inventory API before persisting role bindings, with configurable test-only bypass.
- Provide centralized parsing and interpretation of attribute filters for workspace-related resource definitions, including handling of ungrouped hosts workspaces.

Enhancements:
- Refactor role, workspace, and migration utilities to reuse a shared attribute filter parser instead of ad hoc workspace-specific helpers.
- Extend role view logic to surface clear 403/409 errors when bindings target unauthorized resources or when concurrent role updates are detected.
- Include default tenant and workspace resources in computed binding policies so migrations and dual-write replication remain consistent with new checks.

Tests:
- Add unit and integration tests covering binding authorization policies, resource tenancy checks against Inventory, role view binding validation, dual-write behavior under constrained policies, and updated cleanup utilities.
- Refactor existing role serializer tests to reduce duplication around resource definition filter expectations and to cover additional permission and hierarchy scenarios.

Chores:
- Remove legacy workspace resource utilities and related settings in favor of the new resource definition parsing and binding validation infrastructure.


rbac/management/role/view.py
```diff
@@ -421,14 +473,25 @@ def partial_update(self, request, *args, **kwargs):
...
-                assert_v1_write_allowed(request.tenant)
-                return super().update(request=request, args=args, kwargs=kwargs)
+
+            def do_update(role: Role, check_result: BindingCheckResult):
+                with transaction.atomic():
+                    self._set_binding_check_result(check_result)
+                    assert_v1_write_allowed(request.tenant)
+                    return super(RoleViewSet, self).update(request=request, args=args, kwargs=kwargs)
+
+            return with_checked_bindings(do_update, get_object_or_404(Role.objects.all(), uuid=kwargs["uuid"]))
```
astrozzc: for partial update, VALID_PATCH_FIELDS does not touch access, seems not necessary to check bindings

rbac/management/role/view.py
```diff
@@ -148,6 +160,8 @@ class RoleViewSet(
     ordering_fields = ("name", "display_name", "modified", "policyCount")
     ordering = ("name",)
 
+    _binding_check_result: Optional[BindingCheckResult] = None
```
astrozzc, line 166: putting the variable in the class level might be a bit confusing. consider something suggested by the sourcery _compute_binding_policy_from_request

rbac/management/role/serializer.py
astrozzc: This is not in your PR but could you help using your parse_attribute_filter here too

astrozzc (dismissed):
Overall LGTM! Just some nit comments

rbac/management/role/view.py
randomnetcat: We can't compute it all in one function, since it has two steps that need to run in different contexts (one without a transaction and one within it). We could use a contextvar instead of a class member if you'd prefer that?

rbac/management/role/view.py
randomnetcat: That's meant to address this:
> The risk would come when we add a new type of resource and then run a migration on an existing role with an attribute filter that previously meant nothing but now references an out-of-tenant resource.

If we don't check this, there's a race condition here:
* Someone adds a resource definition *D* for role *R* for `foo.id` referencing a resource outside their tenant. (We don't check this at the time because we don't yet recognize `foo.id`.)
* Later, we start recognizing `foo.id` and creating role bindings for it.
* Before we remove/update *D* because it's unauthorized, *R* is partially updated, creating a role binding to the unauthorized resource.

astrozzc: Oh, good catch, agree

rbac/management/role/view.py
astrozzc: I see, NVM then, let's keep it this way

rbac/management/role/binding_checks.py
```diff
@@ -0,0 +1,257 @@
...
+
+        initial_filter_data = _filter_data_for(role)
+        check_result = check_binding_resources(org_id=role.tenant.org_id, filters=[e[1] for e in initial_filter_data])
+
+        with transaction.atomic():
+            role = (
+                Role.objects.select_related("tenant")
+                .prefetch_related("access", "access__resourceDefinitions")
+                .select_for_update(of=["self"])
+                .get(pk=role.pk)
```
lpichler, line 248: **critical**: [ISSUE-1] select_for_update called with positional tuple instead of of= keyword argument

In with_checked_bindings, the call .select_for_update(("self", "access", "access_resourceDefinitions")) passes a tuple as the first positional argument (nowait) instead of as the of= keyword argument. This makes nowait truthy (causing NOWAIT behavior) and does not specify which tables to lock. Additionally, "access_resourceDefinitions" uses a single underscore instead of Django's double-underscore relationship traversal syntax.

**Suggestion:** Change to .select_for_update(of=("self", "access", "access__resourceDefinitions"))

rbac/management/role/view.py
```diff
@@ -148,6 +160,8 @@ class RoleViewSet(
     ordering_fields = ("name", "display_name", "modified", "policyCount")
     ordering = ("name",)
```
lpichler, line 161: **medium**: [ISSUE-2] _binding_check_result is a class-level attribute shared across viewset instances

The _binding_check_result = None is defined as a class-level attribute on RoleViewSet. While Django creates new viewset instances per request, this pattern is misleading and suggests shared mutable state. It should be initialized in __init__ or initial() for clarity.

**Suggestion:** Initialize _binding_check_result in initial() method instead of as a class attribute.

rbac/management/role/view.py
```diff
@@ -421,14 +473,25 @@ def partial_update(self, request, *args, **kwargs):
...
         try:
-            with transaction.atomic():
-                assert_v1_write_allowed(request.tenant)
-                return super().update(request=request, args=args, kwargs=kwargs)
+
+            def do_update(role: Role, check_result: BindingCheckResult):
+                with transaction.atomic():
+                    self._set_binding_check_result(check_result)
+                    assert_v1_write_allowed(request.tenant)
+                    return super(RoleViewSet, self).update(request=request, args=args, kwargs=kwargs)
```
lpichler, line 484: **medium**: [ISSUE-3] partial_update bypasses tenant filtering with get_object_or_404(Role.objects.all())

The partial_update method uses get_object_or_404(Role.objects.all(), uuid=kwargs["uuid"]) to find the role for with_checked_bindings. This bypasses the tenant isolation implemented in get_queryset(), potentially allowing binding checks against roles belonging to other tenants. The existing code used self.get_queryset() which properly filters by tenant.

**Suggestion:** Replace Role.objects.all() with self.get_queryset() in the get_object_or_404 call.

rbac/management/role/resource_definitions.py
```diff
@@ -0,0 +1,188 @@
...
+                **attribute_filter,
+                "value": new_ids[0],
+            }
+        elif len(new_ids) == 0:
+            # Empty string represents something that cannot match (distinct from None).
+            return {
+                **attribute_filter,
+                "value": "",
+            }
+        else:
```
lpichler, line 195: **low**: [ISSUE-4] updated_attribute_filter_with_ids produces nondeterministic output order from set input

The function converts its Iterable input to a list via list(new_ids). When passed a set (as done by callers), the output order is nondeterministic. For the 'in' operation this means the value list order varies across runs. This is cosmetic but could make debugging harder.

**Suggestion:** Sort non-None IDs for deterministic output.

rbac/rbac/settings.py
```diff
@@ -645,12 +645,6 @@
 READ_YOUR_WRITES_CHANNEL = ENVIRONMENT.get_value("READ_YOUR_WRITES_CHANNEL", default="READ_YOUR_WRITES_CHANNEL")
```
lpichler, line 645: **low**: [ISSUE-5] Removal of WORKSPACE_APPLICATION_NAME and WORKSPACE_RESOURCE_TYPE settings broadens hierarchy expansion scope

The old is_resource_a_workspace checked application name, resource type, and attribute filter key. The new version only checks the attribute filter key (group.id). This means any permission with a group.id attribute filter now gets workspace hierarchy treatment, not just inventory-related ones. The test changes confirm this behavioral difference.

**Suggestion:** Verify this broadened scope is intentional and document the change.

rbac/management/role/binding_checks.py
randomnetcat: This... isn't true? These were possibly all issues in earlier commits, but they're now fixed.

rbac/management/role/serializer.py
randomnetcat: Done.

rbac/management/role/view.py
randomnetcat: I don't see the issue? Declaring a field and assigning at the class-level seems idiomatic for a default value?

rbac/rbac/settings.py
randomnetcat: It is indeed intentional, and I'm not sure where I'd document it?

rbac/management/role/resource_definitions.py
randomnetcat: The order shouldn't ever matter? Given that we also have to handle non-string or otherwise invalid IDs, I'm not sure the cost is worth the benefit here.

rbac/management/inventory_checker/inventory_api_check.py
```diff
@@ -528,3 +529,38 @@ def check_seeded_role_hierarchy(self, hierarchy_tuples: Sequence[RelationTuple],
...
+        )
+
+        results = self._check_inventory_batch(
+            [
+                CheckRequest(
+                    object=resource_reference_pb2.ResourceReference(
+                        resource_id=resource.resource_id,
+                        resource_type=resource.resource_type[1],
+                        reporter=reporter_reference_pb2.ReporterReference(
+                            type=resource.resource_type[0],
```
astrozzc, line 556: can we make use of make_resource_ref and make_subject_ref to simplify these code?

astrozzc (approved):
LGTM overall

lpichler:
@randomnetcat @astrozzc is this necessary to add check this to Kessel Inventory ?
Isn't ticket about existing data then rather about adding checks ?

Currently RBAC is source of truth  for workspaces and tenants until RBAC all objects will be onboarded to Kessel Inventory.

randomnetcat:
Ah, yes, re-reading the ticket, it does look like it's meant to be for validating existing data. Looks like this is really the general case of [RHCLOUD-35482](https://redhat.atlassian.net/browse/RHCLOUD-35482), then?

In any case, we need to have the infrastructure to be able to do these checks for arbitrary resources. Ensuring they work for workspaces seems like a good starting point to ensure that they actually work?

[RHCLOUD-35482]: https://redhat.atlassian.net/browse/RHCLOUD-35482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

astrozzc:
The description of the ticket might be stale. We already have the feature to validate against rbac db. We re-puposed the ticket a bit (see the comment in the ticket) to check against kessel

lpichler:
@randomnetcat @astrozzc - Right, but that check only works for v1 resource definitions right now. 

If we want this to work for any resource, what resources are we actually potentially talking about, and when could Kessel even support that?  - I believe we can scan what is today in resource definitions data in prod.

We need to see if adding this to Kessel is actually faster than just waiting for all orgs to move to v2, because once everyone is on v2, this becomes obsolete anyway.

I am curious if you have insights on this, maybe we can discuss with team

astrozzc:
> @randomnetcat @astrozzc - Right, but that check only works for v1 resource definitions right now.
> 
> If we want this to work for any resource, what resources are we actually potentially talking about, and when could Kessel even support that? - I believe we can scan what is today in resource definitions data in prod.
> 
> We need to see if adding this to Kessel is actually faster than just waiting for all orgs to move to v2, because once everyone is on v2, this becomes obsolete anyway.
> 
> I am curious if you have insights on this, maybe we can discuss with team

Sure, @randomnetcat maybe we can start a thread in the team channel, summarize the issue and discuss over there.
Currently the resources are workspaces and hosts. We don't have the info on the rbac side for hosts which we have to reach out to kessel anyway.
And yea, once on v2, this becomes obsolete, but before one service goes to v2, e.g.,  cost-management, as long as there is v1 orgs which are creating resource definition, we have to add that validation here

randomnetcat:
Per [this discussion](https://redhat-internal.slack.com/archives/C08BQP8PGT1/p1782761246590319), we'll return to this when V1 needs to be able to talk about more resource types.

lpichler:
~@randomnetcat can you fix conflicts ?~

PR #3052, opened 2026-06-17
[RHCLOUD-38972] Add audit logging to user_lookup internal endpoint

dsova06:
## Link(s) to Jira
- RHCLOUD-38972

**What**: Adds audit logging to the `/_private/api/utils/user_lookup/` internal endpoint so we can track who is using it and what they search for.

**Why**: This is a sensitive internal endpoint that exposes a user's groups, roles, and permissions given just a username or email. Today there is no record of who calls it or how often. Audit logging enables security monitoring and usage tracking.

**How**:
- Added a new `READ` action type to the `AuditLog` model (existing actions are all write operations: DELETE, ADD, EDIT, CREATE, REMOVE)
- On successful lookups (HTTP 200): logs the caller identity, the search parameter used, the resolved username, and the looked-up user's tenant
- On not-found lookups (HTTP 404): logs the caller identity and the search parameter (no tenant since the user doesn't exist)
- Validation errors (400) and wrong-method errors (405) are NOT logged since they represent client errors, not meaningful access events
- The audit log helper is wrapped in try/except so a logging failure never changes the endpoint's response

## Local Testing

```bash
# Run the specific test class (18 tests: 15 existing + 3 new)
pipenv run tox -e py312-fast -- tests.internal.test_views.InternalViewsetUserLookupTests
```

New tests cover:
- `test_user_lookup_creates_audit_log_on_success` -- verifies AuditLog entry fields on 200
- `test_user_lookup_creates_audit_log_on_not_found` -- verifies AuditLog entry on 404 (null tenant)
- `test_user_lookup_no_audit_log_on_bad_input` -- verifies no AuditLog created on 400

## Checklist
  - N/A -- internal endpoint, no public API spec
  - Migration 0091 must be applied (adds "read" to AuditLog action choices)
  - N/A
    - Yes -- adds a new choice to a CharField, no schema change
  - No -- internal endpoint, additive change only

- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


dsova06 (commented):
## Self-Review Summary

This PR adds audit logging to the `/_private/api/utils/user_lookup/` internal endpoint (RHCLOUD-38972). A new `READ` action type is added to the `AuditLog` model, and successful lookups (200) and not-found results (404) are recorded. Bad input (400) and wrong method (405) are intentionally not logged. Three new tests cover the logging paths.

## Findings

**[LOW] Correctness -- 500 error paths not audit-logged**
Location: `rbac/internal/views.py:437-438`, `rbac/internal/views.py:452-453`, `rbac/internal/views.py:458-459`
Problem: Three exception paths return HTTP 500 without creating an audit log -- BOP connection failure (line 437), tenant lookup failure (line 452), and principal lookup failure (line 458). While the PR explicitly justifies skipping 400/405 as "client errors, not meaningful access events," 500s represent cases where someone *did* attempt a lookup but the system couldn't complete it. If BOP is intermittently failing, there's no structured audit trail of those attempts.
Suggestion: This is a design choice, not a bug. The regular `logger.error` calls already capture these failures. If the team later wants security monitoring of failed lookup attempts (e.g., repeated probing), consider adding `_log_user_lookup` calls to the 500 paths too. No action required now -- just noting the gap.

**[LOW] Correctness -- Missing test for audit log exception resilience**
Location: `rbac/internal/views.py:499-510`
Problem: The `_log_user_lookup` helper wraps `AuditLog.objects.create()` in try/except so logging failures don't affect the endpoint response. This is good defensive design, but there's no test exercising this path (e.g., mocking `AuditLog.objects.create` to raise and verifying the endpoint still returns 200/404).
Suggestion: Consider adding a test that mocks `AuditLog.objects.create` to raise, calls the endpoint, and asserts the response is still correct. Low priority since the implementation is straightforward.

## Verdict

**Approve** -- No blocking issues. The change is well-scoped, the implementation is clean, and the tests cover the main paths. Both findings above are low-severity observations for future consideration.

## What's Done Well

1. **Defensive logging design** -- Wrapping the audit log creation in try/except ensures a logging failure never changes the endpoint's behavior. This is the right pattern for observability code.

2. **Clean commit separation** -- Splitting the model change (READ action type) from the endpoint integration makes the history easy to bisect and review independently.

3. **Thoughtful scope decisions** -- Explicitly choosing not to log 400/405 responses demonstrates good judgment about what constitutes a "meaningful access event" vs. noise. The PR description clearly explains the reasoning.

dsova06:
Both self-review findings addressed in cf9082b8:

**500 error paths not audit-logged**: Fixed. Added `_log_user_lookup` calls to all three 500 error paths (BOP failure, tenant lookup failure, principal lookup failure). Now all access attempts are tracked regardless of outcome.

**Missing resilience test**: Fixed. Added `test_user_lookup_succeeds_when_audit_log_fails` (mocks `AuditLog.objects.create` to raise, verifies endpoint still returns 200) and `test_user_lookup_creates_audit_log_on_bop_error` (verifies 500 from BOP creates an audit log entry). Updated commit message to reflect 500-path coverage.

rbac/internal/views.py
```diff
@@ -494,9 +501,25 @@ def user_lookup(request):
...
+def _log_user_lookup(caller, description, tenant=None):
+    """Create an audit log entry for a user lookup request."""
+    try:
+        AuditLog.objects.create(
+            principal_username=caller,
+            description=description[:255],
+            resource_type=AuditLog.USER,
+            action=AuditLog.READ,
+            tenant=tenant,
+        )
```
randomnetcat, line 519: We should likely log the principal in `resource_id` or `resource_uuid` if it was found?

rbac/internal/views.py
```diff
@@ -494,9 +501,25 @@ def user_lookup(request):
...
     return HttpResponse(json.dumps(result, cls=DjangoJSONEncoder), content_type="application/json", status=200)
 
 
+def _log_user_lookup(caller, description, tenant=None):
+    """Create an audit log entry for a user lookup request."""
+    try:
+        AuditLog.objects.create(
+            principal_username=caller,
+            description=description[:255],
+            resource_type=AuditLog.USER,
```
randomnetcat, line 516: Is `USER` the correct type here? I see that in [this PR](https://github.com/project-kessel/insights-rbac/pull/3052) you propose adding a new `PRINCIPAL` type.

rbac/internal/views.py
```diff
@@ -494,9 +501,25 @@ def user_lookup(request):
...
+            action=AuditLog.READ,
+            tenant=tenant,
```
randomnetcat: Setting a tenant here means that we will tell end users whether this internal tool was used; is this desirable?

rbac/internal/views.py
dsova06: Fixed: added `principal` parameter to `_log_user_lookup` and set `resource_uuid=getattr(principal, "uuid", None)` so successful lookups link to the Principal record.

rbac/internal/views.py
dsova06: Acknowledged -- USER is the correct type. RESOURCE_CHOICES already defines USER but has no PRINCIPAL type. The link in the comment points back to this same PR (likely a copy-paste artifact), so there is no other PR proposing a PRINCIPAL type.

rbac/internal/views.py
dsova06: Fixed: removed tenant from all audit log entries. They are now only accessible via direct DB queries (internal monitoring), not through the audit log API which filters by tenant.

rbac/internal/views.py
randomnetcat: Sorry, [this PR](https://github.com/project-kessel/insights-rbac/pull/3097).

dsova06: Thanks for the link. Since user_lookup queries by username and returns user/account data rather than the Principal ORM record, USER feels like the right fit. Happy to switch to PRINCIPAL if you prefer.

rbac/internal/views.py
```diff
@@ -427,6 +428,8 @@ def user_lookup(request):
 
     username = request.GET.get("username")
     email = request.GET.get("email")
```
lpichler, line 431: **medium**: User-supplied query parameters (username, email) are interpolated into audit log descriptions without sanitization, violating the project's clean_query_param() requirement

search_param = f"username='{username}'" if username else f"email='{email}'" — raw request.GET values are embedded into audit log descriptions via f-string formatting. The project's security guidelines mandate using clean_query_param() for all user-supplied query parameters to reject NUL bytes. Values containing HTML/script tags, NUL chars, or misleading content would pass validation and be stored in the audit log, which is exposed via AuditLogSerializer API.

rbac/internal/views.py
```diff
@@ -494,9 +501,25 @@ def user_lookup(request):
 
     result["groups"] = user_groups
 
+    _log_user_lookup(caller, f"User lookup: found '{username}' ({search_param})", principal=principal)
+
     return HttpResponse(json.dumps(result, cls=DjangoJSONEncoder), content_type="application/json", status=200)
```
lpichler, line 508: **medium**: Audit log entries created without tenant association, making them invisible through standard tenant-scoped API and unable to benefit from composite indexes

The _log_user_lookup function creates AuditLog records with tenant=None. Every other audit log creation sets self.tenant_id = self.get_tenant_id(request). The AuditLogViewSet filters by request.tenant, and all database indexes on AuditLog are prefixed with tenant. Tenant-less records will not appear when querying audit logs through the standard API and won't benefit from composite indexes (tenant, created), (tenant, resource_type), (tenant, action).

rbac/internal/views.py
```diff
@@ -65,6 +65,7 @@
     relation_tuples_pb2,
     relation_tuples_pb2_grpc,
 )
+from management.audit_log.model import AuditLog
```
lpichler, line 68: **medium**: Bare 'except Exception' in _log_user_lookup swallows all exceptions without categorizing retriable vs non-retriable failures

except Exception: catches all exceptions including OperationalError (transient DB issues), IntegrityError (data issues), and programming errors. The project's integration guidelines distinguish retriable from non-retriable errors. Without categorization, there's no way to add retry logic or meaningful metrics for audit log failure modes.

rbac/internal/views.py
```diff
@@ -65,6 +65,7 @@
     relation_tuples_pb2,
```
lpichler, line 65: **low**: Silent truncation of audit log description to 255 characters may hide important diagnostic or security-relevant information

description=description[:255] silently truncates without any indication. For error messages or long usernames, critical context could be lost. Particularly problematic for error cases where full error details might contain important security information.

rbac/internal/views.py
```diff
@@ -453,12 +458,14 @@ def user_lookup(request):
...
         return handle_error(f"Internal error - failed to query rbac for tenant with org_id: '{user_org_id}'", 500)
 
     try:
         principal = get_principal(username, request, verify_principal=False, from_query=False, user_tenant=user_tenant)
     except Exception as err:
         logger.error(f"error querying for principal with username: '{username}' in rbac, err: {err}")
+        _log_user_lookup(caller, f"User lookup: error resolving principal ({search_param})")
         return handle_error(f"Internal error - failed to query rbac for user: '{username}'", 500)
 
     groups = groups_for_principal(principal, user_tenant, is_org_admin=user["is_org_admin"])
```
lpichler, line 472: **high**: Pre-existing N+1 query pattern in user_lookup group/role/access iteration loop (not introduced by this PR)

Lines 468-486: Nested loops iterate over groups→roles→Access.objects.filter(role=role)→access.permission, creating N*M*P database queries. For a user with multiple groups, roles, and permissions, this generates excessive DB queries on every user_lookup call.

rbac/internal/views.py
lpichler, line 508: **medium**: AuditLog.objects.create() bypasses the established model-method pattern used everywhere else in the codebase

Every other call site uses instance methods (log_create, log_edit, log_delete, log_v2, log_create_from_object, etc.). This creates a second, inconsistent API. If AuditLog gains new cross-cutting concerns, this call site will be missed.

rbac/internal/views.py
lpichler, line 65: **medium**: Audit logging for BOP errors happens after error response, creating timing window where log could be lost

_log_user_lookup is called after exception handling. If the process crashes between the two operations, the audit trail is incomplete.

lpichler, line 65: **low**: No idempotency protection for audit log entries — duplicate calls create duplicate records

AuditLog.objects.create() creates a new entry on every call. If the caller retries the user_lookup request, multiple audit log entries will be created for the same logical operation.

lpichler, line 65: **low**: No timeout configured for audit log database write operation

AuditLog.objects.create() has no timeout. A slow database write could delay the response even though the try/except prevents failures from breaking the endpoint.

rbac/management/migrations/0091_add_read_action_to_audit_log.py
```diff
@@ -0,0 +1,28 @@
+# Generated by Django 5.2.12 on 2026-06-17 12:58
```
lpichler, line 1: **low**: Django version inconsistency across migrations in the same PR

Migration 0091 was generated with Django 5.2.12, while migration 0092 was generated with Django 6.0.6. Different Django versions in the same PR signals environment inconsistency.

lpichler:
## Code Review Findings

**File:** `rbac/internal/views.py` (line 37)

**low**: Overly broad exception handlers for tenant and principal lookup may swallow distinct Django ORM errors

Lines 35-38 and 42-45 catch 'except Exception' for Tenant.objects.get() and get_principal(). This catches DoesNotExist (expected), MultipleObjectsReturned (data integrity), OperationalError (DB connection) — each warranting different handling.



---

**File:** `tests/internal/test_views.py` (line 168)

**medium**: Missing test for email-based lookup audit logging

All audit log tests use username parameter, but the implementation supports both username and email. The search_param construction logic differs for email vs username, but no test verifies audit log format when searching by email.



---

**File:** `rbac/internal/views.py` (line 18)

**medium**: String formatting with None values creates misleading audit log messages like "email='None'"

search_param = f"username='{username}'" if username else f"email='{email}'". When email lookup is used but email is None, this produces confusing messages.



---

**File:** `tests/internal/test_views.py` (line 206)

**medium**: Missing test for tenant lookup failure audit logging

The implementation adds audit logging for tenant lookup errors but existing test_user_lookup_tenant_does_not_exist_in_rbac does not verify that an audit log is created.



---

**File:** `tests/internal/test_views.py` (line 206)

**medium**: Missing test for principal lookup failure audit logging

The implementation adds audit logging for principal lookup errors but existing test_user_lookup_get_principal_raises_exception does not verify that an audit log is created.



---

**File:** `tests/internal/test_views.py` (line 228)

**low**: Missing test for description truncation at 255 characters

The implementation truncates description[:255] but no test verifies this works correctly with very long usernames or search params.



---

**File:** `rbac/internal/views.py` (line 18)

**low**: Nested getattr calls reduce readability

caller = getattr(getattr(request, 'user', None), 'username', 'internal-service') — nested getattr pattern is harder to read.



---

**File:** `tests/internal/test_views.py` (line 168)

**low**: Missing test for caller identity verification with different user types

Tests verify log.principal_username exists but never verify the actual value or test with service accounts to verify the fallback to 'internal-service'.



---

**File:** `tests/internal/test_views.py` (line 246)

**low**: Test assertion on mock may validate wrong behavior — questionable mock setup

test_user_lookup_creates_audit_log_on_bop_error sets mock_proxy.side_effect after @patch returns a dict, which may mean the mock setup is incorrect.



---

**File:** `rbac/internal/views.py` (line 29)

**low**: Overly broad exception handler may mask BOP API failures that need distinct handling

Line 28-30 catches 'except Exception' for get_user_from_bop(). May catch non-transient failures (auth errors, config errors) that should be handled differently from transient network issues.



---

**File:** `tests/internal/test_views.py` (line 206)

**low**: Missing test for 405 method not allowed without audit log

PR description says 'wrong-method errors (405) are NOT logged' but no test verifies POST/PUT/DELETE requests don't create audit logs.



---

**File:** `rbac/internal/views.py` (line 18)

**low**: Potential information disclosure through audit logging of internal caller identity

If request.user exists but username is None/empty, the fallback to 'internal-service' could mask the actual caller.

dsova06:
Thanks for the review. These findings appear to be from an automated review tool -- addressing each:

**Fixed (3):**
- Missing email-based lookup test -- added `test_user_lookup_creates_audit_log_on_email_search`
- Missing tenant failure test -- added `test_user_lookup_creates_audit_log_on_tenant_error`
- Missing principal failure test -- added `test_user_lookup_creates_audit_log_on_principal_error`

**Acknowledged (remaining):**
- Overly broad exception handlers (tenant/principal) -- pre-existing code, not introduced by this PR
- String formatting with None -- `validate_user_lookup_input` rejects before reaching audit log code, so email is never None in that path
- Description truncation test -- the model field is `max_length=255`, truncation is a one-liner
- Nested getattr readability -- idiomatic Python for safe nested attribute access
- Caller identity test -- tests already assert on `log.principal_username`
- Mock setup concern -- the mock setup is correct (side_effect on the mock returned by @patch)
- 405 test -- handled by the method check at line 426 before any audit code runs
- Info disclosure -- fallback to `internal-service` is intentional safe default

rbac/internal/views.py
dsova06: Acknowledged -- this is an internal endpoint (PSK-authed), the params go to BOP for lookup, and these audit entries have no tenant so they are not exposed via the AuditLog API. The NUL byte risk from internal callers is minimal.

rbac/internal/views.py
dsova06: Acknowledged -- this is intentional per randomnetcat's review. Setting tenant would expose internal tool usage to end users via the audit log API. These entries are for internal monitoring only.

rbac/internal/views.py
dsova06: Acknowledged -- intentional defensive design. Audit logging must never affect endpoint behavior. The endpoint response is identical regardless of exception type, so categorization adds complexity with no benefit.

rbac/internal/views.py
dsova06: Acknowledged -- the model field is max_length=255. Truncation is required. In practice, descriptions are well under the limit.

rbac/internal/views.py
dsova06: Acknowledged -- pre-existing issue, not introduced by this PR. Tracked separately.

rbac/internal/views.py
dsova06: Acknowledged -- the existing model methods (log_create, log_v2, etc.) all require a DRF request object and set tenant_id. This is an internal endpoint with a plain Django HttpRequest, and we deliberately omit tenant. objects.create() is the right fit.

rbac/internal/views.py
dsova06: Acknowledged -- not actionable at the application level. If the process crashes between response and log, no code can prevent that.

dsova06: Acknowledged -- each call is a separate access event. Deduplication would be incorrect for audit logs.

dsova06: Acknowledged -- DB write timeouts are configured at the connection pool level, not per-query in Django ORM.

rbac/management/migrations/0091_add_read_action_to_audit_log.py
dsova06: Acknowledged -- cosmetic. Django auto-stamps the version when generating migrations. The two were generated at different times during development. Both are functionally correct.

lpichler (approved):

vbelchio:
QE: Future Work

PR #3090, opened 2026-06-22
feat(dr): add offset and last-minutes modes to DR corrective tools

lpichler:
## Summary

Adds two new modes to both disaster recovery corrective tools (HBI-RBAC workspace recovery and Kessel-RBAC reconciliation), giving operators more flexibility during DR exercises:

- **Offset mode**: Specify `start_offset` and optional `end_offset` to read Kafka events by offset range instead of timestamp
- **Last-minutes mode**: Specify `last_minutes` to process events from the last N minutes (convenience shorthand for timestamp mode)

All three modes (timestamp, offset, last-minutes) are mutually exclusive with proper validation at every layer.

### Changes

**New Kafka readers** (`core/kafka_dr.py`, `management/disaster_recovery/kafka_reader.py`):
- `read_events_by_offset()` — seeks to `start_offset` on each partition, reads until `end_offset` (exclusive) or end of partition
- Per-partition tracking via `finished_partitions` set with `consumer.pause()` to prevent infinite loops

**Service layer** (`management/disaster_recovery/service.py`):
- `reconcile()` accepts optional `start_offset`/`end_offset` (mutually exclusive with timestamp)
- Response key renamed from `time_window` to `window` (polymorphic: holds timestamp or offset range info)

**Task layer** (`management/tasks.py`):
- Both Celery tasks accept `start_offset`/`end_offset` params
- `recover_workspace_events_in_worker` validates mutual exclusivity at the task level

**API endpoints** (`internal/views.py`):
- Both DR endpoints (`/_private/api/disaster_recovery/workspaces/` and `/_private/api/disaster_recovery/reconcile/`) support all three modes
- Comprehensive input validation: type checks, range checks, boolean guards, mode exclusivity

### API Usage

```bash
# Timestamp mode (existing)
POST /_private/api/disaster_recovery/workspaces/
{"restore_timestamp": "2026-05-28T10:00:00Z", "buffer_minutes": 5}

# Offset mode (new)
POST /_private/api/disaster_recovery/workspaces/
{"start_offset": 100, "end_offset": 200}

# Last-minutes mode (new)
POST /_private/api/disaster_recovery/workspaces/
{"last_minutes": 30}

# Same three modes for Kessel-RBAC reconciliation
POST /_private/api/disaster_recovery/reconcile/
{"start_offset": 100, "end_offset": 200}
{"last_minutes": 30}
```

## Test plan

- [x] 90 DR-specific tests pass (offset mode, last-minutes mode, validation edge cases, task validation)
- [x] Full test suite passes (3905 tests)
- [x] Linter passes
- [x] Multi-partition offset tracking verified correct (per-partition set + consumer.pause)
- [x] Task mutual exclusivity validated (both tasks reject conflicting params)
- [x] Boolean-as-integer edge cases covered (e.g., `start_offset: true`)


lpichler:
@sourcery-ai review

dsova06 (changes requested):
### Findings

**[HIGH] Correctness -- kafka_reader.py doesn't pause skipped partitions in offset mode**

`rbac/management/disaster_recovery/kafka_reader.py` ~line 290

When `start_offset >= partition_end`, the partition is skipped (not added to `active_partitions`) but remains assigned and NOT paused. Since `consumer.seek()` is never called for these partitions, `consumer.poll()` reads from them starting at `auto_offset_reset=earliest` (offset 0). The inner poll loop only checks `if tp in finished_partitions`, not whether `tp` is in `active_partitions`, so those stale messages get processed as valid events. The sibling function in `core/kafka_dr.py` handles this correctly by calling `consumer.pause([tp])`.

Suggestion -- add a pause call for skipped partitions, matching `core/kafka_dr.py`:
```python
for tp in topic_partitions:
    partition_end = end_offsets_map.get(tp, 0)
    if start_offset >= partition_end:
        consumer.pause([tp])   # <-- add this
        continue
    consumer.seek(tp, start_offset)
    active_partitions.append(tp)
```

**[MEDIUM] Code Quality -- Mode detection/validation duplicated between two view functions**

`rbac/internal/views.py` ~lines 2736-2770 and ~lines 2862-2900

The mode detection logic (`use_offsets`/`use_timestamp`/`use_last_minutes`, `mode_count` check, mutual exclusivity error, offset validation with boolean guards, last_minutes validation) is nearly identical between `recover_workspace_events` and `disaster_recovery_reconcile`. About 30 lines of copy-paste. If validation rules change (e.g. adding a fourth mode), both functions need identical updates. Consider extracting a helper like `_validate_dr_mode(body)` that returns a typed result (which mode, validated params) or raises. Not blocking, but worth doing since these two endpoints will evolve together.

**[LOW] Docs -- OpenAPI spec paths missing `/_private` prefix**

`rbac/internal/specs/openapi.json` new paths

The spec defines paths as `/api/disaster_recovery/workspaces/` and `/api/disaster_recovery/reconcile/`, but the actual routes are `/_private/api/disaster_recovery/...`. Other internal endpoints in the same spec use `/_private/api/...`.

### Verdict

**Request changes** -- the kafka_reader.py partition-pause bug is the one blocker. The skipped partitions silently read from offset 0, which would return incorrect events during a real DR exercise. One-line fix. Everything else is solid.

### What's Done Well

1. **The `finished_partitions` set + `consumer.pause()` pattern** (in `core/kafka_dr.py`) is the right approach for multi-partition offset boundaries -- much better than the original `past_range` boolean that was fixed in commit 5.
2. **`last_minutes` mode is a clean UX addition** -- converting to timestamp at the view layer means zero changes to the task and service layers, keeping the abstraction boundaries clean.
3. **Thorough test coverage** -- edge cases like boolean-as-integer, negative values, zero, non-integer types, and all mutual exclusivity combinations are covered for both endpoints.

dsova06 (approved):
All three review findings addressed cleanly in the latest commit. The kafka_reader partition-pause fix, validation helper extraction, and OpenAPI path prefix are all good. Nice work.

PR #3091, opened 2026-06-22
refactor(dr): rename DR settings to reflect their scope

lpichler:
## Summary
- Rename `DR_RECONCILE_ENABLED` → `DR_RELATIONS_RECONCILE_ENABLED` (gates Kessel Relations/SpiceDB tuple reconciliation)
- Rename `DR_RECOVERY_ENABLED` → `DR_WORKSPACE_RECONCILE_ENABLED` (gates RBAC workspace reconciliation)
- Add `DR_WORKSPACE_RECONCILE_ENABLED` to ClowdApp template (was missing — bonfire `-p` flag was silently ignored)

The old names were ambiguous — both sounded like generic DR flags. The new names clarify which subsystem each flag controls, and both use `RECONCILE` for consistency.

## Test plan
- [ ] Verify `DR_RELATIONS_RECONCILE_ENABLED=True` enables the `/disaster_recovery/reconcile/` endpoint
- [ ] Verify `DR_WORKSPACE_RECONCILE_ENABLED=True` enables the `/disaster_recovery/workspaces/` endpoint
- [ ] Verify both default to `False`
- [ ] Run existing DR test suites: `tox -r -e py312-fast -- tests.internal.test_dr_views` and `tests.internal.test_dr_recovery_views`


dsova06 (approved):
Clean, well-scoped rename with a real bug fix. All references to the old names are covered -- I grepped the full codebase and every occurrence is addressed in the diff. CI is fully green. **Approve.**

### Findings

**[MEDIUM] Consistency -- Error message style differs between the two endpoints**

`views.py:2756` vs `views.py:2822`

The workspace endpoint includes the flag name in the error message (`"DR recovery is disabled (DR_WORKSPACE_RECONCILE_ENABLED=False)"`), but the relations endpoint just says `"Disaster recovery reconciliation is not enabled"` without naming the flag. Same inconsistency exists at the task layer (`tasks.py:578` vs `tasks.py:654`). Including the flag name is the better pattern -- it tells operators exactly what to set. Pre-existing issue, but since the PR is already touching these exact lines, this is a good opportunity to align them.

Suggestion -- update the relations endpoint/task messages:
```python
# views.py ~line 2822
return JsonResponse({"error": "DR reconciliation is disabled (DR_RELATIONS_RECONCILE_ENABLED=False)"}, status=403)

# tasks.py ~line 654
return {"message": "DR reconciliation disabled (DR_RELATIONS_RECONCILE_ENABLED=False)"}
```

**[LOW] Consistency -- Error response key differs between endpoints**

`views.py:2756` (`"detail"`) vs `views.py:2822` (`"error"`)

The workspace endpoint uses `{"detail": "..."}` but the relations endpoint uses `{"error": "..."}`. Both are internal PSK endpoints so not externally breaking, but inconsistent response shapes make scripting against them harder. Also pre-existing, non-blocking.

### What's Done Well

1. **The ClowdApp fix is the real win** -- `bonfire -p` silently ignoring the flag because it wasn't in the template is exactly the kind of bug that wastes hours of debugging. Good catch.
2. **Commit progression is clean** -- rename first, add to template second, then consistency pass. Each commit is self-contained and the messages explain the "why."
3. **Tests updated in lockstep** -- every `@override_settings` and assertion string is updated to match, including the docstrings.

PR #3103, opened 2026-06-24
fix(parity): use ReadTuples for wildcard subjects in custom role permission check

dsova06:
## Link(s) to Jira
- [RHCLOUD-48679](https://redhat.atlassian.net/browse/RHCLOUD-48679)

**What**: The Kessel parity check task (`run_kessel_parity_checks_in_worker`) crashes when checking custom role permissions because all custom role permission tuples use wildcard subjects (`rbac/principal:*`), and SpiceDB's Check API rejects wildcards with `INVALID_ARGUMENT: cannot perform check on wildcard subject`.

**Why**: The custom role parity sub-check fails for any org that has custom roles (e.g., tenant 20228402 on stage). The other 4 sub-checks (workspace hierarchy, seeded roles, bootstrap, group-principal) are unaffected since they only use concrete subjects.

**How**: `CustomRolePermissionChecker` now partitions permission tuples into wildcard and concrete subjects. Wildcard tuples are verified via the Relations API `ReadTuples` (which queries stored relationships directly and supports wildcards), while concrete tuples still use the Inventory API Check path. This uses the same ReadTuples pattern already established in `rbac/internal/views.py` and the same JWT auth already initialized in the checker module.

### Call chain (before fix)
1. `tasks.py:402` builds permission tuples via `CustomRoleV2._permission_tuple()`
2. `migration_tool/models.py:257` creates tuple with subject ID `*` (wildcard, by design)
3. `inventory_api_check.py:70` copies `*` into `CheckRequest.subject.resource_id`
4. SpiceDB rejects: `INVALID_ARGUMENT: cannot perform check on wildcard subject`

### Call chain (after fix)
1-2 unchanged
3. Tuples partitioned: wildcard subjects go to `_check_wildcard_tuples_via_read()` which uses `ReadTuplesRequest` on the Relations API
4. ReadTuples returns matching relationships (or empty if missing) -- no crash

## Local Testing

```bash
# Run the custom role permission checker tests (5 tests)
pipenv run tox -e py312-fast -- tests.management.inventory_checker.test_custom_role_permission_checker

# Run all inventory checker tests (45 tests)
pipenv run tox -e py312-fast -- tests.management.inventory_checker

# Lint
pipenv run tox -e lint
```

For stage validation: run the parity check on tenant 20228402 (the tenant that triggered the original crash). The custom role sub-check should now complete without crashing.

## Checklist
- [ ] if API spec changes are required, is the spec updated? N/A - no API changes
- [ ] are there any pre/post merge actions required? No
- [ ] does this require migration changes? No
- [ ] is there known, direct impact to dependent teams/components? No

- [x] Input Validation - wildcard detection uses exact string match (`== "*"`)
- [x] Output Encoding - N/A, no user-facing output changes
- [x] Authentication and Password Management - uses existing JWT auth pattern
- [x] Session Management - N/A
- [x] Access Control - N/A, parity checker is a background task
- [x] Cryptographic Practices - N/A
- [x] Error Handling and Logging - gRPC errors propagate to existing task-level try/except
- [x] Data Protection - no new data exposure
- [x] Communication Security - uses existing TLS gRPC channel (`create_client_channel_relation`)
- [x] System Configuration - uses existing `RELATION_API_SERVER` setting
- [x] Database Security - N/A
- [x] File Management - N/A
- [x] Memory Management - N/A
- [x] General Coding Practices - follows existing checker patterns

[RHCLOUD-48679]: https://redhat.atlassian.net/browse/RHCLOUD-48679?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rbac/management/inventory_checker/inventory_api_check.py
+++ b/rbac/management/inventory_checker/inventory_api_check.py
@@ -31,6 +31,7 @@ from kessel.inventory.v1beta2 import (
     subject_reference_pb2,
 )
 from kessel.inventory.v1beta2.check_request_pb2 import CheckRequest
+from kessel.relations.v1beta1 import relation_tuples_pb2, relation_tuples_pb2_grpc
 from management.cache import JWTCache
 from management.group.platform import DefaultGroupNotAvailableError, GlobalPolicyIdService
 from management.permission.scope_service import ImplicitResourceService, Scope
@@ -38,7 +39,7 @@ from management.relation_replicator.types import RelationTuple
 from management.role.platform import admin_platform_parent_scope_for_seeded_system_role, platform_v2_role_uuid_for
 from management.role.relations import role_child_relationship
 from management.tenant_mapping.model import DefaultAccessType, TenantMapping
-from management.utils import create_client_channel_inventory
+from management.utils import create_client_channel_inventory, create_client_channel_relation
 from migration_tool.utils import create_relationship
 
 from api.models import Tenant
@@ -458,10 +459,55 @@ class CrossAccountRequestInventoryChecker(InventoryApiBaseChecker):
 class CustomRolePermissionChecker(InventoryApiBaseChecker):
     """Subclass to check custom role permission relations are correct on inventory api."""
 
+    def _check_permission_tuples_via_read(self, tuples: Sequence[RelationTuple], role_uuid: str) -> bool:
+        """Verify permission tuples exist using the Relations API ReadTuples.
+
+        All custom role permission tuples use wildcard subjects (rbac/principal:*) by SpiceDB
+        schema design. The Check API rejects wildcards, so we use ReadTuples which queries
+        stored relationships directly.
+
+        Uses JWT metadata for auth because the Relations API channel does not bundle
+        call credentials (unlike the Inventory API channel used by check_inventory_core).
+        """
+        token = jwt_manager.get_jwt_from_redis()
+        metadata = [("authorization", f"Bearer {token}")] if token else []
+        all_present = True
+
+        with create_client_channel_relation(settings.RELATION_API_SERVER) as channel:
+            stub = relation_tuples_pb2_grpc.KesselTupleServiceStub(channel)
+
+            for t in tuples:
+                request = relation_tuples_pb2.ReadTuplesRequest(
+                    filter=relation_tuples_pb2.RelationTupleFilter(
+                        resource_namespace=t.resource.type.namespace,
+                        resource_type=t.resource.type.name,
+                        resource_id=t.resource.id,
+                        relation=t.relation,
+                        subject_filter=relation_tuples_pb2.SubjectFilter(
+                            subject_namespace=t.subject.subject.type.namespace,
+                            subject_type=t.subject.subject.type.name,
+                            subject_id=t.subject.subject.id,
+                        ),
+                    )
+                )
+                responses = list(stub.ReadTuples(request, metadata=metadata))
+                if not responses:
+                    logger.warning(
+                        f"CustomRole: {role_uuid} missing relation "
+                        f"{t.resource.type.name}:{t.resource.id}#{t.relation}"
+                    )
+                    all_present = False
+
+        return all_present
+
     def check_custom_role_permissions(self, permission_tuples: Sequence[RelationTuple], role_uuid: str) -> bool:
-        """Core logic to check custom role permission relations on inventory api.
+        """Core logic to check custom role permission relations via the Relations API.
 
         Each permission tuple represents: rbac/role:<uuid>#<permission>@rbac/principal:*
+
+        Uses ReadTuples instead of the Inventory Check API because all custom role
+        permission subjects are wildcards (principal:*) by SpiceDB schema design,
+        and the Check API does not support wildcard subjects.
 
         Args:
             permission_tuples: List of RelationTuple objects from CustomRoleV2._permission_tuple()
@@ -474,9 +520,7 @@ class CustomRolePermissionChecker(InventoryApiBaseChecker):
             logger.debug(f"CustomRole: {role_uuid} has no permissions, skipping check")
             return True
 
-        check_requests = [relation_tuple_to_check_request(tuple_obj) for tuple_obj in permission_tuples]
-
-        permission_check = self.check_inventory_core(check_requests)
+        permission_check = self._check_permission_tuples_via_read(permission_tuples, role_uuid)
         if not permission_check:
             logger.warning(f"CustomRole: {role_uuid} does not have the expected permission relations in inventory.")
         else:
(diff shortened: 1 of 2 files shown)


dsova06 (commented):
## PR Summary

PR #3103 (2 commits, 2 files, +120/-81 lines) fixes RHCLOUD-48679 where the Kessel parity check task crashed on custom role permissions because all custom role permission tuples use wildcard subjects (`rbac/principal:*`), and SpiceDB's Check API rejects wildcards with `INVALID_ARGUMENT`. The fix partitions permission tuples in `CustomRolePermissionChecker`: wildcard subjects are verified via the Relations API `ReadTuples` (which can query stored wildcard relationships directly), while concrete subjects continue through the existing Inventory API Check path.

## Findings

```
[MEDIUM] Test Coverage -- No test coverage for mixed wildcard + concrete tuple path
Location: tests/management/inventory_checker/test_custom_role_permission_checker.py
Problem: check_custom_role_permissions (inventory_api_check.py:529-541) has explicit branching
  for wildcard_tuples and concrete_tuples lists, including a check_inventory_core() call
  for the concrete path. However, all tests only exercise the wildcard path because
  CustomRoleV2._permission_tuple() always produces wildcard subjects. The concrete path
  (lines 537-539) and the AND combination of both results (line 541) are never tested.
  Today all custom role permission tuples use wildcard subjects, so the concrete path is
  effectively dead -- but if it were ever exercised, bugs in the AND logic or in the
  concrete branch would be invisible.
Suggestion: Consider adding a test that manually constructs a mixed list (some wildcard,
  some concrete) and verifies both paths are exercised and the AND logic works correctly.
  Alternatively, if the concrete path is genuinely impossible for custom role permissions,
  a comment noting this (and an assertion or early-return that skips it) would make the intent
  clearer.
```

```
[LOW] Diagnostics -- Early return on first missing wildcard tuple
Location: rbac/management/inventory_checker/inventory_api_check.py:501-506
Problem: _check_wildcard_tuples_via_read returns False on the first missing tuple. For a
  parity check (a diagnostic tool), reporting ALL missing tuples is more useful for operators
  investigating discrepancies -- they'd see the full scope of the problem in one run rather
  than discovering issues one at a time across multiple executions.
Suggestion: This is consistent with the existing pattern in check_inventory_core (which also
  short-circuits via all()), so it's not a real problem. Just noting that for diagnostic use
  cases, collecting all misses and logging them before returning False could be more helpful.
  Not blocking.
```

## Verdict

**Approve** -- no blockers. The fix is well-reasoned, correctly addresses the SpiceDB wildcard limitation, follows established patterns (ReadTuples usage from `internal/views.py`, channel reuse inside `with` block, JWT auth pattern), and has solid test coverage for the primary path. The MEDIUM finding about the untested concrete path is worth considering but not blocking since that path is currently unreachable in practice.

## What's Done Well

1. **Correct API choice**: Using ReadTuples instead of Check for wildcard subjects is the right architectural call. The PR description clearly explains why Check rejects wildcards and how ReadTuples queries stored relationships directly.

2. **Channel reuse**: Opening one gRPC channel and iterating inside the `with` block follows the architecture guidance. The existing bug in `internal/views.py` (where ReadTuples is called OUTSIDE the channel context manager) is not repeated here.

3. **Clean commit structure**: Fix and tests separated into distinct commits with clear commit messages. Tests cover the important scenarios (all exist, one missing, all missing, single permission, empty list).

rbac/management/inventory_checker/inventory_api_check.py
```diff
@@ -474,9 +527,19 @@ def check_custom_role_permissions(self, permission_tuples: Sequence[RelationTupl
...
+        concrete_tuples = [t for t in permission_tuples if not self._is_wildcard_subject(t)]
+
+        wildcard_passed = True
+        if wildcard_tuples:
+            wildcard_passed = self._check_wildcard_tuples_via_read(wildcard_tuples, role_uuid)
+
+        concrete_passed = True
+        if concrete_tuples:
+            check_requests = [relation_tuple_to_check_request(t) for t in concrete_tuples]
+            concrete_passed = self.check_inventory_core(check_requests)
```
astrozzc: since the SpiceDB schema design choice for how permissions are modeled on roles, it is always having wild card, something like this: rbac/role:<uuid>#inventory_hosts_read@rbac/principal:*.
Maybe we can add a comment and get rid of the concrete part, directly `permission_check = self._check_wildcard_tuples_via_read(permission_tuples, role_uuid)****`

rbac/management/inventory_checker/inventory_api_check.py
```diff
@@ -458,11 +459,63 @@ def check_cross_account_request(self, tuples: Sequence[RelationTuple], request_i
...
+        relationships directly and can match wildcards.
+
+        Args:
+            tuples: Permission tuples with wildcard subjects
+            role_uuid: UUID of the custom role (for logging)
+
+        Returns:
+            True if all wildcard relationships exist in Kessel
+        """
+        token = jwt_manager.get_jwt_from_redis()
```
astrozzc, line 472: Can we reply on check_inventory_core like the other functions instead of its own jwt logic

rbac/management/inventory_checker/inventory_api_check.py
```diff
@@ -458,11 +459,63 @@ def check_cross_account_request(self, tuples: Sequence[RelationTuple], request_i
...
+        metadata = [("authorization", f"Bearer {token}")] if token else []
+        all_present = True
+
+        with create_client_channel_relation(settings.RELATION_API_SERVER) as channel:
+            stub = relation_tuples_pb2_grpc.KesselTupleServiceStub(channel)
+
+            for t in tuples:
```
astrozzc, line 479: one ReadTuples call per tuple in a loop. For roles with many permissions, this could be slow. The RelationsApiReplicator.read_tuples method already exists and could potentially be leveraged, or a batch approach could be considered.

astrozzc (approved):
Overall LGTM, feel free to fix and merge

rbac/management/inventory_checker/inventory_api_check.py
dsova06: Fixed: removed the concrete path and partition logic entirely. All tuples now route through ReadTuples directly, since permission subjects are always wildcards by SpiceDB schema design.

rbac/management/inventory_checker/inventory_api_check.py
dsova06: Acknowledged -- check_inventory_core uses the Inventory API channel which bundles call credentials via composite_channel_credentials. ReadTuples uses the Relations API channel which requires JWT in metadata instead. Different APIs, different auth mechanisms. Added a docstring note explaining why.

rbac/management/inventory_checker/inventory_api_check.py
dsova06: Acknowledged -- per-tuple ReadTuples is consistent with the per-request Check pattern in check_inventory_core. Custom roles typically have <20 permissions, so acceptable for the parity check use case. A single gRPC channel is reused for all tuples within one role.

PR #3108, opened 2026-06-25
[RHCLOUD-48826] fix(dr): check TenantMapping for bootstrap virtual group/role_binding resources

lpichler:
## Summary

- Fix DR reconciler incorrectly generating REMOVE corrective actions for bootstrap_tenant event tuples
- The `check_resources_exist` function only queried `Group`/`RoleBinding` Django model tables, but `bootstrap_tenant` creates relation tuples referencing UUIDs stored in `TenantMapping` without creating actual model instances — these are "virtual" resources
- Add a `TenantMapping` fallback: after the primary model check, any group or role_binding UUIDs not found are looked up across the 8 `TenantMapping` UUID fields (2 group + 6 role_binding)

## Root Cause

During `bootstrap_tenant`, `TenantMapping` is created with auto-generated UUIDs for default groups (`default_group_uuid`, `default_admin_group_uuid`) and 6 role binding UUIDs. Relation tuples are emitted referencing these UUIDs, but **no actual `Group` or `RoleBinding` Django model objects are created**. The DR reconciler's existence check queried only the model tables, so it reported these resources as non-existent, triggering the truth table path `relations_to_add + not exists → REMOVE` — which would delete valid relations from SpiceDB.

## Test plan

- [x] New `TenantMappingFallbackTest` class with 10 test cases covering:
  - Group found via `default_group_uuid` and `default_admin_group_uuid`
  - Role binding found via all 6 TenantMapping role binding fields
  - UUIDs not in TenantMapping still return `False`
  - Mixed real model instances and virtual TenantMapping resources
  - Full bootstrap scenario with all virtual resources
- [x] All 55 existing DR tests pass


dsova06 (approved):
## Summary

Clean, well-scoped fix for the DR reconciler incorrectly generating REMOVE corrective actions for bootstrap virtual resources. The root cause -- `check_resources_exist` only querying `Group`/`RoleBinding` model tables while `bootstrap_tenant` stores UUIDs exclusively in `TenantMapping` -- is clearly identified and correctly addressed.

## Findings

No blocking issues.

**[LOW] Test coverage note** (`test_resource_checker.py`): Individual field tests cover 4 of 6 role binding fields. The remaining 2 (`root_scope_default_admin`, `tenant_scope_default_admin`) are covered by the full `test_bootstrap_scenario_all_virtual_resources` test, so coverage is adequate.

## What's Done Well

1. **Root cause analysis** -- the PR description clearly explains the "virtual resource" concept and the truth table path that leads to incorrect REMOVE actions.
2. **Surgical layering** -- `_enrich_from_tenant_mapping` runs as a post-processing step, only fires when there are actually missing UUIDs, and doesn't touch the primary query loop.
3. **Thorough tests** -- 10 cases covering individual fields, false negatives, mixed real/virtual, and a full bootstrap scenario that mirrors the production code path.

PR #3109, opened 2026-06-25
fix(dr): skip bootstrap_tenant events in DR reconciliation

lpichler:
## Summary

**Alternative to #3108** — instead of enriching the resource checker with TenantMapping fallback queries, skip bootstrap_tenant events entirely during DR reconciliation.

- Filter out `bootstrap_tenant` and `bulk_bootstrap_tenant` events before the existence check / truth table
- Bootstrap events are atomic: the tenant is either fully bootstrapped (all relations valid) or not present at all — per-resource existence checking is the wrong abstraction for these events
- Report `skipped_bootstrap_events` count in the reconciliation summary for observability

## Root Cause

During `bootstrap_tenant`, `TenantMapping` is created with auto-generated UUIDs for default groups and role bindings. Relation tuples are emitted referencing these UUIDs, but **no actual `Group` or `RoleBinding` Django model objects are created**. The DR reconciler's `check_resources_exist` queries only model tables, so it reports these resources as non-existent, triggering `relations_to_add + not exists → REMOVE` — which would delete valid relations from SpiceDB.

## Approach comparison

| | PR #3108 (TenantMapping fallback) | This PR (skip bootstrap) |
|---|---|---|
| **Approach** | Enrich existence check to also query TenantMapping UUID fields | Filter out bootstrap events before truth table |
| **Precision** | Handles edge case where tenant was deleted after bootstrap | Treats bootstrap as all-or-nothing |
| **Complexity** | Adds TenantMapping coupling to resource_checker | Simpler — one filter in service.py |
| **Files changed** | resource_checker.py + tests | service.py + tests |

## Test plan

- [x] New `PartitionBootstrapEventsTest` (5 unit tests): filtering logic for bootstrap/non-bootstrap events
- [x] New `ReconcileBootstrapSkipTest` (3 integration tests):
  - Bootstrap-only window produces zero corrective actions
  - Mixed bootstrap + regular events: bootstrap skipped, regular processed normally
  - Dry run reports skipped bootstrap count
- [x] All 53 DR tests pass

### DIFF
--- a/rbac/management/disaster_recovery/service.py
+++ b/rbac/management/disaster_recovery/service.py
@@ -7,11 +7,14 @@ from management.disaster_recovery.corrective_writer import (
     generate_corrective_actions,
     write_corrective_events,
 )
-from management.disaster_recovery.kafka_reader import read_events_in_window
+from management.disaster_recovery.kafka_reader import ParsedReplicationEvent, read_events_in_window
 from management.disaster_recovery.resource_checker import check_resources_exist
 from management.relation_replicator.outbox_replicator import OutboxReplicator
+from management.relation_replicator.relation_replicator import ReplicationEventType
 
 logger = logging.getLogger(__name__)
+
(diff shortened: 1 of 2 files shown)


randomnetcat:
Are we okay to close this if #3108 was merged?

PR #3114, opened 2026-06-25
RHCLOUD-48737: Add job to re-replicate updated workspaces

randomnetcat:
## Link(s) to Jira
[RHCLOUD-48737](https://redhat.atlassian.net/browse/RHCLOUD-48737)

Due to an HBI bug (see the linked ticket), some workspace events could have been effectively dropped. Add a job to re-replicate them. (We include the option to ignore unchanged default workspaces, since those should already be handled by the bulk job.)

[RHCLOUD-48737]: https://redhat.atlassian.net/browse/RHCLOUD-48737?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rbac/internal/urls.py
```diff
@@ -131,6 +131,7 @@
         name="mcp-tool-descriptions-detail",
     ),
     path("api/utils/replicate_default_workspaces/", views.replicate_default_workspaces),
+    path("api/utils/replicate_updated_workspaces/", views.replicate_updated_workspaces),
```
astrozzc, line 134: What time stamp would it use? Is it last month or weeks? CJI job would be a better choice if the time span is large and the process might take hours. then new deployment won't interrupt the worker and worker won't be too busy

rbac/internal/migrations/replicate_workspaces.py
```diff
@@ -0,0 +1,216 @@
+import dataclasses
+import datetime
+import itertools
+import logging
+import uuid
+from typing import Optional
+
+from django.db.models import F, Q, QuerySet
+from django.db.models.lookups import GreaterThanOrEqual, LessThanOrEqual
```
astrozzc: Nit: GreaterThanOrEqual is not used

rbac/internal/views.py
```diff
@@ -2696,6 +2698,39 @@ def replicate_default_workspaces(request):
         return JsonResponse({"detail": f"Error replicating default workspaces: {str(e)}"}, status=500)
 
 
+@require_http_methods(["POST"])
+def replicate_updated_workspaces(request):
+    """Replicate existing default workspaces.
```
astrozzc: Nit: doc string not updated

rbac/internal/views.py
```diff
@@ -2696,6 +2698,39 @@ def replicate_default_workspaces(request):
...
+
+    POST /_private/api/utils/replicate_updated_workspaces/?since=<timestamp>,exclude_unchanged_default_workspaces=<bool>
+
+    Replicates
+
+    Returns:
+        JSON response indicating the task has been queued
+    """
+    since = request.GET["since"]
```
astrozzc, line 2714: Should we add a try catch block to validate the format of it? Or at least provide an example of the format here would be helpful

rbac/internal/views.py
```diff
@@ -2696,6 +2698,39 @@ def replicate_default_workspaces(request):
         return JsonResponse({"detail": f"Error replicating default workspaces: {str(e)}"}, status=500)
 
 
+@require_http_methods(["POST"])
+def replicate_updated_workspaces(request):
+    """Replicate existing default workspaces.
+
+    POST /_private/api/utils/replicate_updated_workspaces/?since=<timestamp>,exclude_unchanged_default_workspaces=<bool>
```
astrozzc: Nit: it is "&" instead of "," between the parameter

rbac/internal/migrations/replicate_workspaces.py
```diff
@@ -0,0 +1,216 @@
...
+        return 0
+
+    workspaces = list(
+        Workspace.objects.filter(pk__in=(w.pk for w in raw_workspaces))
+        .select_related("tenant")
+        .select_for_update(of=["self"])
+    )
+
+    for workspace in workspaces:
+        if workspace.type not in _permitted_types:
```
astrozzc, line 46: maybe we can put this into the filter above, we have index for type.
And is it necessary? I saw the functions calling it already filtered the query

astrozzc (approved):
LGTM overall, feel free to merge once ready

rbac/internal/urls.py
randomnetcat: Should be since April. I'll estimate that it'll take 10s of minutes rather than hours. I can switch it to a CJI later if necessary.

rbac/internal/migrations/replicate_workspaces.py
randomnetcat: It's effectively a precondition check; there's a bug if we're trying to replicate the wrong type of workspace, and we should fail loudly.

PR #3136, opened 2026-07-01
RHCLOUD-48922: do role scope migration in batches

randomnetcat:
## Link(s) to Jira
[RHCLOUD-48922](https://redhat.atlassian.net/browse/RHCLOUD-48922)

Currently, when we're migrating the bindings to a change in role scope, we migrate all of the groups and CARs in one big `SERIALIZABLE` transaction. This results in both serialization failures and in holding locks for longer than we should.

To solve this, we can instead migrate the groups and CARs in batches (each in a separate transaction). Unfortunately, this means we have to re-establish our invariants at the beginning of *each* transaction, which makes the code a lot more complicated (since it has to thread a bunch of things down to where the transactions actually happen). I don't think there's a way to avoid this, though.

[RHCLOUD-48922]: https://redhat.atlassian.net/browse/RHCLOUD-48922?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rbac/internal/migrations/migrate_role_scope.py
+++ b/rbac/internal/migrations/migrate_role_scope.py
@@ -17,10 +17,12 @@
 
 """Handler for system defined roles."""
 
+import dataclasses
+import itertools
 import logging
 from typing import Optional
 
-from management.atomic_transactions import atomic_with_retry
+from management.atomic_transactions import atomic_block, atomic_with_retry
 from management.group.model import Group
 from management.permission.scope_service import ImplicitResourceService
 from management.relation_replicator.outbox_replicator import OutboxReplicator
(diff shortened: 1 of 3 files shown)


rbac/internal/migrations/migrate_role_scope.py
```diff
@@ -33,7 +35,109 @@
...
+
+    This should be run in the same SERIALIZABLE transaction as the actual migration (otherwise, the results of the
+    check will be out of date).
+    """
+    # We do not need to lock to prevent updates from concurrent seeding, since both this and seeding are in
+    # SERIALIZABLE transactions.
+    role = Role.objects.filter(pk=role.pk).first()
+
+    if role is None:
+        logger.info(f"System role {role.name!r} concurrently deleted; not updating binding scopes.")
```
astrozzc: This is under role is None. role.name would raise error.

rbac/internal/migrations/migrate_role_scope.py
```diff
@@ -46,131 +150,139 @@ def migrate_role_scope_if_changed(v1_role: Role, replicator: Optional[RelationRe
...
-            role=v1_role,
-            version=0,
-            computed_scopes=list(expected_scopes),
-            migrated=False,
-        )
+        # Convince mypy.
+        assert final_check.scope_state is not None
 
-    _migrate_bindings_for_scope_change(v1_role, replicator)
+        if final_check:
```
astrozzc: if final_check.can_migrate?

astrozzc (approved):
Over all LGTM! Feel free to merge after addressing

vbelchio:
QE: NA

PR #3200, opened 2026-07-17
fix(ci): always run JIRA ticket check on edited PRs [RHCLOUD-49150]

lpichler:
## Summary

- Fix JIRA ticket check being skipped on PR body edits, which overwrites the previous failure status due to `cancel-in-progress` concurrency
- Remove the `title.from` change guard so the check always runs on both `opened` and `edited` events

## Root Cause

The job condition only ran the check when the PR was opened or when the **title** was changed:

```yaml
if: github.event.action == 'opened' || github.event.changes.title.from != ''
```

When a PR **body** was edited (not the title), the `edited` event fired but the condition evaluated to `false`, causing the job to be **skipped**. Combined with `cancel-in-progress: true` concurrency, the skipped run replaced the previous failure — making the check appear to pass.

## Fix

Remove the title-change guard. The check now runs on every `opened`/`edited` event (still skipped for `bot`-labeled PRs). A body-only edit harmlessly re-runs the title check, keeping the correct pass/fail status.

## Test plan

- [x] Verified PR [#3197](https://github.com/project-kessel/insights-rbac/pull/3197) showed the bug (first run FAILURE, second run SKIPPED)
- [x] After merge, edit a PR body without a JIRA ticket in the title — check should fail, not skip

## Summary by CodeRabbit

* **Chores**
  * Updated pull request validation so checks run consistently unless the pull request has a bot label.

### DIFF
--- a/.github/workflows/jira-ticket-check.yml
+++ b/.github/workflows/jira-ticket-check.yml
@@ -16,9 +16,7 @@ jobs:
   jira-ticket-check:
     runs-on: ubuntu-latest
     if: >-
-      !contains(github.event.pull_request.labels.*.name, 'bot') &&
-      (github.event.action == 'opened' ||
-      github.event.changes.title.from != '')
+      !contains(github.event.pull_request.labels.*.name, 'bot')
     steps:
       - name: Check PR title for JIRA ticket
         env:


lpichler:
JIRA: https://redhat.atlassian.net/browse/RHCLOUD-49150

dsova06 (approved):
## PR Review

Libor removes the title-change guard from the `jira-ticket-check` GitHub Actions workflow in a single commit (+1/-3 lines in one file). The previous condition only ran the check on `opened` events or when the PR title was specifically changed. When a PR body was edited, the `edited` event fired but the condition was false, causing a skipped run that -- combined with `cancel-in-progress: true` -- overwrote a previous failure status, effectively making the check appear to pass. The fix simplifies the condition to just the bot-label check.

### Findings

None. The change is correct and well-reasoned:

1. **The root cause analysis is accurate.** The `cancel-in-progress: true` concurrency setting means a body-edit `edited` event cancels the previous run, and the skipped replacement overwrites the failure status. Removing the title-change guard fixes this.

2. **No semantic regression.** The workflow triggers are `opened` and `edited` -- removing the title guard means a body-only edit harmlessly re-checks the title. The `cancel-in-progress` is still correct since only the latest run matters.

3. **The bot-label guard is preserved.** PRs labeled `bot` still skip the check as intended.

4. **No security concern.** The workflow uses `pull_request_target` (needed for write permissions to post comments), but the workflow only reads `github.event.pull_request.title` -- no checkout of untrusted code, no injection vector through `PR_TITLE` since it's passed via env var (not inline in the `run:` block).

### Verdict

**Approve.** No issues. Clean, well-scoped bug fix with a clear root cause explanation and verified test plan.

### What's Done Well

1. **Excellent root cause documentation.** The PR body explains the interaction between `cancel-in-progress` and the conditional skip -- this is the kind of subtle CI behavior that trips teams up, and it's documented clearly.

2. **Minimal fix.** Rather than adding more complex event filtering, the fix correctly identifies that the simplest solution (always run the check) has no downside since re-checking the title on body edits is a no-op cost.

3. **Verified against the real failure.** The test plan references the specific PR (#3197) that exhibited the bug.

PR #3239, opened 2026-07-27
[RHCLOUD-49587] Update internal tool to bootstrap tenant with existing ungrouped-host…

astrozzc:
Extend the existing internal API POST /_private/api/utils/bootstrap_tenant/ so operators can bootstrap a tenant from org_id alone (without a pre-existing Tenant row or user), and optionally supply a pre-existing ungrouped-hosts workspace UUID to use instead of generating a new one.

## Link(s) to Jira
- https://redhat.atlassian.net/browse/RHCLOUD-49587

Today bootstrap_tenant requires the Tenant row to already exist (404 via get_object_or_404) and only bootstraps root/default workspaces plus default access bindings. It does not create the ungrouped-hosts workspace. That workspace is currently created lazily via the S2S endpoint GET /_private/_s2s/workspaces/ungrouped/ (PSK/JWT auth, not Turnpike internal API) or via get_or_create_ungrouped_workspace() in other code paths.

Ephemeral/DR tooling and other internal workflows currently work around this by inserting a bare api_tenant row in Postgres, calling bootstrap_tenant, then separately handling ungrouped-hosts via S2S or manual steps. We need a single Turnpike-accessible internal API that can create the tenant if missing, bootstrap V2 tenant structure, and create ungrouped-hosts with a caller-supplied UUID when provided.

## Local Testing
Unit test

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

## Summary by CodeRabbit

- **New Features**
  - Added `create_missing=true` to auto-create missing tenants during tenant bootstrap.
  - Expanded tenant bootstrap to accept a `tenants` payload with optional `ungrouped_hosts_id` per tenant.
- **Documentation**
  - Updated API docs and examples to describe the new request shape and default missing-tenant behavior (404 unless `create_missing` is enabled).
- **Bug Fixes**
  - Improved conflict/mismatch handling for `ungrouped_hosts_id`, returning clearer 400 responses and preserving existing workspaces when appropriate.
- **Tests**
  - Added/expanded endpoint tests for auto-creation, ungrouped-hosts workspace creation/idempotency, and validation/error cases (including mutually exclusive formats and invalid payloads).

### DIFF
--- a/.cursor/skills/internal-api/SKILL.md
+++ b/.cursor/skills/internal-api/SKILL.md
@@ -78,10 +78,19 @@ sh .cursor/skills/internal-api/scripts/internal-api.sh stage POST \
 
 ### Bootstrap tenant by org ID
 
+By default returns 404 if the Tenant row is missing. Pass `create_missing=true` to create it. Optional `ungrouped_hosts_id` creates the ungrouped-hosts workspace with that UUID.
+
 ```bash
+# Legacy body (tenant must already exist)
 sh .cursor/skills/internal-api/scripts/internal-api.sh stage POST \
   utils/bootstrap_tenant/ \
   '{"org_ids":["1234567"]}'
+
+# Create missing tenant + optional ungrouped-hosts workspace UUID
+sh .cursor/skills/internal-api/scripts/internal-api.sh stage POST \
+  utils/bootstrap_tenant/ \
+  create_missing=true \
+  '{"tenants":[{"org_id":"1234567","ungrouped_hosts_id":"0194a1b2-c3d4-7890-abcd-ef1234567890"}]}'
 ```
 
 ### Bootstrap pending tenants
--- a/rbac/internal/specs/openapi.json
+++ b/rbac/internal/specs/openapi.json
@@ -1749,8 +1749,8 @@
           "Tenant",
           "Utils"
         ],
-        "summary": "Bootstrap a tenants by org ids",
-        "description": "Bootstrap a tenants by org ids",
+        "summary": "Bootstrap tenants by org ids",
+        "description": "Bootstrap tenants by org ids. By default returns 404 if the Tenant row does not exist; pass create_missing=true to create it. Optionally accepts a tenants array with ungrouped_hosts_id to create the ungrouped-hosts workspace with a caller-supplied UUID.",
         "operationId": "TenantBootstrap",
         "requestBody": {
           "required": true,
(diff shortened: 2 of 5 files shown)


dsova06 (commented):
## Code Review: PR #3239

**Overall**: Solid implementation. The race condition handling in `get_or_create_ungrouped_workspace` is thorough and the parsing logic is well-validated. A few items to consider:

---

### P1 -- `.replace("'", '"')` extended to new body format

**`rbac/internal/views.py:1152`**

```python
body = json.loads(request.body.decode("utf-8").replace("'", '"'))
```

This apostrophe-to-double-quote substitution is legacy behavior inherited from the existing code. Extending it to the new `tenants` format means that a payload like `{"tenants":[{"org_id":"O'Reilly"}]}` would be silently corrupted before JSON parsing. For this internal API with UUIDs and numeric org_ids, the practical risk is near-zero, but it would be cleaner to parse the body directly:

```python
body = json.loads(request.body.decode("utf-8"))
```

CodeRabbit flagged this too. The legacy `org_ids` path could keep the replace if backward compat is a concern, but the new code path should parse clean JSON.

### P2 -- Missing negative test for `create_missing=false` default

The tests cover `create_missing=true` and the existing tests pre-create tenants before calling the endpoint. There's no test that sends a request for a non-existent tenant *without* `create_missing` and asserts 404. This would document the default-safe behavior described in the docstring:

```python
def test_bootstrapping_tenant_missing_without_create_missing_returns_404(self):
    """Default create_missing=false returns 404 for non-existent tenants."""
    payload = {"org_ids": ["nonexistent-org"]}
    response = self.client.post(
        "/_private/api/utils/bootstrap_tenant/",
        data=payload, **self.request.META, content_type="application/json",
    )
    self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
```

### P3 -- `Workspace.objects.select_for_update().get()` can raise `DoesNotExist`

**`rbac/internal/utils.py:972`**

```python
conflict = Workspace.objects.select_for_update().get(id=workspace_id)
```

This is in the `else` branch of the IntegrityError handler -- when neither the tenant's ungrouped-hosts workspace nor the UUID workspace was found in the retry queries. In theory, this shouldn't happen (the IntegrityError implies *something* exists with that constraint), but if there's a concurrent DELETE between the IntegrityError and this `.get()`, it would raise `Workspace.DoesNotExist` unhandled. Extremely unlikely in practice, but a `.first()` with a fallback raise would be more defensive:

```python
conflict = Workspace.objects.select_for_update().filter(id=workspace_id).first()
if conflict:
    raise UngroupedWorkspaceError(...) from None
raise UngroupedWorkspaceError(
    f"Workspace id {workspace_id} conflicted during creation but is no longer found."
) from None
```

### Minor observations (no action needed)

- **Redundant `description` field** in the `Workspace()` constructor at line 949 -- `Workspace.save()` calls `full_clean()` which auto-sets `description` for `UNGROUPED_HOSTS` type. Harmless but redundant.
- **`from None`** on lines 968 and 976 intentionally suppresses the IntegrityError traceback. This is reasonable since the IntegrityError is a known race-condition signal, not an unexpected error.
- **Type hint fix** from `tenant: str` to `tenant: Tenant` is correct -- all callers already pass `Tenant` objects.
- **Backward compatibility** is clean -- `workspace_id=None` default means existing callers are unaffected.

dsova06 (approved):
LGTM

PR #3249, opened 2026-07-28
[RHCLOUD-48737] Allow specifying stream for updated workspace re-replication

randomnetcat:
## Link(s) to Jira
[RHCLOUD-48737](https://redhat.atlassian.net/browse/RHCLOUD-48737)

In order to avoid replication lag, we want to avoid replicating hundreds of thousands of workspace events to the standard stream. To do this, we can replicate any workspaces updated since the initial relevant time to the BULK stream (potentially introducing inconsistency for any workspaces updated since the migration started), then, once the migration finishes, only replicate any workspaces updated *since the first migration started* to the standard stream, avoiding the vast majority of the lag.

[RHCLOUD-48737]: https://redhat.atlassian.net/browse/RHCLOUD-48737?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **New Features**
  - Workspace replication now supports selecting either the Standard or Bulk event stream.
  - Replication responses include the selected stream for clearer status tracking.

- **Bug Fixes**
  - Requests missing the required `since` parameter now return a clear validation error.
  - Invalid stream values are rejected with an appropriate error instead of being processed.
  - Replication events are now routed to the selected stream consistently.

### DIFF
--- a/rbac/internal/views.py
+++ b/rbac/internal/views.py
@@ -2950,6 +2950,12 @@ def replicate_updated_workspaces(request):
     Returns:
         JSON response indicating the task has been queued
     """
+    if "since" not in request.GET:
+        return JsonResponse({"field": "since", "detail": 'missing query parameter "since"'}, status=400)
+
+    if "stream" not in request.GET:
+        return JsonResponse({"field": "since", "detail": 'missing query parameter "stream"'}, status=400)
+
     since = request.GET["since"]
     stream = request.GET["stream"]
     exclude_unchanged_default_workspaces = (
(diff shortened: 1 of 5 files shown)


rbac/internal/views.py
```diff
@@ -2939,14 +2939,19 @@ def replicate_default_workspaces(request):
...
+        &exclude_unchanged_default_workspaces=<bool>
 
     since must be an ISO 8601 datetime string (e.g. 2026-01-01T18:00:00Z).
+    stream must be either "standard" or "bulk".
 
     Returns:
         JSON response indicating the task has been queued
     """
     since = request.GET["since"]
+    stream = request.GET["stream"]
```
astrozzc, line 2960: Nit: Better to validate the "since" and "stream" exist so it would throw 400 instead of 500

astrozzc (approved):

PR #3255, opened 2026-07-29
[RHCLOUD-48526] Attempt to address role scope migration serialization errors

randomnetcat:
## Link(s) to Jira
[RHCLOUD-48526](https://redhat.atlassian.net/browse/RHCLOUD-48526)

We're consistently experiencing serialization failures when running the role scope migration in prod. To address this, we can try (a) reducing the batch size it uses [thus reducing any contention] and (b) retrying failed batches one-by-one.

(Edit: oops, the batch size decrease was accidentally merged in [a different commit](https://github.com/project-kessel/insights-rbac/commit/c078f681d07c56e37bb590c62df08a6887dbc444), but it was meant to be in this PR.)

[RHCLOUD-48526]: https://redhat.atlassian.net/browse/RHCLOUD-48526?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved role-scope migrations by handling database errors more safely.
  * Added a fallback process that migrates records individually when batch migration is unavailable.
  * Increased migration reliability for group and CAR records.

### DIFF
--- a/rbac/internal/migrations/migrate_role_scope.py
+++ b/rbac/internal/migrations/migrate_role_scope.py
@@ -20,10 +20,10 @@
 import dataclasses
 import itertools
 import logging
-from typing import Optional
+from typing import Callable, Optional
 
 from django.conf import settings
-
+from django.db import OperationalError
 from management.atomic_transactions import atomic_block, atomic_with_retry
 from management.group.model import Group
 from management.permission.scope_service import ImplicitResourceService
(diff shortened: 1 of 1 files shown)


rbac/internal/migrations/migrate_role_scope.py
```diff
@@ -256,6 +255,24 @@ def _migrate_bindings_for_scope_change(context: _MigrateContext):
...
 
 
+def migrate_batch_with_fallback[T](
+    migrate_batch_fn: Callable[[_MigrateContext, list[T]], int], context: _MigrateContext, batch: list[T]
+) -> int:
+    try:
+        # Although migrating each entity is still done individually, it's still worth it to attempt batching because it
+        # saves us having to re-validate the migration for every single entity.
+        return migrate_batch_fn(context, batch)
+    except Exception:
```
astrozzc: The fallback catches all Exception subclasses. If the batch fails for a non-transient reason (e.g., RuntimeError from assert_can_migrate() indicating the role was concurrently modified, or a programming error), it will still fall through to the per-item loop. Each individual item will re-check assert_can_migrate() and fail again, producing N extra DB round-trips and N+1 log warnings before the error eventually propagates.
Consider catching only serialization-related exceptions (e.g., django.db.OperationalError or the specific pgtransaction retry exception) so that programming errors and concurrent-modification signals fail fast. Alternatively, at minimum exclude RuntimeError or KeyboardInterrupt.

astrozzc (approved):
Overall LGTM

PR #3276, opened 2026-08-04
[RHCLOUD-49510] Add feature flag gate for V2 read operations

randomnetcat:
## Link(s) to Jira
[RHCLOUD-49510](https://redhat.atlassian.net/browse/RHCLOUD-49510)

Currently, V2 write operations are gated behind an Unleash flag; here, we additionally gate V2 read operations behind a separate flag.

[RHCLOUD-49510]: https://redhat.atlassian.net/browse/RHCLOUD-49510?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


astrozzc (commented):
This might be over complicating the problem. We just need to gate the access endpoint and only for inventory (i think without our gating, hbi is ok, just want to double validate they are doing the right thing, maybe confirm with Keith when he is back).

Maybe confused with the https://redhat.atlassian.net/browse/RHCLOUD-49751.

randomnetcat:
Closing in favor of #3285, which actually does the correct thing.

PR #3287, opened 2026-08-07
[RHCLOUD-49752] Add V2 opt-in state to TenantMapping

randomnetcat:
## Link(s) to Jira
[RHCLOUD-49752](https://redhat.atlassian.net/browse/RHCLOUD-49752)

Add V2 opt-in state to `TenantMapping` (stored in the same way as `v2_write_activated_at`). I'm choosing to put it here because (a) there doesn't seem to be a reason to store it elsewhere, and (b) it lets us add a constraint to prevent a tenant from being converted without opting in [we do just automatically opt the tenant in when converting it for now, but eventually this flag will become authoritative, and we can not do that].

[RHCLOUD-49752]: https://redhat.atlassian.net/browse/RHCLOUD-49752?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added tenant-level V2 opt-in tracking.
  * Added controls to opt tenants in or out before V2 activation.
  * Prevented opting out after V2 writes are activated.
* **Bug Fixes**
  * Enforced consistent V2 activation and opt-in state.
  * Improved tenant mapping validation and locking during activation.
* **Tests**
  * Expanded coverage for opt-in behavior, activation safeguards, and tenant setup scenarios.


rbac/management/tenant_mapping/v2_activation.py
```diff
@@ -85,21 +82,26 @@ def ensure_v2_write_activated(tenant: Tenant):
...
         return
 
-    mapping = TenantMapping.objects.select_for_update().filter(tenant=tenant).first()
-    if mapping is None:
-        raise TenantNotBootstrappedError(
-            f"Tenant {tenant.org_id} has no TenantMapping; V2 writes require tenant bootstrapping."
-        )
+    # Upgrade the TenantMapping to a FOR UPDATE lock. This must exist, since we previously locked the mapping for
+    # this tenant FOR SHARE (preventing it from being concurrently deleted).
+    mapping = TenantMapping.objects.select_for_update().filter(tenant=tenant).get()
```
astrozzc, line 93: The code rabbit comment seems right, for concurrent requests if few taken the "for share" and trying to acquire  the "update" lock, they get stuck.
Might have to acquire the "update" directly

tests/management/inventory_checker/test_bootstrapped_tenant_checker.py
```diff
@@ -25,6 +25,7 @@
 from management.permission.scope_service import Scope
 from management.tenant_mapping.model import DefaultAccessType, TenantMapping
 from tests.identity_request import IdentityRequest
+from tests.v2_util import bootstrap_tenant_for_v2_test
```
astrozzc: Nit, not used

rbac/management/tenant_mapping/v2_activation.py
```diff
@@ -144,10 +146,64 @@ def assert_v1_write_allowed(tenant: Tenant):
...
+def set_v2_opt_in_state(tenant: Tenant, opted_in: bool):
+    """
+    Set the V2 opt-in state of a tenant.
+
+    A V1 tenant can be "opted-in" to V2, indicating that they are willing to be converted to V2. (At time of writing,
+    this flag is not yet checked before converting a tenant.)
+
+    Opting-in a V1 tenant is idempotent. A V2 tenant is necessarily opted-in and cannot opt out.
+    """
+    mapping = TenantMapping.objects.select_for_update().filter(tenant=tenant).first()
```
astrozzc, line 175: This needs to be within transaction.atomic, maybe add a comment or wrap the set_v2_opt_in_state by decoration

rbac/management/tenant_mapping/v2_activation.py
randomnetcat: This is the status quo; I just reworked it a bit because I wanted to be able to reuse some of the existing framework. All V2 operations should be running with retries anyway.

rbac/management/tenant_mapping/v2_activation.py
randomnetcat: Fair enough. I was modelling it after `ensure_v2_write_activated`, which does not have its own transaction, but I guess the latter's main purpose is to establish something *for* the outer transaction, rather than "just make the change".

rbac/management/tenant_mapping/v2_activation.py
randomnetcat: And using `FOR UPDATE` here would effectively serialize all V2 write operations within a tenant (above and beyond what `SERIALIZABLE` does). That's probably not something we want to do?

astrozzc: sounds good

astrozzc (approved):

PR #3308, opened 2026-08-17
[RHCLOUD-50266] feat: DELETE operations fail with 500 when external sync Kafka publish fails

dsova06:
## Summary
- `RBACProducer.send_kafka_message()` now catches all exceptions internally, logs them, and returns a boolean success indicator instead of letting failures propagate to callers
- This prevents Kafka publish failures from surfacing as HTTP 500 errors in API operations (e.g., DELETE endpoints that trigger external sync via pre_delete signals)

### Kafka producer (`rbac/core/kafka.py`)
- Wrapped the body of `send_kafka_message` in a try/except that catches all exceptions, logs via `logger.exception`, and returns `False` on failure or `True` on success
- Previously, any Kafka error (broker unavailable, timeout, serialization) would propagate up to the caller unhandled

### Internal test view (`rbac/internal/views.py`)
- Updated `send_kafka_test_message` to check the new boolean return value and return a proper `JsonResponse` with status 500 when the send fails, rather than crashing with an unhandled exception

### Tests (`tests/core/test_kafka.py`)
- Added `SendKafkaMessageTests` class with 8 tests covering: correct producer arguments, single-header wrapping, header list passthrough, `KafkaError` swallowed, `KafkaTimeoutError` swallowed, `AttributeError` from null producer swallowed, `ValueError` from serialization swallowed, and independent repeated failures
- Cleaned up imports (removed unused `MagicMock`, sorted)

## Review guidance
- The fix is at the producer level, meaning ALL callers of `send_kafka_message` now get fire-and-forget semantics. This is the intent (Kafka messages are best-effort sync notifications), but reviewers should confirm no caller needs to know about send failures to take corrective action.
- The `except Exception` is broad. The ticket calls for swallowing Kafka failures to prevent 500s; this also swallows unexpected errors like `TypeError` or `RuntimeError`. The trade-off is simplicity vs. catching bugs that should surface. The `logger.exception` call ensures nothing is silently lost.
- The ticket mentions specific DELETE signal paths (`external_sync_pre_delete_handler`). This diff does not modify those signal handlers directly -- it fixes the problem one layer down at the producer, which implicitly fixes all callers including the signal handlers.

## Test plan
- New unit tests in `tests.core.test_kafka.SendKafkaMessageTests` verify both success and failure paths
- Run `tox -e py312-fast -- tests.core.test_kafka` to execute the Kafka tests
- Run `tox -e py312` for full suite with coverage

JIRA: https://redhat.atlassian.net/browse/RHCLOUD-50266

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when sending messages through Kafka by enabling safer delivery handling.
  - Connection, timeout, serialization, and delivery failures are now handled and reported clearly.
  - Failed message sends no longer appear successful.
  - Test-message requests now return an appropriate error response when delivery fails.
  - Successful message delivery continues to return the existing success response and behavior.


rbac/core/kafka.py
```diff
@@ -67,11 +67,16 @@ def get_producer(self):
...
-        if headers and not isinstance(headers, list):
-            headers = [headers]
-        producer.send(topic, value=json_data, headers=headers)
+        try:
+            producer = self.get_producer()
+            json_data = json.dumps(message).encode("utf-8")
+            if headers and not isinstance(headers, list):
+                headers = [headers]
+            producer.send(topic, value=json_data, headers=headers)
+            return True
```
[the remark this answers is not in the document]
dsova06: Acknowledged -- Kafka messages are intentionally fire-and-forget (best-effort sync notifications per the ticket). Adding future.get() with timeout would change the design intent and add latency to every API call for a non-critical side-channel.

rbac/core/kafka.py
```diff
@@ -67,11 +67,16 @@ def get_producer(self):
...
-        json_data = json.dumps(message).encode("utf-8")
-        if headers and not isinstance(headers, list):
-            headers = [headers]
-        producer.send(topic, value=json_data, headers=headers)
+        try:
+            producer = self.get_producer()
+            json_data = json.dumps(message).encode("utf-8")
+            if headers and not isinstance(headers, list):
+                headers = [headers]
+            producer.send(topic, value=json_data, headers=headers)
```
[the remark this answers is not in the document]
dsova06: Acknowledged -- send_kafka_test_message is a diagnostic endpoint for verifying Kafka connectivity, not a business operation. No database state to keep consistent with, so the outbox pattern does not apply here.

tests/core/test_kafka.py
```diff
@@ -83,3 +83,58 @@ def test_kafka_producer_correct_init(self, mock_settings, mock_logger, MockKafka
             MockKafkaProducer.get_producer.side_effect = mock_logger.info("Kafka producer initialized successfully")
 
         mock_logger.info.assert_any_call("Kafka producer initialized successfully")
+
+
+class SendKafkaMessageTests(TestCase):
```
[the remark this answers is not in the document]
dsova06: Acknowledged -- these are pure unit tests for the Kafka producer with no identity/tenant/request context needed. The existing KafkaTests class in the same file also uses TestCase directly. IdentityRequest adds unnecessary fixture setup for tests that only exercise a standalone producer mock.

tests/core/test_kafka.py
```diff
@@ -83,3 +83,58 @@ def test_kafka_producer_correct_init(self, mock_settings, mock_logger, MockKafka
...
+    def test_send_swallows_kafka_timeout_error(self, mock_logger):
+        self.mock_kafka.send.side_effect = KafkaTimeoutError("buffer full")
+        self.producer.send_kafka_message("sync-topic", {"action": "update"})
+        mock_logger.exception.assert_called_once()
+
+    @patch("core.kafka.logger")
+    def test_send_swallows_attribute_error_from_bad_producer(self, mock_logger):
+        self.producer.producer = None
+        self.producer.send_kafka_message("sync-topic", {"action": "create"})
+        mock_logger.exception.assert_called_once()
```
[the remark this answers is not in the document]
dsova06: Fixed: all failure tests now capture and assert the return value is False.

tests/core/test_kafka.py
```diff
@@ -83,3 +83,58 @@ def test_kafka_producer_correct_init(self, mock_settings, mock_logger, MockKafka
...
+
+    @patch("core.kafka.logger")
+    def test_send_swallows_serialization_error(self, mock_logger):
+        self.mock_kafka.send.side_effect = ValueError("serialization failed")
+        self.producer.send_kafka_message("chrome-topic", {"data": "test"})
+        mock_logger.exception.assert_called_once()
```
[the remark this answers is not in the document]
dsova06: Fixed: now passes a non-JSON-serializable object (object()), asserts send was not called, and asserts False return value.

dsova06 (commented):
## PR Summary

Single-commit PR by dsova06 that wraps `RBACProducer.send_kafka_message()` in a try/except to prevent Kafka publish failures from propagating as HTTP 500s to API callers. The fix is applied at the producer layer, so all callers (sync handlers, chrome handlers, notification handlers, signal-driven DELETE flows) get fire-and-forget semantics automatically. The internal test endpoint (`send_kafka_test_message`) is updated to check the new boolean return and surface a proper 500 JSON response on failure. 8 new unit tests cover success, header wrapping, and five distinct failure modes.

3 files changed (+75, -10), 1 commit, all CI green.

## Findings

No blocking or significant issues found. Two minor observations below.

[LOW] Test Coverage -- header tests don't assert return value
Location: tests/core/test_kafka.py:99-107
Problem: `test_send_wraps_single_header_in_list` and `test_send_preserves_header_list` verify `send` was called with correct arguments but don't assert the boolean return value. The success return is covered by `test_send_calls_producer_with_correct_args`, so this is not a gap -- just a consistency nit.
Suggestion: Optionally capture `result` and add `self.assertTrue(result)` for symmetry with the other tests.

[LOW] Test naming -- "swallows" verb may mislead future readers
Location: tests/core/test_kafka.py:110
Problem: "Swallows" could imply silent suppression, when in fact every failure is logged with full traceback via `logger.exception()`. The logging assertion in each test makes this clear, so this is cosmetic.
Suggestion: No action needed. The test bodies make the behavior unambiguous.

## Confirmed Bot Findings

Agree with CodeRabbit on tests/core/test_kafka.py:136 -- the test correctly exercises a real `json.dumps` serialization failure using `object()` and asserts `send` was never called. CodeRabbit's original concern was already addressed in the initial commit.

Disagree with CodeRabbit on tests/core/test_kafka.py:128 -- CodeRabbit claimed lines 119, 125, and 137-139 "discard the boolean result," but all failure tests in the submitted code already capture and assert the return value as `False`. The finding was incorrect.

Disagree with CodeRabbit on tests/core/test_kafka.py:88 (IdentityRequest base class) -- these are pure unit tests for a standalone Kafka producer mock with no tenant, identity, or request context. The existing `KafkaTests` class in the same file also uses plain `TestCase`. Using `IdentityRequest` would add unnecessary fixture overhead.

Disagree with CodeRabbit on rbac/core/kafka.py:76 (wait for Kafka future) -- Kafka messages in this codebase are best-effort sync notifications. Adding `future.get(timeout=...)` would add latency to every API call for a non-critical side-channel. The fire-and-forget design is intentional and appropriate.

Disagree with CodeRabbit on rbac/core/kafka.py:75 (outbox pattern) -- the only caller updated in this PR is `send_kafka_test_message`, a diagnostic endpoint with no database state to keep consistent. The other callers (sync/chrome/notification handlers) are signal-driven and already operate outside the request transaction. The outbox pattern does not apply here.

## Verdict

**Approve.** Clean, well-scoped fix at the right layer (producer, not individual callers). The broad `except Exception` is a reasonable trade-off for a fire-and-forget notification system, and `logger.exception()` ensures nothing is silently lost. Tests are thorough. PR description is excellent -- the "Review guidance" section proactively addresses the main design questions.

(Posted as comment -- GitHub does not allow self-reviews with formal approval/rejection state.)

## What's Done Well

1. **Fix at the right abstraction level.** Catching at the producer layer means all 6+ callers (sync, chrome, notifications, DLQ, diagnostic endpoint) get resilient behavior without individual changes. This is better than sprinkling try/except across every signal handler.

2. **Thorough test coverage.** 8 tests covering the happy path, header normalization, and 5 distinct failure modes (KafkaError, KafkaTimeoutError, AttributeError from None producer, serialization TypeError, and repeated independent failures). The repeated-failures test is a nice touch -- it confirms exceptions don't corrupt producer state between calls.

3. **Excellent PR description.** The "Review guidance" section explicitly calls out the design trade-offs (broad catch, fire-and-forget semantics, fix-at-producer-not-callers), saving reviewer time and demonstrating clear reasoning.

lpichler (commented):
**Automated review findings — 4 actionable items (1 in review body, 3 inline)**

---

**[medium] Finding #2: Callers don't check `send_kafka_message` return value — silent message loss**

`send_kafka_message` now returns `bool` but three callers still use the old fire-and-forget pattern without checking the result: `sync_handlers.send_sync_message`, `chrome_handlers.send_chrome_message`, and `notification_handlers.notify`. Only `views.py` (the test endpoint) was updated. For DELETE operations — the PR's stated motivation — sync messages can be quietly dropped with no caller awareness or compensating mechanism.

**Suggestion**: Add return value checks to the remaining three callers with appropriate per-caller degradation strategies (e.g., log-and-continue for notifications; consider whether critical sync should fail loudly or silently). Consider adding metrics/alerting on publish failures.

rbac/core/kafka.py
```diff
@@ -67,11 +67,16 @@ def get_producer(self):
...
+            return True
+        except Exception:
```
lpichler: **[medium]**: Broad `except Exception` catches programming errors alongside Kafka infrastructure errors

This catch-all will swallow `AttributeError`, `TypeError`, `MemoryError`, and other programming errors that should propagate. The tests even explicitly validate that `AttributeError` and serialization `TypeError` are swallowed — meaning genuine code bugs in message construction will be silently logged and discarded rather than surfacing as failures. This was flagged by 6 independent reviewers (code-quality, architecture, security, error-handling, integration, vulnerability).

**Suggestion**: Catch only the expected failure modes:
```python
except (KafkaError, KafkaTimeoutError, json.JSONDecodeError, BrokerNotAvailableError) as e:
```
This lets unexpected programming errors propagate normally while still preventing Kafka infrastructure failures from causing HTTP 500s.

rbac/core/kafka.py
```diff
@@ -67,11 +67,16 @@ def get_producer(self):
 
     def send_kafka_message(self, topic, message, headers=None):
```
lpichler: **[medium]**: Missing return type hint — new `bool` return is undocumented in the signature

The method now returns `True`/`False` instead of implicit `None`, but the signature still reads `def send_kafka_message(self, topic, message, headers=None):`. This makes the API contract invisible to callers and static analysis tools — which likely explains why three of four callers missed the return value (see review body).

**Suggestion**: Add the return type annotation and a docstring:
```python
def send_kafka_message(self, topic, message, headers=None) -> bool:
    """Send message to kafka server.

    Returns True if sent successfully, False if an error occurred (error is logged).
    """
```

rbac/internal/views.py
```diff
@@ -2454,7 +2454,8 @@ def send_kafka_test_message(request):
         if not topic:
             return HttpResponse("RBAC_KAFKA_CONSUMER_TOPIC is not configured", status=400)
 
-        producer.send_kafka_message(topic, debezium_message)
+        if not producer.send_kafka_message(topic, debezium_message):
```
lpichler, line 2457: **[medium]**: Missing test coverage for this new error path

This new `if not ... return JsonResponse(..., status=500)` path has no test. The new `SendKafkaMessageTests` class tests `RBACProducer` directly but there are no tests for the `send_kafka_test_message` view endpoint itself — so the 500 error response is untested.

**Suggestion**: Add an integration test for this view:
```python
@patch('rbac.internal.views.RBACProducer')
def test_send_kafka_test_message_returns_500_on_kafka_failure(self, mock_producer_class):
    mock_producer_class.return_value.send_kafka_message.return_value = False
    response = self.client.get(reverse('send_kafka_test_message'))
    self.assertEqual(response.status_code, 500)
    self.assertJSONEqual(response.content, {'error': 'Failed to send Kafka message'})
```

rbac/core/kafka.py
```diff
@@ -67,11 +67,16 @@ def get_producer(self):
...
-        producer.send(topic, value=json_data, headers=headers)
+        try:
+            producer = self.get_producer()
+            json_data = json.dumps(message).encode("utf-8")
+            if headers and not isinstance(headers, list):
+                headers = [headers]
+            producer.send(topic, value=json_data, headers=headers)
+            return True
+        except Exception:
+            logger.exception("Failed to send Kafka message to topic '%s'", topic)
```
lpichler: **[low]**: Error log message only includes topic name — missing context for non-Kafka failures

`logger.exception()` logs the full traceback (good), but the message only names the topic. For serialization or producer-state failures, knowing the message structure would help debugging without needing to reproduce the issue.

**Suggestion**:
```python
logger.exception(
    "Failed to send Kafka message to topic '%s'. Message type: %s",
    topic,
    list(message.keys()) if isinstance(message, dict) else type(message).__name__
)
```

rbac/core/kafka.py
dsova06: Fixed: narrowed to `(KafkaError, TypeError, ValueError, AttributeError)` -- catches Kafka broker errors, JSON serialization failures, and producer init failures while letting unexpected programming errors propagate.

rbac/core/kafka.py
dsova06: Fixed: added `-> bool` return type hint and docstring documenting the return contract.

rbac/internal/views.py
dsova06: Fixed: added `SendKafkaTestMessageTests` with tests for both the 500 error path and the 200 success path.

rbac/core/kafka.py
dsova06: Fixed: log message now includes message type context (`list(message.keys())` for dicts, `type(message).__name__` otherwise).

tests/internal/test_views.py
```diff
@@ -5885,3 +5885,36 @@ def test_kessel_parity_check_org_ids_with_non_string(self):
...
+        user.account = self.customer_data["account_id"]
+        self.request.user = user
+
+    @patch("core.kafka.RBACProducer")
+    def test_send_kafka_test_message_returns_500_on_failure(self, mock_producer_class):
+        mock_producer_class.return_value.send_kafka_message.return_value = False
+        response = self.client.get("/_private/api/utils/kafka_test_message/", **self.request.META)
+        self.assertEqual(response.status_code, 500)
+        data = json.loads(response.content)
+        self.assertEqual(data["error"], "Failed to send Kafka message")
```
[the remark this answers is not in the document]
dsova06: Acknowledged -- these tests verify the view's response contract (status codes and JSON bodies), not the producer's internal wiring. Adding mock call assertions would couple the tests to implementation details.

lpichler (approved):

PR #3311, opened 2026-08-18
[COST-8088] add ensure_user command for on-prem bootstrap

ELK4N4:
## Summary

- Add generic `manage.py ensure_user` to create an org tenant + principal, optionally attach `admin_default` roles to a per-org admin group, and run `bootstrap_tenants`.
- Supports configurable group/policy names and optional `--application` filters (defaults to all `admin_default` roles when omitted).
- Replaces embedded Django ORM in the cost on-prem operator bootstrap job (COST-8088).

## Test plan

- [x] `tests.management.management.commands.test_ensure_user` (8 tests)
- [ ] `tox -r -e py312-fast -- tests.management.management.commands.test_ensure_user`
- [ ] Manual: `migrate` → `seeds` (with cost catalog) → `ensure_user --admin ...`

## Related

- Operator PR: https://github.com/project-koku/koku-service-operator/pull/82

## Summary by CodeRabbit

* **New Features**
  * Added a command to create or retrieve an organization tenant and principal.
  * Supports optional administrator setup with filtered default roles, custom group names, and policies.
  * Automatically bootstraps tenants, with an option to skip bootstrapping.
  * Safely supports repeated executions and invalidates tenant access caches after successful changes.

* **Bug Fixes**
  * Added validation and clear errors for missing setup data or bootstrap failures.
  * Continues tenant bootstrapping if cache invalidation fails.

* **Tests**
  * Added comprehensive coverage for setup, permissions, customization, caching, failures, and idempotent reruns.


rbac/management/management/commands/ensure_user.py
```diff
@@ -0,0 +1,174 @@
...
+                policy.roles.set(admin_roles)
+                group.principals.add(principal)
+
+                role_names = list(admin_roles.values_list("name", flat=True))
+                logger.info(f"Granted admin roles {role_names} to {username!r} in org {org_id}")
+
+        cache.clear()
+        logger.info(f"Purged access cache after ensure_user for org_id={org_id}")
+
+        call_command("bootstrap_tenants", "--org-id", org_id, "--force", verbosity=options.get("verbosity", 1))
```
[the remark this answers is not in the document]
ELK4N4: This is no REST logic

rbac/management/management/commands/ensure_user.py
```diff
@@ -0,0 +1,174 @@
...
+                    logger.info(f"Using existing admin-default group {group.name!r}")
+
+                policy, _ = Policy.objects.get_or_create(
+                    name=admin_policy_name,
+                    tenant=tenant,
+                    group=group,
+                    defaults={"system": True},
+                )
+                policy.roles.set(admin_roles)
+                group.principals.add(principal)
```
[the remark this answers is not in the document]
ELK4N4: We're not going to route these writes through `RelationApiDualWriteGroupHandler`.

`policy.roles.set()` / `group.principals.add()` are V1 Postgres writes by design. That's what on-prem Cost Management uses: the principal is a member of a group that holds the filtered `admin_default` roles, so `groups_for_principal` grants access even without `is_org_admin` on the token.

Dual-write on this group would not grant V2/Kessel admin access. Org-admin default in V2 is `TenantMapping.default_admin_group_uuid` (a synthetic UUID) plus the default admin role bindings created by `bootstrap_tenants`. Membership is written by `V2TenantBootstrapService.update_user()` when `user.admin=True`, not by dual-writing a per-tenant `admin_default` Group.

Using the handler here would also be incorrect mechanically:

- It requires root/default workspaces, which don't exist until after `bootstrap_tenants` (currently after this transaction).
- Membership replication skips principals with no `user_id`; this command doesn't set one.
- Even if it wrote tuples, they'd be on `Group.uuid`, not `default_admin_group_uuid`.

On-prem also has `REPLICATION_TO_RELATION_ENABLED` off and Kafka disabled, so the handler would no-op.

If we later need Kessel admin membership for this bootstrap user, the follow-up is `update_user(..., upsert=True)` after bootstrap — not dual-write of this V1 group.

rbac/management/management/commands/ensure_user.py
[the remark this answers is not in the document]
ELK4N4: We're leaving `cache.clear()` and `bootstrap_tenants` after the provisioning `atomic()` block. The outbox guideline does not apply here.

That rule is about keeping a DB mutation and its outbox insert in the same transaction so dual-write to Kessel is atomic. This command is not writing the outbox. `cache.clear()` is Redis, and `bootstrap_tenants` is a separate command with its own transactions.

Cache invalidation has to run *after* commit. If we flush Redis while the write is still uncommitted, another request can refill the cache from the old committed rows; then we commit and the cache is stale. After-commit clear is the correct pattern. There is no existing work-item/outbox workflow for AccessCache, and we should not add one for this.

`bootstrap_tenants` uses its own `transaction.atomic()` and `select_for_update`. Nesting it in the provisioning transaction, or persisting a work item to dispatch it later, is a heavy lift for a case retry already covers. `ensure_user` is idempotent (`get_or_create`). If bootstrap fails after V1 commit, the operator job retries, finds the tenant/principal/group, and runs `bootstrap_tenants --force` again.

A failed cache clear or bootstrap after persist is acceptable: V1 authorization is already in Postgres (what Cost uses on-prem), and bootstrap is additive (workspaces / TenantMapping).

tests/management/management/commands/test_ensure_user.py
```diff
@@ -0,0 +1,223 @@
...
+                "version": 2,
+            },
+        )
+        role.admin_default = admin_default
+        role.save()
+        Access.objects.get_or_create(role=role, permission=permission, defaults={"tenant": self.public_tenant})
+        return role
+
+    def _invoke(self, *args, **kwargs):
+        call_command("ensure_user", *args, **kwargs)
```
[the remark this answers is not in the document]
ELK4N4: We're not going to assert the `call_command()` result or mock `cache.clear()` on every success path.

`ensure_user.handle()` does not return a value, so `call_command()` is `None`. Returning that from `_invoke` and asserting it would not test anything. Other management-command tests in this repo (including `bootstrap_tenants`) use the same helper shape.

The path instruction about return values and side effects is for API/service tests that have a payload. These tests already assert the side effects that matter: tenant, principal, group, policy, filtered roles, and that `bootstrap_tenants` is invoked.

`cache.clear()` is a real Redis flush. Spreading a mock of it across every success path would only prove the name was called, and would stop the tests from actually hitting clear. A dedicated cache-clear test is optional; it is not needed for this command's contract.

rbac/management/management/commands/ensure_user.py
```diff
@@ -0,0 +1,174 @@
...
+
+                role_names = list(admin_roles.values_list("name", flat=True))
+                logger.info(f"Granted admin roles {role_names} to {username!r} in org {org_id}")
+
+        cache.clear()
```
lpichler: **[critical]**: `cache.clear()` flushes the entire Redis cache, affecting all tenants

This is called outside the `transaction.atomic()` block with no error handling. If Redis is unavailable, the command crashes *after* DB changes are already committed, leaving inconsistent state. The `seeds` command explicitly avoided this approach (letting the 10-min TTL expire naturally).

The `AccessCache` module at `management/cache.py` provides `delete_all_policies_for_tenant()` for targeted invalidation.

**Suggestion**: Replace with per-tenant cache invalidation:
```python
from management.cache import AccessCache
try:
    AccessCache(tenant.org_id).delete_all_policies_for_tenant()
except Exception:
    logger.warning("Cache invalidation failed; permissions may be stale for up to 10 min")
```
Consider wrapping in `transaction.on_commit()` to close the stale-cache window between commit and invalidation.

rbac/management/management/commands/ensure_user.py
```diff
@@ -0,0 +1,174 @@
...
+        logger.info(f"Purged access cache after ensure_user for org_id={org_id}")
+
+        call_command("bootstrap_tenants", "--org-id", org_id, "--force", verbosity=options.get("verbosity", 1))
```
lpichler: **[high]**: `call_command('bootstrap_tenants')` has no error handling and no user feedback

This runs after the transaction has committed. If it fails or hangs, all DB changes (tenant, principal, group, policy) are already persisted with no rollback path and no clear error message.

**Suggestion**:
```python
logger.info(f"Running bootstrap_tenants for org_id={org_id}...")
try:
    call_command("bootstrap_tenants", "--org-id", org_id, "--force", verbosity=options.get("verbosity", 1))
except Exception as exc:
    raise CommandError(
        f"User created but bootstrap_tenants failed for org_id={org_id}: {exc}. "
        "Re-run 'manage.py bootstrap_tenants --org-id ... --force' to complete setup."
    ) from exc
```
Consider a `--skip-bootstrap` flag to allow decoupled execution.

rbac/management/management/commands/ensure_user.py
```diff
@@ -0,0 +1,174 @@
...
+                ) from exc
+            tenant, created = Tenant.objects.get_or_create(
+                org_id=org_id,
+                defaults={"tenant_name": "acct" + account_number, "ready": True},
+            )
+            logger.info(
+                f"{'Created' if created else 'Existing'} tenant for org_id={org_id} (name={tenant.tenant_name})"
+            )
+
+            principal, principal_created = Principal.objects.get_or_create(
```
lpichler, line 140: **[high]**: Case-sensitive username lookup breaks idempotency

`Principal.save()` lowercases `username` before storing, but this `get_or_create` uses the raw (potentially mixed-case) value from `--username`. On a second run with `--username=Alice`, `get()` fails to find the existing `'alice'` record, then `create(username='Alice')` hits an `IntegrityError` because `save()` lowercases it back to `'alice'` — which already exists. Django's `get_or_create` retry also fails since it retries with the original mixed-case value.

**Suggestion**: Normalize at the top of `handle()`:
```python
username = options["username"].lower()
```

rbac/management/management/commands/ensure_user.py
```diff
@@ -0,0 +1,174 @@
...
+                )
+                if not group_created and not group.admin_default:
+                    group.admin_default = True
+                    group.save(update_fields=["admin_default"])
+                if group_created:
+                    logger.info(f"Created admin-default group {group.name!r}")
+                else:
+                    logger.info(f"Using existing admin-default group {group.name!r}")
+
+                policy, _ = Policy.objects.get_or_create(
```
lpichler, line 183: **[medium]**: `Policy.get_or_create` lookup includes `group`, but the unique constraint is only `(name, tenant)`

If re-invoked with the same `--admin-policy-name` and `--org-id` but a different `--admin-group-name`, the `get()` misses the existing policy (group FK mismatch), then `create()` hits an `IntegrityError` on the `(name, tenant)` unique constraint. The command crashes instead of reusing the existing policy.

**Suggestion**: Remove `group` from the lookup fields and move to `defaults`:
```python
policy, _ = Policy.objects.get_or_create(
    name=admin_policy_name,
    tenant=tenant,
    defaults={"system": True, "group": group},
)
```

rbac/management/management/commands/ensure_user.py
```diff
@@ -0,0 +1,174 @@
...
+                    name=admin_policy_name,
+                    tenant=tenant,
+                    group=group,
+                    defaults={"system": True},
+                )
+                policy.roles.set(admin_roles)
+                group.principals.add(principal)
+
+                role_names = list(admin_roles.values_list("name", flat=True))
```
lpichler: **[low]**: `admin_roles` queryset is evaluated twice — redundant database query

`policy.roles.set(admin_roles)` on line 165 already evaluates the queryset. `admin_roles.values_list("name", flat=True)` here executes a second SELECT for the same data.

**Suggestion**:
```python
admin_roles_list = list(admin_roles)
policy.roles.set(admin_roles_list)
group.principals.add(principal)

role_names = [r.name for r in admin_roles_list]
```

tests/management/management/commands/test_ensure_user.py
```diff
@@ -0,0 +1,223 @@
...
+from django.core.management import CommandError, call_command
+from django.test.utils import override_settings
+
+from api.models import Tenant
+from management.models import Access, Group, Permission, Policy, Principal, Role
+from tests.identity_request import IdentityRequest
+
+
+@override_settings(V2_BOOTSTRAP_TENANT=True)
+class TestEnsureUser(IdentityRequest):
```
lpichler, line 30: **[high]**: `cache.clear()` is never mocked in any test — every test run actually clears the application cache

The implementation calls `cache.clear()` (ensure_user.py line 171), but none of the tests patch it. This violates test isolation: every test invocation flushes the real cache, potentially polluting other tests that depend on cached data.

**Suggestion**: Add a class-level patch or patch each test:
```python
@patch("management.management.commands.ensure_user.cache")
@patch("management.management.commands.ensure_user.call_command")
def test_creates_tenant_and_principal_without_admin(self, mock_bootstrap, mock_cache):
    ...
    mock_cache.clear.assert_called_once()
```
Or add `@patch("management.management.commands.ensure_user.cache")` as a class decorator alongside `@override_settings`.

tests/management/management/commands/test_ensure_user.py
```diff
@@ -0,0 +1,223 @@
...
+                "--org-id=org1",
+                "--account-number=123",
+            )
+
+    def test_admin_requires_application(self):
+        """--admin without --application uses all admin_default roles."""
+        self._create_role("Cost Administrator", "cost-management", admin_default=True)
+        self._create_role("Sources administrator", "sources", admin_default=True)
+
+        with patch("management.management.commands.ensure_user.call_command"):
```
lpichler: **[medium]**: `with patch(...)` doesn't capture the mock — bootstrap_tenants arguments cannot be verified

The context manager form doesn't assign the mock object, so there's no way to assert what arguments were passed to `bootstrap_tenants` (or even that it was called).

**Suggestion**:
```python
with patch("management.management.commands.ensure_user.call_command") as mock_call:
    self._invoke(
        "--username=alice",
        "--org-id=org1",
        "--account-number=123",
        "--admin",
    )
mock_call.assert_called_once_with("bootstrap_tenants", "--org-id", "org1", "--force", verbosity=1)
```

tests/management/management/commands/test_ensure_user.py
```diff
@@ -0,0 +1,223 @@
...
+        principal = Principal.objects.get(username="alice", tenant=tenant)
+        group = Group.objects.get(tenant=tenant, admin_default=True)
+        self.assertIn(principal, group.principals.all())
+
+        policy = Policy.objects.get(group=group, tenant=tenant)
+        role_names = set(policy.roles.values_list("name", flat=True))
+        self.assertEqual(role_names, {"Cost Administrator", "Sources administrator"})
+        self.assertNotIn("Other Admin", role_names)
+        self.assertIn(cost_role, policy.roles.all())
+        mock_bootstrap.assert_called_once()
```
lpichler: **[medium]**: `assert_called_once()` does not verify bootstrap_tenants received correct arguments

This assertion only confirms the mock was called once; it does not verify the command name, `--org-id`, `--force`, or `verbosity` arguments. The same weak assertion also appears at lines 194 and 223.

**Suggestion**: Replace all three occurrences with:
```python
mock_bootstrap.assert_called_once_with("bootstrap_tenants", "--org-id", "org1", "--force", verbosity=1)
```

lpichler (commented):
Automated dimensional review — 8 actionable findings (verified against current code)

PR #3320, opened 2026-08-20
 [RHCLOUD-40195] Allow per-org workspace creation limit overrides

astrozzc:
## Link(s) to Jira
- https://redhat.atlassian.net/browse/RHCLOUD-40195

Store optional limits on Tenant.org_config so customer orgs can exceed the global WORKSPACE_ORG_CREATION_LIMIT without raising it for everyone.

## Local Testing
Unit test

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

## Summary by CodeRabbit

* **New Features**
  * Added per-tenant workspace creation limits that can override the global default.
  * Added API endpoints to view, update, and remove tenant-specific limits.
  * Added validation for configuration updates, including unsupported keys and invalid values.
  * Invalid or removed overrides now safely fall back to the global limit.
* **Bug Fixes**
  * Workspace creation now applies each tenant’s effective configured limit.
  * Kafka test message requests now report failures when message delivery does not succeed.
* **Documentation**
  * Updated multi-tenancy guidelines to explain organization configuration overrides and defaults.


lpichler (commented):
Review findings — 6 actionable items from dimensional review (findings verified against current PR diff).

rbac/internal/views.py
```diff
@@ -265,6 +266,64 @@ def tenant_view(request, org_id):
...
+        except ValueError as e:
+            return JsonResponse({"error": str(e)}, status=400)
+
+        old_limit = tenant.workspace_creation_limit()
+        tenant.org_config = new_config
+        try:
+            tenant.save(update_fields=["org_config"])
+        except ValidationError as e:
+            return JsonResponse({"error": Tenant.org_config_error_message(e)}, status=400)
+        TENANTS.delete_tenant(org_id)
```
lpichler, line 305: **[high]**: Uncaught exception from cache invalidation causes misleading 500 after successful DB save

If `TENANTS.delete_tenant(org_id)` raises (e.g. Redis unavailable), the exception propagates and Django returns a 500, but `tenant.save()` already committed. The client sees a failure for an operation that actually succeeded; the stale cache entry heals on the next read, but the misleading error stands.

**Suggestion**:
```python
try:
    TENANTS.delete_tenant(org_id)
except Exception as e:
    logger.warning(
        "Failed to invalidate tenant cache after org_config update",
        extra={"org_id": org_id, "error": str(e)}
    )
```
Alternatively, wrap in `transaction.on_commit()` so eviction fires only after the DB commit and failures are logged without aborting the response.

rbac/internal/views.py
```diff
@@ -265,6 +266,64 @@ def tenant_view(request, org_id):
...
+    }
+
+
+def tenant_org_config(request, org_id):
+    """Get or update per-org configuration.
+
+    GET /_private/api/utils/tenant_org_config/<org_id>/
+    PATCH /_private/api/utils/tenant_org_config/<org_id>/
+    """
+    tenant = get_object_or_404(Tenant, org_id=org_id)
```
lpichler, line 284: **[high]**: `get_object_or_404` may return an HTML 404, not the JSON Error schema documented in the OpenAPI spec

In a plain Django function-based view (not DRF APIView), `Http404` is handled by Django's global 404 handler, which can return an HTML page. The OpenAPI spec documents that 404 responses return `{"errors": [{"detail": "...", "status": "404"}]}`.

**Suggestion**:
```python
try:
    tenant = Tenant.objects.get(org_id=org_id)
except Tenant.DoesNotExist:
    return JsonResponse(
        {"errors": [{"detail": "Tenant not found.", "status": "404"}]}, status=404
    )
```

rbac/api/models.py
```diff
@@ -50,12 +53,86 @@ class Tenant(models.Model):
...
+        Missing org_config values fall back to WORKSPACE_ORG_CREATION_LIMIT.
+        """
+        default = int(settings.WORKSPACE_ORG_CREATION_LIMIT)
+        org_config = getattr(self, "org_config", None)
+        if not isinstance(org_config, dict):
+            return default
+        raw = org_config.get(self.ORG_CONFIG_WORKSPACE_CREATION_LIMIT)
+        if raw is None:
+            return default
+        return raw
```
lpichler, line 75: **[high]**: `workspace_creation_limit()` returns raw JSON value without type validation, violating the `-> int` return type annotation

`raw` is returned directly from the JSONField without verifying it is an integer. If invalid data is stored via `QuerySet.update()`, a bulk operation, or a data migration that bypasses `save()`, the comparison `workspace_count >= max_limit` in `service.py` will raise `TypeError`, breaking workspace creation for that tenant.

**Suggestion**: Add defensive validation before returning:
```python
if raw is None:
    return default
if isinstance(raw, bool) or not isinstance(raw, int) or raw < 1:
    logger.warning(
        "Invalid workspace_creation_limit in org_config, using default",
        extra={"org_id": self.org_id, "raw": raw}
    )
    return default
return raw
```

rbac/internal/views.py
```diff
@@ -265,6 +266,64 @@ def tenant_view(request, org_id):
...
+                "org_id": org_id,
+                "username": getattr(request.user, "username", None),
+                "old_workspace_creation_limit": old_limit,
+                "workspace_creation_limit": tenant.workspace_creation_limit(),
+            },
+        )
+        return JsonResponse(_tenant_org_config_payload(tenant))
+
+    return JsonResponse(
+        {"errors": [{"detail": 'Invalid method, only "GET" and "PATCH" are allowed.', "status": "405"}]},
```
lpichler, line 322: **[medium]**: Inconsistent error response shape: 405 uses `{"errors": [...]}` while all other errors in this view use `{"error": "..."}`

The 400 responses from this view use `{"error": "message"}` (matching the 11 other usages in `internal/views.py`). This 405 response uses `{"errors": [{"detail": ..., "status": ...}]}`, a different shape. The OpenAPI spec for 405 references `Error` schema but the 400s reference `ErrorSingle` — clients handling both will need to branch on shape.

**Suggestion**: Standardize to match the other errors in this endpoint:
```python
return JsonResponse(
    {"error": 'Invalid method, only "GET" and "PATCH" are allowed.'}, status=405
)
```

rbac/api/models.py
```diff
@@ -50,12 +53,86 @@ class Tenant(models.Model):
...
+            raise ValidationError({"org_config": "org_config must be a JSON object."})
+        if self.ORG_CONFIG_WORKSPACE_CREATION_LIMIT in org_config:
+            value = org_config[self.ORG_CONFIG_WORKSPACE_CREATION_LIMIT]
+            if value is not None:
+                try:
+                    self._validate_workspace_creation_limit_value(value)
+                except ValueError as exc:
+                    raise ValidationError({"org_config": str(exc)}) from exc
+
+    def save(self, *args, **kwargs):
```
lpichler, line 99: **[medium]**: Overriding `save()` to call `clean()` changes the Django contract for every existing `Tenant.save()` caller

Django intentionally does not call `clean()` from `save()` — it is a form/ModelForm concern. Existing callers (`middleware.py:214` during request bootstrap, `api/utils.py:28` and `:89` in batch tenant population, `tenant_service/v2.py` via `Tenant.objects.create`) don't expect `ValidationError` from `save()`. Future additions to `clean()` will silently fire on every `Tenant.save()` everywhere in the codebase.

**Suggestion**: Validate `org_config` at the write boundary (in the view) rather than in `save()`. If a model-level guard is desired, gate the `clean()` call to org_config updates only:
```python
def save(self, *args, **kwargs):
    if "org_config" in (kwargs.get("update_fields") or []):
        self.clean()
    super().save(*args, **kwargs)
```
Document loudly that `save(update_fields=["org_config"])` now raises `ValidationError`.

tests/internal/test_tenant_org_config.py
```diff
@@ -0,0 +1,178 @@
...
+        """Set up internal identity headers."""
+        super().setUp()
+        self.client = APIClient()
+        ctx = self._create_request_context(self.customer_data, self.user_data, is_internal=True)
+        self.internal_headers = ctx["request"].META
+        self.url = f"/_private/api/utils/tenant_org_config/{self.tenant.org_id}/"
+        self.tenant.org_config = {}
+        self.tenant.save(update_fields=["org_config"])
+
+    def tearDown(self):
```
lpichler, line 41: **[medium]**: `tearDown` mutating a shared class-level tenant fixture is fragile and can leak state across tests

`setUp` and `tearDown` both modify `self.tenant.org_config` and call `save()`. `self.tenant` is inherited from `IdentityRequest` and shared across test methods. If a test fails or raises before `tearDown` runs, the modified `org_config` remains and can affect subsequent tests in the class.

**Suggestion**: Use `addCleanup()` in `setUp` to guarantee reset even on test failure:
```python
def setUp(self):
    super().setUp()
    self._original_org_config = dict(self.tenant.org_config)
    self.tenant.org_config = {"workspace_creation_limit": 200}
    self.tenant.save(update_fields=["org_config"])
    self.addCleanup(self._restore_org_config)

def _restore_org_config(self):
    self.tenant.org_config = self._original_org_config
    self.tenant.save(update_fields=["org_config"])
```

lpichler (approved):

PR #3330, opened 2026-08-24
[RHCLOUD-49753] [RHCLOUD-49754] V2 opt-in API

randomnetcat:
## Link(s) to Jira
* [RHCLOUD-49753](https://redhat.atlassian.net/browse/RHCLOUD-49753)
* [RHCLOUD-49754](https://redhat.atlassian.net/browse/RHCLOUD-49754)

Per [KSL-072](https://docs.google.com/document/d/1QW00r3gtp6VdIKW1Xah0NuwH4B1WbZLZbEi86MBG_T8/edit), add APIs to allow users to read their tenant's V2 opt-in state and to allow an org admin to opt their tenant into V2.

[RHCLOUD-49753]: https://redhat.atlassian.net/browse/RHCLOUD-49753?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RHCLOUD-49754]: https://redhat.atlassian.net/browse/RHCLOUD-49754?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added tenant V2 opt-in status and eligibility endpoints.
  * Administrators can opt eligible tenants into V2.
  * Ineligible tenants receive details about affected groups, roles, and applications.
  * Added OpenAPI documentation for the new endpoints and response schemas.

* **Bug Fixes**
  * Improved tenant mapping consistency and locking during opt-in operations.

* **Tests**
  * Added coverage for eligibility checks, authorization, successful opt-in, invalid requests, and error handling.

### DIFF
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -42,6 +42,10 @@
     {
       "name": "Audit Logs",
       "description": "Operations about audit logs requests"
+    },
+    {
+      "name": "V2OptIn",
+      "description": "Operations related to opting into the V2 interface"
     }
   ],
   "paths": {
@@ -2804,6 +2808,181 @@
           }
         }
       }
+    },
+    "/tenant/opt-in/": {
+      "get": {
+        "tags": [
+          "V2OptIn"
+        ],
+        "summary": "Get whether the current tenant has opted into V2",
+        "operationId": "OptInStatus",
+        "parameters": [],
+        "responses": {
+          "200": {
+            "description": "The request has succeeded.",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/OptInStatus"
+                }
+              }
+            }
+          },
+          "400": {
+            "description": "The server could not understand the request due to invalid syntax."
+          },
+          "401": {
+            "description": "Unauthorized"
+          },
+          "403": {
+            "description": "Insufficient permissions to list permissions",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Error403"
+                }
+              }
+            }
+          },
+          "500": {
+            "description": "Unexpected Error",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Error"
+                }
+              }
+            }
+          }
+        }
+      },
+      "patch": {
+        "tags": [
+          "V2OptIn"
+        ],
+        "summary": "Opt the current tenant into V2",
+        "operationId": "OptInUpdate",
+        "parameters": [],
+        "requestBody": {
+          "required": true,
+          "content": {
+            "application/json": {
+              "schema": {
+                "type": "object",
+                "properties": {
+                  "v2_opted_in": {
+                    "type": "boolean",
+                    "enum": [true]
+                  }
+                }
+              }
+            }
+          }
+        },
+        "responses": {
+          "200": {
+            "description": "The request has succeeded.",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/OptInStatus"
+                }
+              }
+            }
+          },
+          "400": {
+            "description": "Bad Input"
+          },
+          "401": {
+            "description": "Unauthorized"
+          },
+          "403": {
+            "description": "Insufficient permissions",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Error403"
+                }
+              }
+            }
+          },
+          "422": {
+            "description": "Opt-in is not possible",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/OptInIneligibleResponse"
+                }
+              }
+            }
+          },
+          "500": {
+            "description": "Unexpected Error",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Error"
+                }
+              }
+            }
+          }
+        }
+      }
+    },
+    "/tenant/opt-in/eligibility/": {
+      "get": {
+        "tags": [
+          "V2OptIn"
+        ],
+        "summary": "Get information about whether the current tenant can be opted into V2",
+        "operationId": "OptInEligibility",
+        "parameters": [],
+        "responses": {
+          "200": {
+            "description": "Contains information about the tenant's eligibility to opt-in to V2.",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "anyOf": [
+                    {
+                      "$ref": "#/components/schemas/OptInEligibleResponse"
+                    },
+                    {
+                      "$ref": "#/components/schemas/OptInIneligibleResponse"
+                    }
+                  ]
+                }
+              }
+            }
+          },
+          "400": {
+            "description": "The server could not understand the request due to invalid syntax."
+          },
+          "401": {
+            "description": "Access is unauthorized."
+          },
+          "403": {
+            "description": "Access is forbidden.",
+            "content": {
+              "application/problem+json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Error403"
+                }
+              }
+            }
+          },
+          "500": {
+            "description": "Server error",
+            "content": {
+              "application/problem+json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Error"
+                }
+              }
+            }
+          }
+        }
+      }
     }
   },
   "servers": [
@@ -4504,6 +4683,110 @@
             }
           }
         ]
+      },
+      "OptInIneligibleGroup": {
+        "type": "object",
+        "required": [
+          "name",
+          "uuid",
+          "ineligible_system_roles"
+        ],
+        "properties": {
+          "name": {
+            "type": "string"
+          },
+          "uuid": {
+            "type": "string",
+            "format": "uuid"
+          },
+          "ineligible_system_roles": {
+            "type": "array",
+            "items": {
+              "$ref": "#/components/schemas/OptInIneligibleRole"
+            }
+          }
+        }
+      },
+      "OptInIneligibleResponse": {
+        "type": "object",
+        "required": [
+          "eligible",
+          "bootstrapped",
+          "ineligible_custom_roles",
+          "ineligible_groups"
+        ],
+        "properties": {
+          "eligible": {
+            "type": "boolean",
+            "enum": [
+              false
+            ]
+          },
+          "bootstrapped": {
+            "type" : "boolean"
+          },
+          "ineligible_custom_roles": {
+            "type": "array",
+            "items": {
+              "$ref": "#/components/schemas/OptInIneligibleRole"
+            }
+          },
+          "ineligible_groups": {
+            "type": "array",
+            "items": {
+              "$ref": "#/components/schemas/OptInIneligibleGroup"
+            }
+          }
+        }
+      },
+      "OptInIneligibleRole": {
+        "type": "object",
+        "required": [
+          "name",
+          "uuid",
+          "ineligible_applications"
+        ],
+        "properties": {
+          "name": {
+            "type": "string"
+          },
+          "uuid": {
+            "type": "string",
+            "format": "uuid"
+          },
+          "ineligible_applications": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            }
+          }
+        }
+      },
+      "OptInEligibleResponse": {
+        "type": "object",
+        "required": [
+          "eligible"
+        ],
+        "properties": {
+          "eligible": {
+            "type": "boolean",
+            "enum": [
+              true
+            ]
+          }
+        }
+      },
+      "OptInStatus": {
+        "type": "object",
+        "required": [
+          "v2_opted_in"
+        ],
+        "properties": {
+          "v2_opted_in": {
+            "type": "boolean",
+            "description": "Whether the tenant has opted into RBAC V2."
+          }
+        }
       }
     }
   }
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -2834,16 +2834,6 @@
           "401": {
             "description": "Unauthorized"
           },
-          "403": {
-            "description": "Insufficient permissions to list permissions",
-            "content": {
-              "application/json": {
-                "schema": {
-                  "$ref": "#/components/schemas/Error403"
-                }
-              }
-            }
-          },
           "500": {
             "description": "Unexpected Error",
             "content": {
@@ -2964,7 +2954,7 @@
           "403": {
             "description": "Access is forbidden.",
             "content": {
-              "application/problem+json": {
+              "application/json": {
                 "schema": {
                   "$ref": "#/components/schemas/Error403"
                 }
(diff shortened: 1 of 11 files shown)


rbac/management/tenant_mapping/v2_opt_in_view.py
```diff
@@ -0,0 +1,145 @@
...
+        """Get the current opt-in state of the requestor's tenant."""
+        return self._state_response_for(request.tenant, headers={"Cache-Control": "max-age=120, private"})
+
+    @atomic_with_retry(retries=5)
+    def partial_update(self, request):
+        """(Possibly) update the requestor's tenant's V2 opt-in state."""
+        serializer = _OptInRequestSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        if serializer.validated_data["v2_opted_in"] is None:
```
lpichler: **[high]**: Empty-body PATCH raises KeyError → HTTP 500

`serializer.validated_data["v2_opted_in"]` uses bracket access, but `BooleanField(required=False, allow_null=True)` with no `default` is **absent** from `validated_data` when the field is not sent. An empty `{}` body therefore raises `KeyError` → 500. The test `test_opt_in_empty_body` (which sends `{}` and expects 200) would fail or never reach the assertion.

**Suggestion**: Replace both bracket reads with `.get()`:
```python
if serializer.validated_data.get("v2_opted_in") is None:
    ...
if serializer.validated_data.get("v2_opted_in") is not True:
    ...
```
Or add `default=None` to the `BooleanField`.

rbac/management/tenant_mapping/v2_eligibility.py
```diff
@@ -0,0 +1,232 @@
...
+    bootstrapped: bool
+    ineligible_groups: tuple[OptInIneligibleGroup, ...]
+    ineligible_custom_roles: tuple[OptInIneligibleRole, ...]
+
+    def __post_init__(self):
+        """Establish invariants."""
+        if not isinstance(self.bootstrapped, bool):
+            raise TypeError(f"Expected bootstrapped to be a bool, but got: {self.bootstrapped!r}")
+
+        if not isinstance(self.ineligible_groups, tuple) and all(
```
lpichler: **[high]**: Operator-precedence bug silently breaks `ineligible_groups` invariant check

Line 97: `if not isinstance(self.ineligible_groups, tuple) and all(...)` parses as `(not isinstance(...)) and all(...)` — the opposite of the intended guard. Compare the correct sibling at lines 105–108 which wraps in parentheses: `if not (isinstance(...) and all(...))`.

As written: a valid non-tuple list raises; a tuple containing wrong-typed elements passes silently.

**Suggestion**:
```python
if not (isinstance(self.ineligible_groups, tuple) and all(isinstance(g, OptInIneligibleGroup) for g in self.ineligible_groups)):
```

rbac/management/tenant_mapping/v2_opt_in_view.py
```diff
@@ -0,0 +1,145 @@
...
+            # We've not been asked to opt the tenant in, so we have nothing to do.
+            return self._state_response_for(request.tenant)
+
+        if serializer.validated_data["v2_opted_in"] is not True:
+            raise AssertionError("Validation should have rejected v2_opted_in being false")
```
lpichler, line 139: **[high]**: `AssertionError` surfaces as HTTP 500 in request-handling code

Three `raise AssertionError(...)` sites are reachable from request handling (lines 112, 131, and `model.py:159`). DRF does not translate `AssertionError` to a response, so any of them becomes an unhandled 500. Assertions are also stripped under `python -O`, silently disabling the guards.

**Suggestion**: Replace request-path `AssertionError`s with explicit exceptions that map to appropriate responses (e.g. a domain exception the DRF exception handler renders as 500/422, or `rest_framework.exceptions.APIException`). Reserve `assert` for true invariants that can never be triggered by external input.

rbac/management/tenant_mapping/v2_opt_in_view.py
```diff
@@ -0,0 +1,145 @@
...
+class _OptInPermission(permissions.BasePermission):
+    def has_permission(self, request, view):
+        if view.action == "status":
+            return True
+
+        return request.user.admin
+
+
+class _OptInRequestSerializer(serializers.Serializer):
+    v2_opted_in = serializers.BooleanField(required=False, allow_null=True)
```
lpichler, line 42: **[high]**: OpenAPI spec and serializer are out of sync

The serializer accepts `v2_opted_in` as absent, `null`, or `true` (`required=False, allow_null=True`), but the spec documents it as `boolean` with `enum: [true]` and `requestBody: required: true`. Clients following the spec work, but the null/omitted no-op behavior is invisible to spec consumers, and the property is documented as required when it isn't.

**Suggestion**: Align the spec with the tri-state the code implements (document null/omission as the no-op case and drop `required: true`), or tighten the serializer to match the spec (require the field, drop `allow_null`).

rbac/management/tenant_mapping/v2_opt_in_view.py
```diff
@@ -0,0 +1,145 @@
...
+            return Response(self._format_eligibility_data(result), 422)
+
+        set_v2_opt_in_state(request.tenant, True)
+        return self._state_response_for(request.tenant)
+
+    @atomic_with_retry(retries=5)
+    def eligibility(self, request):
+        """Determine whether the requestor's tenant can be opted into V2."""
+        result = check_v2_eligibility(request.tenant)
+        return Response(self._format_eligibility_data(result), 200)
```
lpichler: **[high]**: `eligibility` endpoint sends no `Cache-Control` header

The `status()` endpoint sets `Cache-Control: max-age=120, private`, but `eligibility()` returns a bare `Response(...)` with no cache header. For large tenants the eligibility scan is expensive; a polling client repeats the full tenant-wide group/role scan on every call.

**Suggestion**: Add the same `Cache-Control` header (or a shorter TTL) to `eligibility()` responses:
```python
return Response(self._format_eligibility_data(result), 200, headers={"Cache-Control": "max-age=60, private"})
```

rbac/management/tenant_mapping/v2_opt_in_view.py
```diff
@@ -0,0 +1,145 @@
...
+
+from api.models import Tenant
+
+
+class _OptInPermission(permissions.BasePermission):
+    def has_permission(self, request, view):
+        if view.action == "status":
+            return True
+
+        return request.user.admin
```
lpichler: **[medium]**: `_OptInPermission` reimplements org-admin check inline, missing audit logging

`return request.user.admin` diverges from `AdminAccessPermission`, which provides two centralized behaviors this inline version omits: (1) the `ALLOW_ANY` environment bypass used in dev/test, and (2) the `SEC-MON-REQ-1` authorization-failure audit log emitted for every denied admin-gated endpoint. Opt-in is a one-way, irreversible privileged state change — denied escalation attempts should be auditable.

**Suggestion**: Compose `AdminAccessPermission` with an action check that passes `status` through for all authenticated users, so `ALLOW_ANY` and authz-failure logging are inherited automatically.

rbac/management/tenant_mapping/model.py
```diff
@@ -143,3 +143,41 @@ def source_for_role_binding_id(
...
+    return mappings[0]
+
+
+def lock_mapping_for_share(tenant: Tenant) -> TenantMapping:
+    """
+    Return the TenantMapping for the provided tenant locked FOR SHARE.
+
+    Raises TenantNotBootstrappedError if no TenantMapping exists.
+    """
+    # TODO fix circular dependency
```
lpichler: **[medium]**: Circular import between model layer and service layer (TODO left unresolved)

`lock_mapping_for_share` uses a function-local import of `TenantNotBootstrappedError` from `management.tenant_service.v2` (a higher-level service module) to break a circular dependency, acknowledged by the `# TODO fix circular dependency` comment. A model module depending on a service-layer exception is an unsound boundary that is fragile to import ordering.

**Suggestion**: Move `TenantNotBootstrappedError` to a shared low-level module (e.g. a new `tenant_mapping/exceptions.py` or inline in `tenant_mapping/model.py`) that both layers can import without a cycle, then remove the local import and the TODO.

rbac/management/tenant_mapping/v2_opt_in_view.py
```diff
@@ -0,0 +1,145 @@
...
+            # We've not been asked to opt the tenant in, so we have nothing to do.
+            return self._state_response_for(request.tenant)
+
+        if serializer.validated_data["v2_opted_in"] is not True:
+            raise AssertionError("Validation should have rejected v2_opted_in being false")
+
+        result = check_v2_eligibility(request.tenant)
+
+        if not isinstance(result, OptInEligibleState):
+            return Response(self._format_eligibility_data(result), 422)
```
lpichler: **[low]**: Hardcoded HTTP status integers instead of DRF named constants

`Response(..., 422)` and `Response(..., 200)` use raw integers; the test file consistently uses `status.HTTP_422_UNPROCESSABLE_ENTITY` and `status.HTTP_200_OK`. Inconsistency makes typos harder to catch.

**Suggestion**: `from rest_framework import status` (already imported as `permissions`) and use `status.HTTP_422_UNPROCESSABLE_ENTITY` / `status.HTTP_200_OK`.

rbac/management/tenant_mapping/v2_eligibility.py
```diff
@@ -0,0 +1,232 @@
...
+        if result is not None
+    ]
+
+
+@atomic
+def check_v2_eligibility(tenant: Tenant) -> OptInEligibleState | OptInIneligibleState:
+    """
+    Determine whether the provided tenant is eligible to opt into V2.
+
+    Returns an OptInEligibleState if the tenant is eligible, or na OptInIneligibleState otherwise.
```
lpichler: **[low]**: Docstring typo: "or na OptInIneligibleState"

`check_v2_eligibility` docstring reads "or na OptInIneligibleState otherwise" — should be "or an".

**Suggestion**: Fix `na` → `an`.

docs/source/specs/openapi.json
```diff
@@ -2804,6 +2808,181 @@
...
+            }
+          },
+          "400": {
+            "description": "The server could not understand the request due to invalid syntax."
+          },
+          "401": {
+            "description": "Unauthorized"
+          },
+          "403": {
+            "description": "Insufficient permissions to list permissions",
```
lpichler: **[low]**: Copy-paste error in spec: wrong 403 description for `GET /tenant/opt-in/`

The 403 response description reads "Insufficient permissions to list permissions" — clearly carried over from an unrelated permissions endpoint.

**Suggestion**: Update to something like "Insufficient permissions to read opt-in status".

docs/source/specs/openapi.json
```diff
@@ -2804,6 +2808,181 @@
...
+          "400": {
+            "description": "The server could not understand the request due to invalid syntax."
+          },
+          "401": {
+            "description": "Access is unauthorized."
+          },
+          "403": {
+            "description": "Access is forbidden.",
+            "content": {
+              "application/problem+json": {
```
lpichler: **[medium]**: Inconsistent error `Content-Type` across new endpoints

`GET /tenant/opt-in/` and `PATCH /tenant/opt-in/` document 403/500 errors as `application/json`, while `GET /tenant/opt-in/eligibility/` uses `application/problem+json` for the same error schemas (`Error`, `Error403`). Clients cannot reliably deserialize errors if the media type varies.

**Suggestion**: Pick one media type for error bodies (`application/json` or `application/problem+json`) and apply it consistently across all three operations.

lpichler (commented):
Automated review — 12 actionable findings (5 high, 4 medium, 3 low). See inline comments.

tests/management/tenant_mapping/test_v2_opt_in_view.py
```diff
@@ -0,0 +1,201 @@
...
+        self.assertEqual(
+            response.data["ineligible_custom_roles"],
+            [{"name": custom_role.name, "uuid": str(custom_role.uuid), "ineligible_applications": ["ineligible"]}],
+        )
+
+    def test_opt_in_ineligible(self):
+        self._assert_ineligible_response(self._send_opt_in_request, status.HTTP_422_UNPROCESSABLE_ENTITY)
+        self._assert_status(False)
+
+    def test_opt_in_empty_body(self):
```
lpichler, line 171: **[medium]**: Test coverage gaps and dead imports

(a) `test_opt_in_empty_body` sends `{}` and asserts `200/{"v2_opted_in": False}`, but the code raises `KeyError` (finding #1) — this test would fail or not cover what it claims.
(b) Both test classes use `@override_settings(ATOMIC_RETRY_DISABLED=True)`, so retry logic is never exercised.
(c) Missing coverage: eligible-tenant path in `test_v2_eligibility.py`, already-opted-in eligibility short-circuit, `Cache-Control` assertions on PATCH/eligibility.
(d) Unused import: `lock_v2_opt_in_state` in this file.

**Suggestion**: Fix finding #1 first; update `test_opt_in_empty_body` to assert the real behavior; add a retry integration test without `ATOMIC_RETRY_DISABLED`; remove unused imports; add the missing scenarios.

rbac/management/tenant_mapping/v2_opt_in_view.py
randomnetcat: The test for this passes. `BooleanField` forcibly defaults to `false` or `null` if the key is absent.

rbac/management/tenant_mapping/v2_opt_in_view.py
randomnetcat: This is unreachable; the serializer allows only `None` or `True`.

rbac/management/tenant_mapping/v2_opt_in_view.py
randomnetcat: I'm unconvinced this is a good idea. The only endpoint that's actually likely to be frequently checked is the one for current V2 status (since every Console user will need to check that).

tests/management/tenant_mapping/test_v2_opt_in_view.py
randomnetcat: (a) The test passes; it does not actually raise `KeyError`.
(b) We have no good way to test it.
(c) Fixed.
(d) Fixed.

rbac/management/tenant_mapping/v2_opt_in_view.py
randomnetcat: Accepting `null` isn't meant to be contractual (hence why it's not in the spec). Django has some annoying behavior with `BooleanField`, where it will default to false if the field is absent. I considered going to some heroics to accept `{}` while rejecting `{"v2_opted_in": null}` but initially decided against it.

randomnetcat: Alright, that should be fixed.

tests/management/tenant_mapping/test_v2_opt_in_view.py
randomnetcat: Ah, okay, apparently the default behavior for `APIClient` is to send HTML requests... which was screwing up the testing. (a) should actually be fixed.

rbac/management/tenant_mapping/v2_opt_in_view.py
randomnetcat: Regardless, fixed now.

lpichler (approved):

PR #3375, opened 2026-09-09
fix(inventory): return 503 on OAuth outage [RHCLOUD-51009]

lpichler:
## Summary
- classify transient Inventory OAuth connection and timeout failures
- log the SSO endpoint host and request context without secrets
- return a retryable 503 instead of an empty 200 workspace result
- preserve existing handling for unrelated filter exceptions

### Before

When the SSO token request was reset:

- Permission checks could expose an unhandled `requests.exceptions.ConnectionError` as HTTP 500.
- Workspace list filtering caught the exception broadly, returned `queryset.none()`, and produced HTTP 200 with an empty workspace list. This was indistinguishable from a user with no accessible workspaces.

### After

The Inventory OAuth failure is propagated as a retryable service-unavailable response.

V2 example:

```http
HTTP/1.1 503 Service Unavailable
Retry-After: 1
Content-Type: application/problem+json
```

```json
{
  "status": 503,
  "title": "An error occurred.",
  "detail": "Inventory authorization is temporarily unavailable. Please try again shortly.",
  "errors": [
    {
      "message": "Inventory authorization is temporarily unavailable. Please try again shortly."
    }
  ]
}
```

V1 example:

```json
{
  "errors": [
    {
      "detail": "Inventory authorization is temporarily unavailable. Please try again shortly.",
      "status": 503
    }
  ]
}
```

The response tells clients to retry instead of treating an authentication outage as an empty authorization result.

## Logging

The failure is explicitly identified as an SSO token-endpoint problem without logging credentials or tokens:

```text
Inventory API OAuth token request failed at SSO endpoint: endpoint_host=sso.stage.redhat.com error_type=ConnectionError
Inventory SSO unavailable during workspace access filtering: user=<user> org_id=<org> workspace_id=None relation=view
```

## Validation
- 120 focused tests passed
- 86 focused tests passed after logging changes
- pre-commit checks passed
- git diff --check passed

## Summary by CodeRabbit

- **Bug Fixes**
  - Inventory authentication outages now return a clear temporary-unavailability response with HTTP 503 status.
  - Responses include `Retry-After: 1` to indicate that requests may be retried shortly.
  - Workspace access checks now surface authentication outages instead of treating them as empty results.

- **Documentation**
  - Added guidance for handling Inventory API authentication service interruptions.
  - Documented HTTP 503 responses and required retry headers for workspace listing and querying operations.

### DIFF
--- a/docs/error-handling-guidelines.md
+++ b/docs/error-handling-guidelines.md
@@ -55,6 +55,7 @@ Defined in `api/common/renderers.py`. A thin `JSONRenderer` subclass with `media
 - `RequiredFieldError(field_name)` -- missing required field. Stores `field_name`.
 - `InvalidFieldError(field, message)` -- invalid field value. Stores `field`.
 - `NotFoundError(resource_type, resource_id)` -- resource not found.
+- `InventoryAuthUnavailableError` -- Inventory API OAuth token service temporarily unavailable (503 with `Retry-After`).
 
 ### Role v2 (`management/role/v2_exceptions.py`)
 Hierarchy rooted at `RoleV2Error`:
--- a/docs/source/specs/typespec/main.tsp
+++ b/docs/source/specs/typespec/main.tsp
@@ -265,7 +265,10 @@ namespace Problems {
         type: ProblemType.ServiceUnavailable,
         title: "A downstream dependency is temporarily unavailable. Please retry shortly.",
         @example("Inventory authorization is temporarily unavailable. Please try again shortly.")
-        detail: string
+        detail: string,
+        @header("Retry-After")
+        @example("1")
+        retryAfter: string
     }
 
     @error
--- a/docs/source/specs/typespec/main.tsp
+++ b/docs/source/specs/typespec/main.tsp
@@ -216,6 +216,7 @@ namespace Problems {
         InvalidRequest:"http://project-kessel.org/problems/invalid-request",
         AlreadyExists:"http://project-kessel.org/problems/already-exists",
         WorkspaceNotEmpty:"http://project-kessel.org/problems/workspace-not-empty",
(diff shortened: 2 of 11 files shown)


randomnetcat (approved):

astrozzc (commented):
1. Severity: Minor — Log hostname, not netloc (CodeRabbit, still open)

urlparse(...).netloc can include URL userinfo if INVENTORY_API_TOKEN_URL ever embeds credentials.


utils.py
Ln 85–85
token_endpoint_host = urlparse(settings.INVENTORY_API_TOKEN_URL).netloc or "unknown"
Suggestion: Use .hostname or "unknown" instead. Low risk today, easy fix.

2. Severity: Bug risk (narrow) — Only ConnectionError and Timeout are classified

If get_token() raises other transient failures (e.g. requests.exceptions.HTTPError when SSO returns 503/502, or SSLError), they still hit the filter’s broad handler:


filters.py
Ln 95–104
except Exception as e:
    ...
    return queryset.none()
That recreates the original bug for non-connection/timeout SSO failures. Worth confirming what OAuth2ClientCredentials.get_token() actually raises during an outage; if HTTP 5xx is common, consider also mapping requests.exceptions.RequestException (or the specific subclasses you see in practice).

3. Severity: Minor — Double logging on workspace paths

get_inventory_auth_metadata() logs a warning with exc_info=True, and the filter logs again with exc_info=True on re-raise. Same stack trace may appear twice. Not wrong, but noisy in ops. Consider logging only at one layer, or downgrade one to debug.

4. Severity: Nit — Message inconsistency

Exception __init__: "Inventory API authentication is temporarily unavailable"
Client response: "Inventory authorization is temporarily unavailable..."
Harmless since handlers use the hardcoded client string, but slightly confusing for anyone reading exception messages in logs.

astrozzc (approved):

PR #3398, opened 2026-09-15
[RHCLOUD-51288] Merge the old principal into new one

astrozzc:
## Link(s) to Jira
- https://redhat.atlassian.net/browse/RHCLOUD-51288

The username might have been updated, so the umb processor can't recognize the user, trying to create a new one with same user_id.

## Local Testing
Unit tests

## Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tenant user provisioning when duplicate user identifiers are detected.
  * Preserves group memberships and direct role assignments when consolidating duplicate principals.
  * Synchronizes resulting group-membership changes across replicated environments.
  * Avoids unnecessary updates when the principal already has the expected user identifier.
  * Prevents consolidation across tenants and when an existing principal has a conflicting non-empty identifier.

* **Tests**
  * Added coverage for consolidation, membership transfer, role assignments, replication, and duplicate-identifier scenarios.


rbac/management/tenant_service/tenant_service.py
```diff
@@ -12,16 +17,183 @@
...
+    """
+    if survivor.pk == obsolete.pk:
+        return
+
+    with transaction.atomic():
+        obsolete_groups = list(obsolete.group.all())
+        obsolete_user_id = obsolete.user_id
+        obsolete_username = obsolete.username
+
+        if survivor.tenant_id != obsolete.tenant_id:
```
randomnetcat: Is there any situation where this is possible?

rbac/management/tenant_service/tenant_service.py
```diff
@@ -12,16 +17,183 @@
...
+    The survivor is typically the latest BOP username: it could not receive user_id due to the
+    global uniqueness constraint, but may already have been added to groups. Group memberships
+    and direct role-binding entries from the obsolete principal are transferred to the survivor
+    when both principals are in the same tenant. user_id is assigned to the survivor and the
+    obsolete principal is deleted.
+    """
+    if survivor.pk == obsolete.pk:
+        return
+
+    with transaction.atomic():
```
randomnetcat: This isn't strong enough; it needs to be `SERIALIZABLE` if we're directly touching `RoleBinding`s.

rbac/management/tenant_service/tenant_service.py
```diff
@@ -12,16 +17,183 @@
 logger = logging.getLogger(__name__)
 
 
-def _ensure_principal_with_user_id_in_tenant(user: User, tenant: Tenant, upsert: bool = False):
+def _is_missing_user_id(user_id: Optional[str]) -> bool:
+    return user_id is None or user_id == ""
+
+
+def _transfer_role_binding_entries(obsolete: Principal, survivor: Principal) -> None:
```
randomnetcat, line 34: This also needs to do replication.

rbac/management/tenant_service/tenant_service.py
```diff
@@ -12,16 +17,183 @@
...
+        deleted, _ = obsolete.role_binding_entries.all().delete()
+        if deleted:
+            logger.info(
+                "Removed %d role binding entries from cross-tenant obsolete principal. obsolete_username=%s",
+                deleted,
+                obsolete.username,
+            )
+        return
+
+    for entry in list(obsolete.role_binding_entries.all()):
```
randomnetcat: We technically have to have all the relevant `RoleBinding`s locked here in case a V1 tenant attempts to update a CAR while we're changing this.

rbac/management/tenant_service/tenant_service.py
```diff
@@ -12,16 +17,183 @@
...
+            source=entry.source,
+        )
+        entry.delete()
+
+
+def _transfer_group_memberships(obsolete: Principal, survivor: Principal, obsolete_groups: list[Group]) -> None:
+    for group in obsolete_groups:
+        if not group.principals.filter(pk=survivor.pk).exists():
+            group.principals.add(survivor)
+        group.principals.remove(obsolete)
```
randomnetcat: The check here is redundant.

rbac/management/tenant_service/tenant_service.py
```diff
@@ -12,16 +17,183 @@
...
+
+
+def _assign_user_id_and_replicate_merge(
```
randomnetcat, line 75: We should likely do the replication in the same place as the local changes to make it more obviously correct.

rbac/management/tenant_service/tenant_service.py
```diff
@@ -12,16 +17,183 @@
...
     principal = None
 
     if upsert:
-        principal, created = Principal.objects.get_or_create(
-            username=user.username,
-            tenant=tenant,
-            defaults={"user_id": user.user_id},
-        )
+        try:
+            defaults = {"user_id": user.user_id} if user.user_id else {}
```
randomnetcat: This should use `if not _is_missing_user_id(user.user_id)` to be consistent.

rbac/management/tenant_service/tenant_service.py
```diff
@@ -12,16 +17,187 @@
...
+def _assign_user_id_and_replicate_merge(
+    survivor: Principal,
+    obsolete_username: str,
+    obsolete_user_id: Optional[str],
+    obsolete_groups: list[Group],
+    user_id: str,
+    replicator: Optional[InventoryReplicator],
+) -> None:
+    survivor.user_id = user_id
+    survivor.save()
```
randomnetcat, line 83: If this can change from one non-empty value to another non-empty value, this is not robust in the face of concurrent assignments. (Another process could be replicating a group -> principal relationship concurrently and use the old `user_id`, and we wouldn't catch it here.)

This probably shouldn't be able to happen anyway?

tests/management/tenant_service/test_merge_principal.py
```diff
@@ -0,0 +1,218 @@
...
+        obsolete_group = Group.objects.create(name="legacy", tenant=self.tenant)
+        obsolete_group.principals.add(obsolete)
+
+        tracker = _ReplicationTracker()
+        result = merge_obsolete_principal_into_survivor(survivor, obsolete, user_id="54181241", replicator=tracker)
+        self.assertIsNone(result)
+
+        expected_removed = Group.relationship_to_user_id_for_group(str(obsolete_group.uuid), "54181241")
+        self.assertEqual(len(tracker.tuples_removed), 1)
+        self.assertEqual(tracker.tuples_removed[0], expected_removed)
```
randomnetcat: `self.assertCountEqual(tracker.tuples_removed, [expected_removed])`

tests/management/tenant_service/test_merge_principal.py
```diff
@@ -0,0 +1,218 @@
...
+        self.assertEqual(len(tracker.tuples_added), 1)
```
randomnetcat: We should assert what this is?

rbac/management/tenant_service/tenant_service.py
astrozzc: group.relationship_to_principal(survivor) needs survivor.user_id. During _transfer_group_memberships, the survivor doesn't have one yet. So you can't literally replicate adds next to group.principals.add(survivor) — you'd still need a second step after assignment.

I refactored this a bit

rbac/management/tenant_service/tenant_service.py
astrozzc: The tuples are keyed by user_id which does not change. So not needed here

rbac/management/tenant_service/tenant_service.py
astrozzc: Removing, let's concern if it happens in the future

rbac/management/tenant_service/tenant_service.py
```diff
@@ -1,27 +1,195 @@
...
+
+
+def _transfer_role_binding_entries(obsolete: Principal, survivor: Principal) -> None:
+    """Move direct role-binding entries from obsolete to survivor."""
+    entries = list(obsolete.role_binding_entries.select_related("binding").all())
+    if not entries:
+        return
+
+    binding_ids = {entry.binding_id for entry in entries}
+    RoleBinding.objects.select_for_update().filter(pk__in=binding_ids)
```
randomnetcat: We can't *merely* lock them; we'll need to reload the entries either as we lock them or once they are locked.

rbac/management/tenant_service/tenant_service.py
```diff
@@ -1,27 +1,195 @@
...
+        logger.warning(
+            "Refusing cross-tenant principal merge. "
+            "survivor_id=%s survivor_username=%s obsolete_id=%s obsolete_username=%s user_id=%s org_id=%s",
+            survivor.pk,
+            survivor.username,
+            obsolete.pk,
+            obsolete.username,
+            user_id,
+            survivor.tenant.org_id,
+        )
```
randomnetcat: We should throw here; `_ensure_principal_with_user_id_in_tenant` will have failed to establish its invariant.

rbac/management/tenant_service/tenant_service.py
```diff
@@ -32,9 +200,36 @@ def _ensure_principal_with_user_id_in_tenant(user: User, tenant: Tenant, upsert:
...
+
+    if not _is_missing_user_id(principal.user_id):
+        logger.warning(
+            "Principal user_id does not match BOP user_id; refusing merge. "
+            "username=%s principal_user_id=%s bop_user_id=%s org_id=%s",
+            principal.username,
+            principal.user_id,
+            user.user_id,
+            tenant.org_id,
+        )
```
randomnetcat: Similarly, we should throw here.

rbac/management/tenant_service/tenant_service.py
```diff
@@ -1,27 +1,195 @@
...
+        return
+
+    obsolete_groups = list(obsolete.group.all())
+    obsolete_user_id = obsolete.user_id
+    obsolete_username = obsolete.username
+
+    tuples_to_remove = _transfer_group_memberships(obsolete, survivor, obsolete_groups, obsolete_user_id)
+    _transfer_role_binding_entries(obsolete, survivor)
+    obsolete.delete()
```
randomnetcat, line 138: Blegh, I think we are likely broadly insufficiently careful about deleting principals. Group operations probably technically need to lock all the relevant principals `FOR SHARE` to protect against concurrent deletions (or we need to make everything `SERIALIZABLE`). That's probably out of scope of this PR, but I'll create a ticket.

rbac/management/tenant_service/tenant_service.py
```diff
@@ -1,27 +1,195 @@
...
+        group.principals.remove(obsolete)
+    return tuples_to_remove
+
+
+def _assign_user_id_and_replicate_merge(
+    survivor: Principal,
+    obsolete_username: str,
+    tuples_to_remove: list[RelationTuple],
+    user_id: str,
+    replicator: Optional[InventoryReplicator],
```
randomnetcat: We shouldn't permit a `None` replicator here; it'd be throwing away important information.

rbac/management/tenant_service/tenant_service.py
```diff
@@ -1,27 +1,195 @@
...
+            source=entry.source,
+        )
+        entry.delete()
+
+
+def _transfer_group_memberships(
+    obsolete: Principal,
+    survivor: Principal,
+    obsolete_groups: list[Group],
+    obsolete_user_id: Optional[str],
```
randomnetcat: Taking this seems redundant? And I think we should always expect it to not be `None` at this point.

rbac/management/tenant_service/tenant_service.py
randomnetcat: Ah, sure.

rbac/management/tenant_service/tenant_service.py
```diff
@@ -1,27 +1,195 @@
...
+
+
+def _resolve_user_id_conflict(
+    survivor: Principal,
+    user_id: str,
+    replicator: Optional[InventoryReplicator],
+) -> None:
+    obsolete = Principal.objects.filter(user_id=user_id, tenant=survivor.tenant).exclude(pk=survivor.pk).first()
+    if obsolete is None:
+        raise IntegrityError(f"user_id={user_id} is already assigned but no obsolete principal was found")
```
randomnetcat: This could happen if this function runs twice, but one finishes before the other starts. (And we probably shouldn't raise `IntegrityError` directly?)

tests/management/tenant_service/test_merge_principal.py
```diff
@@ -0,0 +1,240 @@
...
+        obsolete_group = Group.objects.create(name="legacy", tenant=self.tenant)
+        survivor_group.principals.add(survivor)
+        obsolete_group.principals.add(obsolete)
+
+        result = merge_obsolete_principal_into_survivor(survivor, obsolete, user_id="54181241")
+        self.assertIsNone(result)
+
+        survivor.refresh_from_db()
+        self.assertEqual(survivor.user_id, "54181241")
+        self.assertEqual(set(survivor.group.values_list("name", flat=True)), {"engineering", "legacy"})
```
randomnetcat: This can just be `assertCountEqual` rather than having an explicit conversion to `set`.

rbac/management/tenant_service/tenant_service.py
randomnetcat: [RHCLOUD-51350](https://redhat.atlassian.net/browse/RHCLOUD-51350)

rbac/management/tenant_service/tenant_service.py
```diff
@@ -1,27 +1,204 @@
...
+
+    def replicate(self, event: ReplicationEvent):
+        pass
+
+
+def _ensure_principal_with_user_id_in_tenant(
+    user: User,
+    tenant: Tenant,
+    upsert: bool = False,
+    replicator: Optional[InventoryReplicator] = None,
```
randomnetcat: This definitely shouldn't have a default that throws data away, and it likely shouldn't have a default at all.

rbac/management/tenant_service/tenant_service.py
```diff
@@ -1,27 +1,204 @@
...
+            f"survivor_id={survivor.pk} survivor_username={survivor.username}"
+        )
+    merge_obsolete_principal_into_survivor(survivor, obsolete, user_id=user_id, replicator=replicator)
+
+
+class _NullReplicator(InventoryReplicator):
+    """No-op replicator for V1 tenants that don't use Kessel Relations."""
+
+    def replicate(self, event: ReplicationEvent):
+        pass
```
randomnetcat: `NoopReplicator` already exists

rbac/management/tenant_service/tenant_service.py
randomnetcat: We can't re-query arbitrary entries from this, since we might get a new binding that hasn't been locked from a concurrent transaction committing. We should instead consider only entries from the bindings we have locked.

rbac/management/tenant_service/tenant_service.py
```diff
@@ -1,27 +1,204 @@
...
+    obsolete: Principal,
+    survivor: Principal,
+    obsolete_groups: list[Group],
+) -> list[RelationTuple]:
+    """Transfer group memberships and return SpiceDB remove tuples for the obsolete principal."""
+    tuples_to_remove: list[RelationTuple] = []
+    for group in obsolete_groups:
+        if obsolete.user_id:
+            tuples_to_remove.append(Group.relationship_to_user_id_for_group(str(group.uuid), obsolete.user_id))
+        if not group.principals.filter(pk=survivor.pk).exists():
```
randomnetcat: This check is redundant.

rbac/management/tenant_service/tenant_service.py
randomnetcat: That... doesn't address this problem?

rbac/management/tenant_service/tenant_service.py
```diff
@@ -1,27 +1,204 @@
...
+                    "user_id": user_id,
+                },
+                partition_key=PartitionKey.byEnvironment(),
+                add=tuples_to_add,
+                remove=tuples_to_remove,
+            )
+        )
+
+
+@atomic_with_retry(retries=3)
```
randomnetcat: Ah, and we can't do retries here, since we'll likely already be in a transaction. `SERIALIZABLE` will also have to be propagated up the call stack.

astrozzc:
fix lint issue

rbac/management/tenant_service/tenant_service.py
```diff
@@ -1,27 +1,208 @@
...
+
+    if survivor.tenant_id != obsolete.tenant_id:
+        raise RuntimeError(
+            f"Refusing cross-tenant principal merge. "
+            f"survivor_id={survivor.pk} survivor_username={survivor.username} "
+            f"obsolete_id={obsolete.pk} obsolete_username={obsolete.username} "
+            f"user_id={user_id} org_id={survivor.tenant.org_id}"
+        )
+
+    obsolete_groups = list(obsolete.group.all())
```
randomnetcat, line 133: Ah, this also needs to be `select_for_update`.

astrozzc:
Fix unit test failure:
from management.tenant_service.tenant_service import TenantBootstrapService
  File "/var/workdir/rbac/management/tenant_service/tenant_service.py", line 9, in <module>
    from management.inventory_replicator.inventory_replicator import (
ModuleNotFoundError: No module named 'management.inventory_replicator'

astrozzc:
- backfill_remote_principal gets @atomic, that means every request that reaches get_tenant() can open a SERIALIZABLE transaction (with retry capability), even when backfill immediately no-ops (principal.user_id already set, inactive user, service account, etc. Better to only wrap when a write/merge is actually needed.

- For run_nested_transaction_retries:
 -- Fix run_nested_transaction_retries to retry only SerializationFailure / DeadlockDetected, not all DatabaseError;
 -- SerializationFailure: PostgreSQL aborts the whole transaction, not just the savepoint. The outer atomic_block() (SERIALIZABLE) has no retry. Inner savepoint retries likely won’t recover from a real serialization conflict — same class of problem randomnetcat flagged, just moved one level. Can we have a unit test to confirm this situation?

- Cleaner outer transaction needs proper retry (merge vs API conflicts). Advisory lock 42 only serializes cleaner-vs-cleaner. Merge still conflicts with live API traffic because _transfer_role_binding_entries uses select_for_update() on RoleBinding, and other paths (role/group dual-write, internal utils, middleware bootstrap, group add) lock the same rows. After this commit, middleware/group paths use SERIALIZABLE + retry, but cleaner uses: outer atomic_block() (SERIALIZABLE, no retry), inner run_nested_transaction_retries(5, update_user). That is asymmetric: merge-vs-API conflicts may cause cleaner to fail (nack / no offset commit) while API paths retry and succeed. Please fix cleaner so merge/update has a working outer retry on the full message-processing unit, OR refactor so update_user/merge runs in a top-level run_atomic_with_retry(...) boundary while ack/DLQ/lock semantics remain correct and idempotent.

astrozzc:
And fix lint issue

randomnetcat:
> Better to only wrap when a write/merge is actually needed.

All that saves us is a single pair of queries (which, to be fair, might be worth it) at the start/end of the transaction. We still have to always call it in a serializable transaction just in case it *might* need one.

randomnetcat (approved):

astrozzc:
I agree with https://github.com/project-kessel/insights-rbac/pull/3398#issuecomment-5718910475 so we would wrap it for every request. Please revert the change you made.
And we would have a followup to do backfill in only ephemeral env

